PR #15, opened 2015-02-20
Dependency cleanup time!

ajacksified:
Rename 6to5 to babel, and also upgrade some other deps.

Will need to bump plugin versions before this is merged, then minor version bump this.


btholt:
:fish:

PR #16, opened 2015-03-02
ready

danehansen:
:eyeglasses: @ajacksified @btholt

### DIFF
--- a/assets/less/base.less
+++ b/assets/less/base.less
@@ -1,5 +1,6 @@
 // Core variables and mixins
 @import "variables.less";
+@import "mixins.less";
 
 @import "../../lib/snooboots/less/mixins.less";
 
@@ -12,6 +13,7 @@
 @import "../../lib/snooboots/less/type.less";
 @import "../../lib/snooboots/less/code.less";
 
+@import "reset.less";
 @import "components/panels.less";
 @import "components/buttons.less";
 @import "components/listing.less";
@@ -19,3 +21,65 @@
 @import "components/comment.less";
 
 @import "components/layout.less";
+
+.twirly {
+  &.before:before, &.after:after {
+    position: absolute;
+    content: "";
+    border: 4px solid transparent;
+    border-left-color: @white;
+    top: 13px;
+    .transition(0.2s);
+  }
+
+  &.before:before {
+    left: 5px;
+  }
+
+  &.after:after {
+    right: -7px;
+  }
+
+  &.opened {
+    &.before:before, &.after:after {
+      -webkit-transform: rotate(90deg) translate(2px, 2px);
+      transform: rotate(90deg) translate(2px, 2px);
+    }
+  }
+}
+
+.stalagmite, .stalactite {
+  width: 18px;
+  height: 10px;
+  position: absolute;
+  left: 5px;
+  overflow: hidden;
+
+  &:after {
+    content: "";
+    width: 10px;
+    height: 10px;
+    position: absolute;
+    background: @white;
+    .rotate(45deg);
+    left: 4px;
+  }
+}
+
+// teepee
+  .stalagmite {
+    bottom: 100%;
+
+    &:after {
+      bottom: -5px;
+    }
+  }
+
+// hangs
+  .stalactite {
+    top: 100%;
+
+    &:after {
+      top: -5px;
+    }
+  }
--- a/assets/less/components/SideNav.less
+++ b/assets/less/components/SideNav.less
@@ -1,0 +1,55 @@
+.SideNav {
+  position: fixed;
+  z-index:10;
+  top: 0;
+  bottom: 0;
+  overflow-y: auto;
+  padding-top: 55px;
+  display: inline-block;
+  background-color: @gray;
+  color: @white;
+  font-size: @font-size-base + 3px;
+  .translate(-100%, 0);
+  .transition(0.2s);
+
+  &.opened {
+    .translate(0, 0);
+  }
+
+  ul {
+    padding-left: 16px;
+  }
+
+  button, a {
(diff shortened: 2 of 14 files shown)


btholt:
Let's get your dev environment set up to use two spaces instead of tabs.

assets/less/components/SideNav.less
```diff
@@ -0,0 +1,55 @@
+#SideNav {
```
btholt: Browsers don't see capital letters when reading attributes from the DOM. Instead of CapitalCasing, do dashes. `#SideNav` would be `#side-nav`.

assets/less/base.less
```diff
@@ -19,3 +20,32 @@
...
+		border: 4px solid transparent;
+		border-left-color: @white;
+		top: 13px;
+		-webkit-transition: 0.2s;
+		-moz-transition: 0.2s;
+		-o-transition: 0.2s;
+		transition: 0.2s;
+	}
+
+	&.before:before {
```
btholt: Are the `.before`, `.after`, `.open` classes states? As in, they get added and taken off to start and end animations or some other state of component? If so, I think it'd smart to be prefix those with `.is-` (ie `.is-before`, etc.) to signify that it is a state rather than a stand alone state.

btholt:
You checked in a `.DS_Store` file. We should add a `.gitignore` line for that.

assets/less/base.less
```diff
@@ -19,3 +20,32 @@
...
 @import "components/layout.less";
+
+.twirly {
+	&.before:before, &.after:after {
+		position: absolute;
+		content: "";
+		border: 4px solid transparent;
+		border-left-color: @white;
+		top: 13px;
+		-webkit-transition: 0.2s;
```
btholt: These transitions should be a in a mixin or we should pull in a library like [LESS Prefixer](http://lessprefixer.com/) to handle doing all the browser prefixing.

assets/less/components/SortDropdown.less
```diff
@@ -0,0 +1,55 @@
...
+		position: absolute;
+		background: @white;
+		box-shadow: 0 0 5px 0 rgba(0,0,0,0.5);
+		padding:0 10px;
+		z-index:10;
+		line-height:25px;
+		opacity:0;
+		visibility:hidden;
+
+		div {
```
btholt: I haven't looked too much at the markup but I think it'd be better to style less on the tags like `div` and more to style on classes. Can you give classes to the divs you want affected? This would be hard to maintain in the future because any `<div></div>` in the future we add will have these stylings.

Not just in this case, but any base we're styling just based on tag.

assets/less/components/TopNav.less
```diff
@@ -0,0 +1,102 @@
...
+#TopNav {
+	height:@height;
+	background-color:@blue;
+	position:fixed;
+	width:100%;
+	z-index:10;
+	-webkit-user-select:none;
+	top:0;
+
+	.right > *, .left > * {
```
btholt: Wildcard selectors (`*`) are a really, really inefficient selector. In this particular case, the browser on its first pass is going to select every element in the DOM and then check to see if their parents have `.left` or `.right` in them.

Also, if we ever wanted to add *anything* later to those blocks, this would be difficult to deal with.

assets/less/components/SortDropdown.less
```diff
@@ -0,0 +1,55 @@
+.SortDropdown {
+	position: relative;
+
+	.twirly.after:after {
+		border-left-color:@blue;
+		top:12px;
+	}
+
+	.tab {
+		padding:none;
```
ajacksified: Add spaces after your `:`s

assets/less/components/SortDropdown.less
```diff
@@ -0,0 +1,55 @@
...
+			margin-right:5px;
+			vertical-align:sub;
+		}
+	}
+
+			a {
+				color: @gray;
+			}
+
+		.triangle {
```
ajacksified: I also have a triangle (listing.less, `.panel-heading:after`); let's nuke the one in `listing.less` and use this class instead, while you're at it, instead of defining triangles twice.

assets/less/components/TopNav.less
```diff
@@ -0,0 +1,102 @@
...
+		display:block;
+		float:left;
+		height:100%;
+	}
+
+	.left {
```
ajacksified: There's an existing `pull-left` class you can use for this.

assets/less/components/TopNav.less
```diff
@@ -0,0 +1,102 @@
...
+		float:left;
+	}
+
+		.hamburger {
```
ajacksified: De-nest this, and as much as possible of the classes below; prefer classes to element selectors so that it's easier to read and re-use styles.

assets/less/reset.less
```diff
@@ -0,0 +1,9 @@
+button {
```
ajacksified: Instead of adding this to the reset, is there an existing style you could use, or a better element? Check out the bootstrap docs (http://getbootstrap.com).

ajacksified:
It would be better to have a more descriptive sentence for your commit - `Add styles for topnav` or something to that effect

assets/less/components/TopNav.less
```diff
@@ -0,0 +1,97 @@
+@height:45px;
+
+.TopNav {
```
btholt: Another nitpick: can you resave your .less files using 2 spaces instead of tabs? It'll help keep it uniform.

btholt:
:nail_care:

I think after this it should be ready to merge.

btholt:
:fish: This looks great now. Thanks for putting up with my nitpicks.

PR #17, opened 2015-03-05
Bump version to v0.3.13. Minor fix to gulpfile and omitted superagent dependency.

btholt:
:eyeglasses: @ajacksified 

Mostly wanted to make sure you knew I'm adding superagent back in. The build fails without it so I assume it's still used.


ajacksified:
:fish:

PR #19, opened 2015-03-10
Implement cookie-based session-like behavior

ajacksified:
Also bump to latest horse/horse-react once https://github.com/reddit/horse/pull/1 and https://github.com/reddit/horse-react/pull/1 are merged.

:eyeglasses: @btholt, and also @spladug for interest and @JordanMilne in case I did something dumb


src/oauth.es6.js
```diff
@@ -27,15 +33,15 @@ var oauthRoutes = function(app) {
     var redirectURI = OAuth2.authCode.authorizeURL({
       redirect_uri: redirect,
       scope: 'history,identity,mysubreddits,read,subscribe,vote,submit,save',
-      state: req.csrfToken(),
+      state: 'lambeosaurus',
```
ajacksified: Going to reimplement CSRF in a different PR.

src/oauth.es6.js
```diff
@@ -47,26 +53,23 @@ var oauthRoutes = function(app) {
...
-      }
+    token = OAuth2.accessToken.create(result);
 
-      var user = yield app.V1Api(req).users.get(options);
+    this.cookies.set('token', token.token.access_token, { signed: true });
 
-      this.session.user = user;
-      this.redirect(req.session.redirect || '/');
+    var options = {
+      user: 'me', //the current oauth api doesn't return userid. :(
```
btholt: Can we bug platform to fix this? Seems like an oversight.

src/routes.jsx
```diff
@@ -261,9 +262,9 @@ function routes(app) {
       text: ctx.body.text
     });
 
-    var api = app.V1Api(this);
+    var api = app.V1Api(props.token);
 
-    if (!(this.session && this.session.token && this.session.token.access_token)) {
+    if (!(this.token)) {
```
btholt: Probably doesn't need to be wrapped in extra parens.

src/views/pages/listing.jsx
```diff
@@ -71,7 +72,7 @@ class ListingPage extends React.Component {
 
     if (!loading) {
       listingElement = (
-        <Listing listing={listing} single={ true } session={ session } api={api} expanded={ true } />
+        <Listing listing={ listing } single={ true } user={ user } token={ token } api={ api } expanded={ true } />
```
btholt: On these long attributes declaration, I think it's stylistically nice to put each on a new line.

```javascript
<Listing
  listing={ listing }
  single={ true }
  user={ user }
  token={ token }
  api={ api }
  expanded={ true }
/>
```

src/oauth.es6.js
ajacksified: I'm hoping that we end up implementing JWTs in OAuth and get a signed json blob back. We have an api authentication meeting on Friday to discuss it.

src/routes.jsx
ajacksified: MORE PARENS)))))))

src/views/pages/listing.jsx
ajacksified: Fair enough - I'll make a style pass on the jsx files as a separate commit after this review.

src/oauth.es6.js
JordanMilne: Currently you need the `identity` OAuth scope to learn the user's identity, how would this interact with that?

ajacksified: Do you mean JWTs, or do you mean what I have now? I'm currently getting back all of the data I need.

With JWTs, when you OAuth response returns, you'd get (at minimum) an object with only user id, and with proper scopes, you'd get more data in your user object.

src/oauth.es6.js
```diff
@@ -47,26 +53,23 @@ var oauthRoutes = function(app) {
...
+      headers: {
+        Authorization: 'bearer ' + token.token.access_token,
+      },
     }
+
+    var user = yield app.V1Api(token.token.access_token).users.get(options);
+
+    this.cookies.set('user', JSON.stringify(user), { signed: true });
```
JordanMilne: Hmm, is it possible to have `cookies.set()` optionally add a `Secure` flag to all cookies? Prod's going to be HTTPS-only so they'll all need one.

ajacksified: Sure - I'll read from the config file to determine if this should be secure.

src/config.es6.js
```diff
@@ -22,7 +22,7 @@ var config = {
   googleAnalyticsId: process.env.GOOGLE_ANALYTICS_ID,
   googleAnalyticsDomain: process.env.GOOGLE_ANALYTICS_DOMAIN,
 
-  keys: ['lambeosaurus'],
+  keys: [ process.env.SERVER_SIGNED_COOKIE_KEY || 'lambeosaurus' ],
```
JordanMilne: This should complain if it falls back to the placeholder key

/me remembers when we accidentally broke CSRF prevention on www. for a week

JordanMilne:
:nail_care: Just a couple nits, looks good!

btholt:
:nail_care: A few nits from here as well. There are a few of those really long `<Component … />` tags.

src/config.es6.js
ajacksified: Done! Throws out a `console.warn`.

PR #20, opened 2015-03-11
Allow gulp in npm-scripts to be found on the path.

spladug:
In a regular install, the executables are automatically linked into
node_modules/.bin/ by `npm install` and `npm run` will add that
directory onto the PATH, so this should work out of the box.

This allows us some flexibility in where `node_modules` actually exists
which will make production deployment easier.

You can verify that the path will work with `npm run env | grep PATH`.

:eyeglasses: @ajacksified

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "main": "src/app.es6.js",
   "scripts": {
     "start": "node --harmony index.js",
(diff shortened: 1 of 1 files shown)


ajacksified:
:fish:

spladug:
a35135c74dd625162724f788d152fbc74fd3064c

PR #24, opened 2015-03-12
Such oauth many security wow

ajacksified:
:eyeglasses: @JordanMilne

### DIFF
--- a/src/oauth.es6.js
+++ b/src/oauth.es6.js
@@ -1,3 +1,6 @@
+var crypto = require('crypto');
+var scmp = require('scmp');
+
 // set up oauth routes
 var oauthRoutes = function(app) {
   var router = app.router;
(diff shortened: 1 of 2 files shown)


src/oauth.es6.js
```diff
@@ -40,14 +42,51 @@ var oauthRoutes = function(app) {
     });
   }
 
+  function constantTimeCompare(actual, expected) {
```
JordanMilne: We could just use [`scmp`](https://www.npmjs.com/package/scmp) since [the `csrf` module already pulls it in](https://github.com/pillarjs/csrf/blob/master/package.json#L14).

src/oauth.es6.js
```diff
@@ -40,14 +43,36 @@ var oauthRoutes = function(app) {
...
+    var state = hmac(referer);
+    var redirectURI = referer;
+
+    if ((!this.get('Referer')) || (this.get('Referer') && this.get('Referer').slice(0, origin.length) === origin)) {
+      redirectURI = OAuth2.authCode.authorizeURL({
+        redirect_uri: redirect,
+        scope: 'history,identity,mysubreddits,read,subscribe,vote,submit,save',
+        state: `${state}|${referer}`,
+      });
+    }
```
JordanMilne: I'd replace `redirectURI` with `/`, or `403` in the `else` case. No point redirecting back if the referer wasn't valid.

src/oauth.es6.js
```diff
@@ -40,14 +43,36 @@ var oauthRoutes = function(app) {
...
-    });
+    var origin = app.getConfig('origin');
 
-    this.cookies.set('redirect', this.get('Referer') || '/', cookieOptions);
+    // referer spelled wrong according to spec.
+    var referer = this.get('Referer') || '/';
+    var state = hmac(referer);
+    var redirectURI = referer;
+
+    if ((!this.get('Referer')) || (this.get('Referer') && this.get('Referer').slice(0, origin.length) === origin)) {
```
JordanMilne: Don't think this check is right: 

```javascript
< "h.reddit.com.whatever.com".slice(0, "h.reddit.com".length)
> "h.reddit.com"
```

ajacksified: Changed `origin` to `var origin = app.getConfig('origin') + '/';`

JordanMilne:
:fish: looks good!

PR #29, opened 2015-03-17
Load in routes *after* mutator plugins

ajacksified:

### DIFF
--- a/src/views/pages/index.jsx
+++ b/src/views/pages/index.jsx
@@ -100,7 +100,7 @@ class IndexPage extends React.Component {
       <main>
         { loading }
         <TopSubnav user={ user } sort={ sort } list='comments' baseUrl={ this.props.url }/>
-        <div className='container listing-container'>
+        <div className='container listing-container' ref='listings'>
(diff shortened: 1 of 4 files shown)


src/views/pages/index.jsx
```diff
@@ -100,7 +100,7 @@ class IndexPage extends React.Component {
       <main>
         { loading }
         <TopSubnav user={ user } sort={ sort } list='comments' baseUrl={ this.props.url }/>
-        <div className='container listing-container'>
+        <div className='container listing-container' ref='listings'>
```
btholt: Is this ref used somewhere? I don't see it.

btholt:
:nail_care:

src/views/pages/index.jsx
ajacksified: It's used in the `ads` plugin.

btholt: We can use refs. I just thought this one was unused. LGTM

btholt:
:fish:

PR #30, opened 2015-03-18
User profile first pass

ajacksified:
First pass on user profile. Needs some design work on the tabs and the overview page.

Waiting on API changes from @kemitche to access trophies without oauth tokens.

:eyeglasses: @danehansen, @btholt

### DIFF
--- a/assets/less/components/dropdown.less
+++ b/assets/less/components/dropdown.less
@@ -10,6 +10,11 @@
     visibility: visible;
   }
 }
+  .Dropdown.Dropdown-inline {
+    float: none;
+    display: inline-block;
+    margin-right: 1em;
+  }
 
   .Dropdown-tab {
     padding: none;
--- a/src/views/pages/userGild.jsx
+++ b/src/views/pages/userGild.jsx
@@ -1,0 +1,108 @@
+import React from 'react';
+import q from 'q';
+import querystring from 'querystring';
+
+import LoadingFactory from '../components/Loading';
+var Loading;
+
+import TopNavFactory from '../components/TopNav';
+var TopNav;
+
+import UserProfileNavFactory from '../components/UserProfileNav';
+var UserProfileNav;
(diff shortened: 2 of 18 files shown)


assets/less/components/dropdown.less
```diff
@@ -10,6 +10,11 @@
     visibility: visible;
   }
 }
+  .Dropdown.Dropdown-inline {
```
btholt: I'm assuming the semantic meaning of the capital letter is that it coincides with a React component? If so, that's a rad idea.

ajacksified: Yup, that's a @danehansen thing.

src/views/pages/userGild.jsx
```diff
@@ -0,0 +1,108 @@
...
+          <div className='well well-lg'>
+            <p>Sorry, this isn't ready yet!</p>
+          </div>
+        </div>
+      </main>
+    );
+  }
+
+  static populateData (api, props, synchronous) {
+    var defer = q.defer();
```
btholt: I thought we were going with native promises? Q is fine; I just think we should be consistent.

ajacksified: We've discussed it, but we haven't done the refactoring yet to get there in `reddit-mobile` and `snoode`. That'll follow at some point.

btholt:
:fish: I'm fine with it as is. I think mixing types of promise may eventually come back to haunt this project, but maybe not.

src/views/pages/userGild.jsx
ajacksified: Edit: found some native promises, I'll go ahead and swap this out, it's at least in the right direction then.

ajacksified: Doubleedit: since all of the pages use this pattern, I'll refactor them all at once in another pr.

PR #31, opened 2015-03-19
Fix video and preview sizing and add play icon

danehansen:
:eyeglasses: @ajacksified @btholt

### DIFF
--- a/assets/less/components/listing.less
+++ b/assets/less/components/listing.less
@@ -17,7 +17,7 @@
   background-color: @white;
   margin-bottom: 18px;
 
-  a:hover {
+  a: hover {
     text-decoration: none;
   }
 
@@ -39,7 +39,7 @@
       border-bottom: @gray-border;
     }
 
-    .stalactite:after {
+    .stalactite: after {
       .shadow;
       border: @gray-border;
     }
@@ -47,11 +47,6 @@
 
   .img-preview {
     width: 100%;
-    object-fit: cover;
-    object-position: center;
-
-    -moz-object-fit: cover;
-    -moz-object-position: center;
   }
 
   .img-nsfw {
@@ -60,7 +55,6 @@
 
   .listing-video iframe, .listing-video object {
     width: 100%;
-    height: 200px;
   }
 
   .img-responsive, .listing-selftext {
@@ -98,7 +92,7 @@
 .external-image {
   position: relative;
 
-  &:before {
+  &: after {
     font-family: 'Glyphicons Halflings';
     content: "\e066";
 
@@ -122,7 +116,7 @@
    overflow: hidden;
    position: relative;
 
-   &:after {
+   &: after {
      content: '';
      position: absolute;
      height: 3em;
@@ -151,7 +145,7 @@
   width: 16px;
   display: inline-block;
   background-image: url('/img/gilded.gif');
-  background-position-y: -4px; 
+  background-position-y: -4px;
 }
 
 .vote-form {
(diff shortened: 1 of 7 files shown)


assets/less/components/listing.less
```diff
@@ -43,12 +43,11 @@
 
   .img-preview {
     width: 100%;
-    // max-height: 200px;
-    object-fit: cover;
-    object-position: center;
+    // object-fit: cover;
```
btholt: If they're no longer in use, delete them.

assets/less/components/listing.less
```diff
@@ -174,3 +167,41 @@
 .listing-author-flair {
   margin-left: .25em;
 }
+
+.ratio16x9 {
+  position:relative;
```
ajacksified: Consistent spaces after `:`s

src/views/components/PlayIcon.jsx
```diff
@@ -0,0 +1,50 @@
...
+class PlayIcon extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+    this._open = this._open.bind(this);
+  }
+
+  render() {
+    return (
+      <SVG width={20} height={20} className='PlayIcon'>
```
ajacksified: Just curious - why SVG instead of using the font?

Also, just found out that the stock Android browser (pre-Chrome) doesn't do SVG, so we'll need some kind of icon font or image as a fallback for non-SVG browsers. We should figure out what we want to do for the future.

src/views/components/TopNav.jsx
```diff
@@ -23,7 +23,7 @@ function shorten (text, len) {
 }
 
 class TopNav extends React.Component {
-  constructor(props) {
+  constructor( props ) {
```
ajacksified: Undo all of these whitespace changes - no spaces for function params.

src/views/components/TopNav.jsx
```diff
@@ -76,23 +76,23 @@ class TopNav extends React.Component {
...
-  _onMouseLeave(str) {
-    this.setState({rollover:''});
+  _onMouseLeave( str ) {
+    this.setState( {rollover:''} );
   }
 
-  _onClick(str) {
-    console.log('TopNav._onClick', str);
+  _onClick( str ) {
+    console.log( 'TopNav._onClick', str );
```
ajacksified: Nuke the `log`

ajacksified:
Functionality looks good! :nail_care:

lib/snooboots
```diff
@@ -1 +1 @@
-Subproject commit 1490d875f06b6109e6980b27da260c8fa8303bbd
+Subproject commit e4d1db570ec9bbc50556fa35dcab5d400762a9d4
```
ajacksified: This reverts my upgrade of snooboots; was this intentional?

src/views/components/PlayIcon.jsx
danehansen: because found out tavish had a play icon he designed out

lib/snooboots
danehansen: no

src/views/components/PlayIcon.jsx
danehansen: never used a default android browser, but according to this we should be fine: http://caniuse.com/#feat=svg

ajacksified: >  tavish had a play icon he designed out
:+1:

> we should be fine
I brought my old android phone in today; does not render any of the svgs.

danehansen: strange. looked from ready that it was only old version of android browser, then 1 version that didn't support clipping, then the rest should be fine. lets check it out when you get back.

PR #36, opened 2015-03-20
Svg fallback

danehansen:
:eyeglasses: @btholt @ajacksified


ajacksified:
This causes an ugly flash of un-svg'd text when the client initially overrides the server. Instead of text, could it be replaced with either a raster image (for now) or could we add the icons into the icon font? (If an image, you could use the [image replacement](http://getbootstrap.com/css/#helper-classes-image-replacement) classes to have text with a css-applied image override.)

danehansen:
the flash is fixed. we can pass anything we want as the fallback. its setup to take a text fallback, image url, or a custom component. they are all set as a text fallback for now since we dont have assets made for anything else.

ajacksified:
:fish:

PR #39, opened 2015-03-24
Restyle comment page further

btholt:
:eyeglasses: @danehansen or @ajacksified


src/views/components/Comment.jsx
```diff
@@ -19,6 +25,8 @@ class Comment extends React.Component {
       collapsed: this.props.comment.hidden,
       showReplyBox: false,
       showTools: false,
+      favorited: false,
+      optionsOpen: false
```
ajacksified: nit: trailing comma (also line 68 below)

ajacksified:
Thoughts:

* The favorite icon is off-center: (chrome canary, OSX)

screen shot 2015-03-23 at 5 29 11 pm
(I wonder if we even need it, or if we can just put it in the menu?)
* The backgrounds look a bit out of place; do the icons not stick out enough as gray on blue?
* The alignment is weird between usernames, text, and icons:
screen shot 2015-03-23 at 5 33 14 pm

assets/less/variables.less
```diff
@@ -19,6 +19,8 @@
 
 @white: #fff;
 @gray: #505453;
+@icon-background-gray: #ADADAD;
```
danehansen: this line becomes redundant with what i have in SVG.less

ajacksified:
:nail_care:

src/views/components/EllipsisIcon.jsx
```diff
@@ -17,9 +17,10 @@ class EllipsisIcon extends React.Component {
   }
 
   render() {
+    var fill = this.props.fill || "#b8b8b8";
     return (
       <SVG width={20} height={20} fallbackText='menu'>
-        <g fill="#b8b8b8">
+        <g fill={fill}>
```
danehansen: undo!

src/views/components/ListingDropdown.jsx
```diff
@@ -52,7 +52,7 @@ class ListingDropdown extends React.Component {
   render() {
     var opened = this.state.opened;
     var listing = this.props.listing;
-    var button = <button><EllipsisIcon opened={opened}/></button>;
+    var button = <button><EllipsisIcon opened={opened} fill={this.props.iconFill} /></button>;
```
danehansen: undo!

danehansen:
:nail_care:

btholt:
@ajacksified: 

- Favorite button was in the comps. It was a star but I thought it looked too much like the gold icon so I changed it to a plus. I'm indifferent to it because I never use it but we should probably ask design.
- Added backgrounds for two reasons:
    - I was able to totally reuse the `Vote` component. Super nice to be totally dry there; just a bit of CSS to make it fit. Restyling it makes it more difficult to re-use the logic and animations.
    - I thought it made look more like a touchable button. I'm open to disagreement. This is also consistent with the look of the voting buttons on the `Listing` component.
    - I recognized that it didn't line up but I wasn't sure how to remedy it. It styled by dividing its space into 20% spaces and then centering the text in its respective 20% container. Open to suggestions on how to do it better.

ajacksified:
:+1:

As for lining up- I think it would help if the names and text lined up; if the buttons are off, that's okay, but with names, text, and buttons being misaligned, it looks broken rather than intentional.

btholt:
:haircut: 

Punted on the favorite icon. Should be easy to remove if we want to later. Fixed the alignment. Left backgrounds in.

ajacksified:
:fish:

(Note: I think the dropdown icon coloring can be changed to white instead of gray after #40 gets merged in.)

btholt:
Merged with 7c7d6ce07cf6db84915691a6cca21611a16bc675

PR #40, opened 2015-03-24
Make svgs colors come from css

danehansen:
also made some more consistancies between svg icons and also centralized emitter events to a new static class

:eyeglasses: @btholt @ajacksified


assets/less/components/SVG.less
```diff
@@ -0,0 +1,23 @@
+.SVG-fill {
+  fill: #b8b8b8;
```
ajacksified: Can you make this a variable? (Or use one of the existing `gray` variables?)

src/Constants.es6.js
```diff
@@ -0,0 +1,13 @@
+class Constants {
```
ajacksified: This could just export an object instead of a class, unless you need it to be a class for some reason.

src/views/components/PostIcon.jsx
```diff
@@ -9,21 +9,21 @@ class PostIcon extends React.Component {
...
         </g>
         <g ref='pencil'>
-          <path fill='#b8b8b8' d='M11.859481,12.920886l-6.772291,1.99185l1.99185-6.772289l9.549256-9.549258 c0.713877-0.713877,1.87553-0.713877,2.589405,0l2.191036,2.191034c0.713875,0.713877,0.713875,1.875529,0,2.589406 L11.859481,12.920886z'/>
-          <path fill='#fff' d='M17.62425-0.412876c0.16501-0.16501,0.432543-0.165011,0.597555,0l2.191034,2.191035 c0.165012,0.165011,0.16501,0.432545,0,0.597555l-0.896332,0.896333l-2.788591-2.78859L17.62425-0.412876 M15.731992,1.479381 l2.78859,2.788591l-6.77229,6.772289l-2.78859-2.788589L15.731992,1.479381 M8.162962,9.446782l2.39022,2.390221 l-3.386145,0.995924L8.162962,9.446782 M5.829183,7.39849l-2.82178,9.594069l9.59407-2.821778l9.900674-9.904946 c1.164682-1.269361,1.132223-3.249847-0.097456-4.479526l-2.191034-2.191035 c-1.229679-1.229678-3.210165-1.262138-4.479527-0.097455l-0.002137-0.002137L5.829183,7.39849z'/>
+          <path className='SVg-fill' d='M11.859481,12.920886l-6.772291,1.99185l1.99185-6.772289l9.549256-9.549258 c0.713877-0.713877,1.87553-0.713877,2.589405,0l2.191036,2.191034c0.713875,0.713877,0.713875,1.875529,0,2.589406 L11.859481,12.920886z'/>
```
ajacksified: `SVG` rather than `SVg`

ajacksified:
:nail_care:

danehansen:
:haircut: 

note to self: only merge this after brian's current PR

assets/less/components/SVG.less
```diff
@@ -0,0 +1,23 @@
...
+.SVG-stroke {
+  stroke: @light-gray;
+}
+
+.SVG-stroke-bg {
+  stroke: @white;
+}
+
+.DownvoteIcon.altered .SVG-fill {
+  fill: #9494ff;
```
ajacksified: `@periwinkle`a

assets/less/components/SVG.less
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.UpvoteIcon.altered .SVG-fill {
+  fill: #ff4500;
```
ajacksified: `@orangered`

src/Constants.es6.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
ajacksified: For es6, should be this instead of `module.exports`:

```
var constants = { ... }
export default constants
```

ajacksified:
A couple minor things, then good to go. :nail_care:

src/Constants.es6.js
danehansen: works fine as is syntax-wise, and even though this isn't a proper class i prefer to think of files like this as a static class and stick with caps

ajacksified: It operates the same way `config` does - and [only constructors should be capitalized](https://github.com/reddit/styleguide/tree/master/javascript#naming-conventions). This should be consistent.

Using es6 `export` syntax makes it consistent with the rest of the entire project.

ajacksified:
`git mv src/Constants.es6.js src/constants.es6.js`, then change everything to `constants from '../../constants'` everywhere, and then good to go.

ajacksified:
Looks like the `git mv src/Constants.es6.js src/constants.es6.js` didn't get included in your changes, so it own't build on case-sensitive file systems. :fish: after that.

btholt:
#39 has been merged so that's not blocking anymore.

ajacksified:
Merged: https://github.com/reddit/reddit-mobile/commit/fd9f132df977004f86f7fc103a53384ea98a4709

PR #42, opened 2015-03-25
Embed tracking pixel from api headers

ajacksified:
Also did a little cleanup.

Don't merge until https://github.com/reddit/reddit-public/pull/3411 (add "Access-Control-Expose-Headers" to responses, which will let us actually read the XHR headers we need)

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -103,7 +103,12 @@ function initialize(bindLinks) {
       });
     }
 
+    // Don't re-render tracking pixel on first load. App reads from state
+    // (bootstrap) on first load, so override state, and then set the proper
+    // config value after render.
+    app.setState('renderTracking', false);
     app.render(fullPathName(), true);
+    app.config.renderTracking = true;
   });
 }
 
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -40,12 +40,6 @@ var Layout;
 import BodyLayoutFactory from './views/layouts/BodyLayout';
 var BodyLayout;
 
-function wrap(fn, ctx, ...args) {
-  return new Promise(function(resolve, reject) {
-    fn.apply(ctx, args).then(resolve, reject);
-  });
-}
-
 // The main entry point to this file is the routes function. It will call the
 // React factories to get at the mutated react elements, and map routes.
 function routes(app) {
@@ -77,6 +71,7 @@ function routes(app) {
       params: ctx.params,
       api: app.V1Api(ctx.token),
       url: ctx.url,
+      renderTracking: app.getConfig('renderTracking'),
     };
 
     return Object.assign({}, defaultProps, ctx.props, props);
@@ -99,9 +94,12 @@ function routes(app) {
       sort: sort,
     });
 
-    var data = yield wrap(IndexPage.populateData, this, props.api, props, this.renderSynchronous, this.useCache);
+    var data = yield IndexPage.populateData(props.api, props, this.renderSynchronous, this.useCache);
 
-    props = Object.assign({}, data, props);
+    props = Object.assign({
+      data: data,
+      app: app,
+    }, props);
 
     try {
       var key = 'index-' + (this.params.subreddit || '') + stringify(this.query);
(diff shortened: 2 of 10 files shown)


btholt:
:fish:

ajacksified:
@btholt - ended up rewriting everything, if you could review again.

Depends on https://github.com/reddit/snoode/pull/5

assets/js/client.es6.js
```diff
@@ -103,7 +103,12 @@ function initialize(bindLinks) {
       });
     }
 
+    // Don't re-render tracking pixel on first load. App reads from state
+    // (bootstrap) on first load, so override state, and then set the proper
+    // config value after render.
+    app.setState('renderTracking', false);
```
ajacksified: Should fix this hackiness soon.

ajacksified:
Ended up rewriting most of it due to a snoode upgrade that made it cleaner to read metadata on the client side.

:eyeglasses: @btholt

btholt:
Can you fix `"reddit-mobile-plugin-metrics": "1.1.0",` to say `"reddit-mobile-plugin-metrics": "1.0.1",` in `package.json` (version number needs to be changed.) Just discovered it's not right.

src/routes.jsx
```diff
@@ -196,8 +197,10 @@ function routes(app) {
 
     var data = yield UserGildPage.populateData(props.api, props, this.renderSynchronous, this.useCache);
 
-    props = Object.assign({}, data, props);
-    props.app = app;
+    props = Object.assign({
```
btholt: Neat. Didn't know about `Object.assign`.

src/views/components/TrackingPixel.jsx
```diff
@@ -0,0 +1,29 @@
...
+  shouldComponentUpdate (nextProps, nextState) {
+    return false;
+  }
+
+  render () {
+    if (this.props.url) {
+      var trackingUrl = this.props.url + '&' + Math.random();
+
+      return (
+        <img src={ trackingUrl } style={{ display: 'none' }} width='0' height='0' />
```
btholt: I don't really know too much about tracking pixels, but the way I understand it, browsers will not request images if they display none and/or if they have no height/width which is why people will make them 1x1 and `position: absolute; top: -999px; left: -999px;`. Does that apply here?

ajacksified: I haven't seen that; chrome has been requesting the image. I'll check in firefox / safari to be sure.

src/views/pages/index.jsx
```diff
@@ -163,17 +181,15 @@ class IndexPage extends React.Component {
     }
 
     // Initialized with data already.
-    if (typeof props.listings !== 'undefined') {
-      api.hydrate('links', options, props.listings);
+    if (props.data && typeof props.data.data !== 'undefined') {
```
btholt: Are you using `typeof` here because `null`, `false`, or `0` mean something else? Otherwise, wouldn't `props.data && props.data.data` would work, right?

ajacksified: Nope - using it because `[]` is false.

src/views/components/TrackingPixel.jsx
spladug: I don't know if it matters, but if you want to be consistent with reddit.com's use of the pixels, it does:

```javascript
var img = new Image()
img.src = 'url'
```

ajacksified: All of the browsers appear to be requesting the pixel as written.

src/views/pages/index.jsx
btholt: I'm still a little lost on this. So `[]` should be considered false as well and thus should enter the block to hydrate? Or `[]` should be considered falsey and should move on because it has something there?

ajacksified: Having `[]` shows that you have data from an api call which returned an empty data set. (Or, having `null` means you made an api call to retrieve a resource at a specific id, and it returned nothing at that id.)

It should only continue in the case that you aren't calling this function with a response pre-loaded - which would cause this field to be `undefined`, rather than `[]` or `null`.

ajacksified: Having said all of that, this could probably be condensed down to `if props.data` by itself; `data` should be undefined (falsey) or it should include hydrated data.

ajacksified:
@btholt - metrics [should be v1.1.0](https://github.com/reddit/reddit-mobile-plugin-metrics/commit/b3b496c3de77974e39fa13135150fa2acae84f56)

btholt:
Then we should publish it to [npm](https://www.npmjs.com/package/reddit-mobile-plugin-metrics). :smile:

src/views/pages/index.jsx
btholt: Oh, okay. That makes sense. So we're specifically trying to avoid entering the block on `null`. That's what I was confused about. Sounds good.

ajacksified:
Whoops. Done!

btholt:
:fish:

PR #43, opened 2015-03-30
Style polish

ajacksified:
:eyeglasses: @danehansen or @btholt 

Fixes:

* Expand area under top bar to vertical height of 36px (currently at 30)
* List view headlines should be 18px Arial regular (appearing 14px bold on iOS)
* Subreddit titles should be Arial regular (currently Arial bold)
* Remove self.subreddit text from all list view posts
* Reduce vertical space between list item entries from 19px to 8px
* Remove stroke from edge of burder menu
* OP's username should be in bold rather than have blue box around it


btholt:
:fish: LGTM

PR #44, opened 2015-03-30
Design polish

ajacksified:
:eyeglasses: @danehansen or @btholt 

Fixes:

* Vote Icons large and clipped - should be 16px height & weight, color #dedede
* Down arrow by "hot" should be pointed down and centered horizontally
* List view meta-information should be limited to one line of text and fade to white if it extends past one line
* Meta-information text should be all Arial regular #919192
* Dotted graphic artifacts appearing by expanding comment button
* Timestamps on comments should be reduced (3h, not 3 hours ago, etc - as in list view)
* Mods green, admins red, special dark-red

### DIFF
--- a/src/views/components/DownvoteIcon.jsx
+++ b/src/views/components/DownvoteIcon.jsx
@@ -14,7 +14,7 @@ class DownvoteIcon extends React.Component {
 
   render() {
     return (
-      <SVG className={'DownvoteIcon' + (this.props.altered ? ' altered' : '')} width={20} height={20} fallbackText='downvote'>
+      <SVG className={'DownvoteIcon' + (this.props.altered ? ' altered' : '')} width={16} height={16} fallbackText='downvote'>
         <defs>
           <clipPath id={this._maskID}>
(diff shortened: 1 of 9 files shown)


src/views/components/DownvoteIcon.jsx
```diff
@@ -14,14 +14,14 @@ class DownvoteIcon extends React.Component {
 
   render() {
     return (
-      <SVG className={'DownvoteIcon' + (this.props.altered ? ' altered' : '')} width={20} height={20} fallbackText='downvote'>
+      <SVG className={'DownvoteIcon' + (this.props.altered ? ' altered' : '')} width={16} height={16} fallbackText='downvote'>
```
danehansen: leave all the sizing with the icons at a consistent 20x20 pixel size. then with css you can make the outer svg smaller and everything else will scale inside accordingly.

danehansen: but ideally within the design all the icons would be the same size anyways for visual consistency.

PR #46, opened 2015-03-31
Support multis

ajacksified:
Also remove leading / from /r/ and /m/ and /u/ and whatnot. And get rid of /u/.

:eyeglasses: @danehansen or @btholt or @jo-asakura

### DIFF
--- a/src/redirects.jsx
+++ b/src/redirects.jsx
@@ -11,6 +11,10 @@ function routes(app) {
   app.router.get('/user/:user', function *(next) {
     this.redirect('/u/' + this.params.user);
   });
+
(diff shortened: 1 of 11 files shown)


src/redirects.jsx
```diff
@@ -11,6 +11,10 @@ function routes(app) {
   app.router.get('/user/:user', function *(next) {
     this.redirect('/u/' + this.params.user);
   });
+
+  app.router.get('/user/:user/m/:multi', function *(next) {
+    this.redirect('/u/' + this.params.user + '/m/' + this.params.multi);
```
ajacksified: Will es6-string-templateify this whole file after this merge.

jo-asakura:
:fish:

PR #47, opened 2015-03-31
Hide nsfw

ajacksified:
nsfw-gif

:eyeglasses: @danehansen or @jo-asakura or @btholt

### DIFF
--- a/src/views/components/Listing.jsx
+++ b/src/views/components/Listing.jsx
@@ -66,6 +66,7 @@ class Listing extends React.Component {
 
   buildImage(url, embed, fixedRatio) {
     var html5 = gifToHTML5(url);
+
     if (html5) {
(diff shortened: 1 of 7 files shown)


src/views/components/Listing.jsx
```diff
@@ -196,11 +191,13 @@ class Listing extends React.Component {
...
-        expanded: !this.state.expanded,
-      });
-    }
+    this.setState({
+      expanded: !this.state.expanded,
+    });
+  }
+
+  isNSFW(listing={}) {
+    return listing.title.match(/nsf[wl]/gi) || listing.over_18;
```
jo-asakura: Will fail if not listing is provided and default value doesn't give us `title` property.

jo-asakura:
:fish:

PR #49, opened 2015-03-31
Even more fixes

ajacksified:
As the comments describe.

Notes:

* scrollcache is unbounded; I don't expect it to be an issue, but I could hide it behind an LRU cache if this makes anyone nervous
* `this.props.app.emit('page:update');` boilerplate is annoying, maybe it's time to make a base page class


ajacksified:
:eyeglasses: @btholt or @jo-asakura or @danehansen

jo-asakura:
:fish:

PR #55, opened 2015-04-06
Adding sidebar in

jo-asakura:
:eyeglasses: @danehansen


danehansen:
:fish:

src/views/pages/subredditAbout.jsx
```diff
@@ -0,0 +1,124 @@
...
+    var sort = this.props.sort || 'hot';
+
+    if (this.state.data.meta && this.props.renderTracking) {
+      tracking = (<TrackingPixel url={ this.state.data.meta.tracking } />);
+    }
+
+    var htmlDump;
+    if (!loading) {
+      var data = this.state.data.data;
+      var prettyHtml = utils.cleanHtml(data.description_html);
```
JordanMilne: This has the same issue as https://github.com/reddit/reddit-text-js/pull/1, we should be undoing the API's useless HTML escaping in Snoode rather than here so we don't depend on implementation details of the API's escaping mechanism in our apps. Also, it'd magically fix a bunch of other things

spladug:
nit: per the [reddit standard styleguide](https://github.com/reddit/styleguide/tree/master/general-style-guidelines#git-workflow), partial-work commits like "more progress on...", "final touches...", and "cleanup" should all just be squashed together as they're not relevant to future denizens of the codebase.

specifically, each commit should do one logical change to the code and should do it completely.  the commit history should describe changes to the project and the reasoning behind them, not a log of work.

src/views/components/TopNav.jsx
```diff
@@ -35,16 +70,24 @@ class TopNav extends React.Component {
 
     this.state = {
       subredditName: subredditName,
+      subredditId: null,
       rollover: '',
       sideNavOpen: false,
+      loaded: !(props.subredditName && props.token),
```
ajacksified: We could get away with setting this to `true`, and just quietly updating the `is susbscriber` option in the menu; it would make the site seem quite a bit faster.

src/views/components/TopNav.jsx
```diff
@@ -21,6 +40,22 @@ function shorten(text, len) {
   return text;
 }
 
+function loadSubredditData(ctx) {
+  // NOTE:
+  //  cannot call with `ctx.props` because `ctx.props.data`
+  //  could lead to a wrong cached resource
+  SubredditAboutPage.populateData(ctx.props.api, {
```
ajacksified: This won't render properly on the server side; the subreddit should instead be loaded from the index / about pages and be passed in. The data could go into state and be passed into individual components for a progressive load.

src/views/components/TopNav.jsx
```diff
@@ -79,22 +157,67 @@ class TopNav extends React.Component {
...
+      });
+
+      props.api.subscriptions.post(options)
+        .done(function (data) {
+          if (!Object.keys(data.data).length) {
+            this.setState({
+              userIsSubscribed: !this.state.userIsSubscribed
+            });
+          } else {
+            window.console.warn(data);
```
ajacksified: Send to an error page instead?

src/views/pages/subredditAbout.jsx
ajacksified: :+1: for moving to snoode.

PR #60, opened 2015-04-08
Fix build process to ignore certain large files; fix watcher

ajacksified:
Reduces initial build time by about 75%, and subsequent builds using `watch` are about 4s. Not perfect, but a whole lot better.

run `npm run dev` for a single, dev-only build; or `npm run watch` for a watched build.

:eyeglasses: @danehansen or @jo-asakura


jo-asakura:
:fish:

PR #61, opened 2015-04-09
Replace current loading image with svg animation

danehansen:


assets/less/components/TopNav.less
```diff
@@ -9,11 +9,6 @@
   border-bottom: @gray-border;
 }
 
-  .TopNav .loading-indicator {
```
jo-asakura: This helps to set a proper size on a loading indicator inside of the header.

danehansen: i dont understand why the loader ever even gets put inside the TopNav in the first place, but it certainly is never given any text or content and it's fixed position now for the page so doesn't really matter what its inside of, it will just be in the center of the window.

jo-asakura: in order to display subreddit's subscription/unsubscription information we need to pull API data and when user navigates between pages we don't reload the header, so there should a loading event while data gets pulled down.

danehansen:
@jo-asakura :haircut:

jo-asakura:
:fish:

PR #65, opened 2015-04-09
Move `cleanhtml` to snoode

ajacksified:
:eyeglasses: @JordanMilne 

Depends on https://github.com/reddit/snoode/pull/9


JordanMilne:
Whoops, almost [missed this](https://github.com/reddit/reddit-mobile/blob/master/src/views/pages/listing.jsx#L165), this can go too. Might actually be able to take out all the `mapComments` stuff. 

Is there anywhere else we do decoding like this?

:nail_care:

ajacksified:
`mapComments` also does a significant amount of auto-collapsing, so I don't think we can remove that yet - but we can get rid of decoding everywhere. I'll make another pass.

ajacksified:
:haircut:

JordanMilne:
:fish:

PR #66, opened 2015-04-14
Make icon font

danehansen:


ajacksified:
To run the `icons` job, you need to install lodash; can you add that to the package.json?

ajacksified:
Icon files should also be piped through `rev` (see https://github.com/reddit/reddit-mobile/blob/master/buildTasks/less.js#L20-24) to create a new filename and manifest json for s3 cache busting.

ajacksified:
:nail_care:

PR #69, opened 2015-04-16
Search

jo-asakura:
:eyeglasses: @ajacksified


src/routes.jsx
```diff
@@ -173,6 +177,54 @@ function routes(app) {
...
+
+    this.body = page;
+    this.layout = Layout;
+    this.props = props;
+  }
+
+  app.router.get('/search', searchPage);
+  app.router.get('/search/:query', searchPage);
+  app.router.get('/r/:subreddit/search', searchPage);
+  app.router.get('/r/:subreddit/search/:query', searchPage);
```
ajacksified: I'd suggest using `redirects.jsx` to combine these all into a single, canonical url. (redirect `/search/:query` to `/search?q={params.query}`, etc)

src/views/components/SearchBar.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+function debounce(func, wait) {
```
ajacksified: :no_bell: since we're pulling in lodash in some of our dependencies anyway, you could also use it here if you want.

src/views/pages/search.jsx
```diff
@@ -0,0 +1,331 @@
...
+    this._lastQueryKey = null;
+
+    this.state = {
+      results: props.results || {},
+      subreddits: props.subreddits || {},
+      loaded: props.results.data || props.subreddits.data
+    };
+  }
+
+  _composeUrl(data) {
```
ajacksified: No need to prefix any of these with `_`.

src/views/components/SearchBar.jsx
jo-asakura: Yeah, could be replaced with `_`. Didn't have it when was working on it.

jo-asakura: Private stuff.. saw that you have some `_func` in react components already, so decided to follow the idea.

src/views/pages/search.jsx
jo-asakura: Private stuff.. saw that you have some `_func` in react components already, so decided to follow the idea.

ajacksified: Oh, you're right - I missed those that @danehansen added. No change needed, we can continue to use `_`.

src/routes.jsx
jo-asakura: Why? You will end up with `/search?q=test` URLs facing end user.. do you want this? `this.redirect` modified URL on the first page load.

ajacksified: I do want that; having one, canonical url for everything is simpler.

ajacksified:
Just needs a rebase, and :fish:

PR #70, opened 2015-04-16
Fix for the search prev.button; styling issues

jo-asakura:
:eyeglasses: @ajacksified

### DIFF
--- a/assets/less/components/search.less
+++ b/assets/less/components/search.less
@@ -61,25 +61,27 @@
     display: block;
     margin: 10px 0;
     text-align: left;
+    white-space: nowrap;
 
-    img,
-    span {
+    > img,
+    > span {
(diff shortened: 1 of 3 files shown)


assets/less/components/search.less
```diff
@@ -61,25 +61,27 @@
     display: block;
     margin: 10px 0;
     text-align: left;
+    white-space: nowrap;
 
-    img,
-    span {
+    > img,
```
ajacksified: Didn't catch this earlier - it would be better to style these individually. De-nesting them from search and giving these a class like `.subreddit-icon` would make them reusable elsewhere.

ajacksified:
Semi-:fish:; if you want to get this round of bugfixes out now, and clean up the css in another pr, that's fine.

PR #71, opened 2015-04-16
Use image previews instead of full-resolution

ajacksified:
Don't merge until we get the all-clear from @xiongchiamiov.

:eyeglasses: @jo-asakura or @danehansen

### DIFF
--- a/src/views/components/Listing.jsx
+++ b/src/views/components/Listing.jsx
@@ -14,6 +14,7 @@ import PlayIconFactory from '../components/icons/PlayIcon';
 var PlayIcon;
 
 var imgMatch = /\.(?:gif|jpe?g|png)/gi;
+var gifMatch = /\.(?:gif)/gi;
 var gfyRegex = /https?:\/\/(?:.+)\.gfycat.com\/(.+)\.gif/;
 
 function isImgurDomain(domain) {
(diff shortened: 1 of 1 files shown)


src/views/components/Listing.jsx
```diff
@@ -54,6 +54,18 @@ class Listing extends React.Component {
     this.state = {
       expanded: this.props.expanded,
     };
+
+    if (typeof window !== 'undefined') {
```
jo-asakura: Would just `if (window)` work?

src/views/components/Listing.jsx
```diff
@@ -107,6 +119,42 @@ class Listing extends React.Component {
     );
   }
 
+  previewImageUrl(listing, expanded) {
```
jo-asakura: Just a thought.. would it be nice to have a safe-guard for nullable listing? If listing is `null` some things below will error out.

jo-asakura:
:fish:

src/views/components/Listing.jsx
ajacksified: good call - I'll update.

src/views/components/Listing.jsx
ajacksified: Nope, throws on the server side.

PR #80, opened 2015-04-20
Registration

ajacksified:
:eyeglasses: @kemitche @JordanMilne @jo-asakura @danehansen

### DIFF
--- a/assets/less/components/search.less
+++ b/assets/less/components/search.less
@@ -25,10 +25,7 @@
   }
 
   .container h4 {
-    cursor: default;
-    color: @blue;
-    letter-spacing: .1em;
-    text-transform: uppercase;
+    &:extend(.title);
   }
 
   .summary-container {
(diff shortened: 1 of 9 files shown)


kemitche:
:black_nib: I sign off on the flow used and leave it to the other reviewers to check on things.

JordanMilne:
Same as @kemitche. Looks fine security-wise, but I'll defer to others for everything else

assets/less/components/search.less
```diff
@@ -18,10 +18,7 @@
   }
 
   .container h4 {
-    cursor: default;
-    color: @blue;
-    letter-spacing: .1em;
-    text-transform: uppercase;
+    &:extend(.title);
```
jo-asakura: :no_bell: You can also do `.container h4:extend(.title) {}`.

ajacksified: Yeah - I think as a seperate pr, it might make sense to change the search titles to just use the `.title` class instead.

jo-asakura:
I would say looks good :fish:

PR #83, opened 2015-04-20
Add a "this is in beta" message to the top

ajacksified:
:eyeglasses: @danehansen 

screen shot 2015-04-20 at 4 41 16 pm


ajacksified:
Sets a cookie and won't show up again after you close it, unless you clear cookies / do it in incognito.

ajacksified:
:eyeglasses: @jo-asakura  too

src/views/components/BetaBanner.jsx
```diff
@@ -0,0 +1,50 @@
...
+      show: props.show,
+    };
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return (nextProps !== this.props || nextState !== this.state);
+  }
+
+  _close () {
+    document.cookie = 'hideBetaBanner=true; expires=Fri, 31 Dec 9999 23:59:59 GMT';
```
JordanMilne: Setting a cookie that expires past 2037 will cause it to expire instantly in certain browsers (due 32bit timestamps)

assets/less/components/text.less
```diff
@@ -211,3 +211,18 @@ dd {
   letter-spacing: .1em;
   text-transform: uppercase;
 }
+
+@media screen and (min-width: 0px) {
```
jo-asakura: Does it do anything?

src/server.jsx
```diff
@@ -139,6 +139,9 @@ class Server {
   }
 
   * modifyRequest (next) {
+    var showBetaBanner = !this.cookies.get('hideBetaBanner');
```
jo-asakura: Emm.. `this.showBetaBanner = !this.cookies.get('hideBetaBanner');` should do it :smile:

assets/less/components/text.less
ajacksified: Overrides an existing media tag from bootstrap. Kinda hacky =/

src/server.jsx
ajacksified: derp. yeah :D

src/views/pages/userActivity.jsx
```diff
@@ -77,7 +77,7 @@ class UserActivityPage extends React.Component {
     }
 
     return (
-      <main className="user-page user-activity">
+      <div className="user-page user-activity">
```
jo-asakura: Have you checked all the pages for layout issues? Asking only because before it was `main` and it defined global styles for all the `main` elements.. to overwrite some properties additional classes were added, like in this particular example, `user-page` overwrites `margin-top`, or in different case it's `background` and etc.

jo-asakura:
:fish: I assume you tested it :smile:

PR #86, opened 2015-04-21
Add 'search all the subreddits' label when no results in subreddit search

jo-asakura:
:eyeglasses: @ajacksified


ajacksified:
:fish:

PR #87, opened 2015-04-21
Pass useragent through to snoode per request

ajacksified:
Depends on https://github.com/reddit/snoode/pull/16.

:eyeglasses: @jo-asakura 

Also notifying @rram


jo-asakura:
:fish:

PR #89, opened 2015-04-21
Set api headers and user agent for auth requests

ajacksified:
:eyeglasses: @jo-asakura


ajacksified:
Note: needs to be cleaned up, this copypaste is gross.

jo-asakura:
:fish:

PR #96, opened 2015-04-22
Minor enhancement to the listing titles such that on a mobile device, long-pressing the title will spawn the context menu instead of spawning a copy selector.

justinnichols:
It looks like, on iOS at least, when the text of an &lt;a&gt; is housed inside a container, it will not allow a context menu on long-press and instead treats it as a copy selector action.  This enhancement is intended to keep the same look/feel to the listing title, but allow for the normal context menu on long-press.


ajacksified:
Thanks! Merged :D

PR #98, opened 2015-04-22
Add config for pass-through headers

ajacksified:
:eyeglasses: @jo-asakura 

Notifying @rram and @spladug - will allow additional config


jo-asakura:
:fish:

PR #100, opened 2015-04-22
Add snoo animation from previous version back into new version

danehansen:


src/views/components/MobileButton.jsx
```diff
@@ -31,12 +31,12 @@ class MobileButton extends React.Component {
         <a
           className={this.props.className + (this.state.hover ? ' hover' : '')}
           onClick={this.props.onClick}
-          onMouseEnter={isTouch ? null : this._over}
-          onMouseLeave={isTouch ? null : this._out}
-          onMouseMove={move}
-          onTouchMove={move}
-          onTouchStart={isTouch ? this._over : null}
-          onTouchEnd={isTouch ? this._out : null}
+          onMouseEnter={this._touch ? null : this._over}
```
jo-asakura: I added `isTouch` instead of static variable `this._touch` (https://github.com/danehansen/reddit-mobile/blob/snoo_animation/src/views/components/MobileButton.jsx#L28). Most mobile devices need *real-time* detection not just on init.

src/views/components/SVG.jsx
```diff
@@ -37,6 +41,10 @@ class SVG extends React.Component {
       return <span className='placeholder'/>;
     }
   }
+
+  componentDidMount() {
+    this._touch = Utils.touch();
```
jo-asakura: Remove it from here and move it up to `render` to get all the mobile devices working (https://github.com/danehansen/reddit-mobile/blob/snoo_animation/src/views/components/MobileButton.jsx#L28).

danehansen: if its in render it would make that call a whole bunch of times, and leaving it in componentDidMount keeps it from erroring out when rendering on node

jo-asakura: It's also fixed.. it works in node without erroring out. It also fixes all the bugs ios/android devices have.

danehansen: i dont think you understand what that is being used for. it is used in enabling a rollover state purely for the snoo logo in the header. it has nothing to do with any existing ios or android bugs.

jo-asakura: Yep, it doesn't have anything to do with existing bugs.. I'm just saying that I fixed bunch in them in recent hours and most of them were because of `this._touch = Utils.touch();` in `componentDidMount` instead of `render`. You can leave it as is in `SVG.jsx` but `MobileButton.jsx` should be reverted back to master branch version.

src/views/components/MobileButton.jsx
danehansen: the wya Utils.touch() works this code makes no effective difference and creates redundant calls

jo-asakura: There is zero difference between verifying your own property vs property on window (this is exactly what `Utils.touch()` does). In addition, it's better to check for browser's features support when you render the component and not at initializing/mounting state.

danehansen: in what instance would a change of whether a device is touch enabled happen?

danehansen: ah i didn't see until just now that you changed Utils.touch, so all of these comments probably haven't been making sense to you. but can you explain why that change to Utils.touch fixes anything? it seems like just a bunch of unnecessary checking to me.

jo-asakura: Apparently, persisting the touch state as a class property didn't work as it's expected. Dropdown element `...` invokes `render` method 2-3 times depends on the situation and page it's on.. one time out of those renders `this._touch` is null and since you check for `undefined` (not `null`) in `Utils.touch` it was getting ignored. Thus, the times when `this._touch` is null and it's mobile device like ios or android dropdown menu stopped responding because even handlers wasn't set properly.

The final decision is up to you.. but there are at least two reported issues that were fixed by using direct touch detection not cached version and they should stay working :smile:

danehansen:
:haircut: @jo-asakura

jo-asakura:
:fish: 

Once it's rolled live we need to verify that dropdowns work on mobile devices as expected.

ajacksified:
PR :100: :cake:

PR #102, opened 2015-04-23
Node.js and IO.js README links

milkshoes:
Update to the README file in which the two options for tooling / running
the app and setup "node.js" and "io.js" now have an easily accessible
link to download directly from "Getting Up and Running" instructions
section.


ajacksified:
Thanks!

PR #103, opened 2015-04-23
Add embed based on securitiy protocol

jo-asakura:
:eyeglasses: @ajacksified


src/views/components/Listing.jsx
```diff
@@ -48,6 +48,19 @@ function gifToHTML5(url) {
   }
 }
 
+function adjustUrlToAppProtocol(url, app) {
```
ajacksified: We may end up with broken images from sites that don't implement https. An `onerror` should probably be added to the image / iframe to remove it in case it 404s.

ajacksified:
:nail_care:

jo-asakura:
:haircut: @ajacksified

jo-asakura:
:haircut: @ajacksified added two small fixes -^

ajacksified:
:fish:

PR #108, opened 2015-04-23
Add media.oembed under protocol check as well

jo-asakura:
:eyeglasses: @ajacksified or @danehansen


ajacksified:
someday we'll eliminate all of this duplicated logic

but today is not that day

:fish:

jo-asakura:
lol.. so, true :smile:

PR #109, opened 2015-04-24
Fix lib import and componentDidMount bugs

danehansen:
:eyeglasses: @ajacksified @jo-asakura 
jack - part of this i beleive to resolve the weirdness we saw earlier
alex - part of this gets to the root of the touch detection problem... that the render function doesn't necessarily fired after a componenet gets mounted


jo-asakura:
Agree, adding it to state should help. :fish:

PR #113, opened 2015-04-27
Fix: Scroll in the menu on Safari on iOS 8.3

jo-asakura:
:eyeglasses: @danehansen

### DIFF
--- a/assets/less/components/SideNav.less
+++ b/assets/less/components/SideNav.less
@@ -30,7 +30,12 @@
       }
 
       &.opened {
-        max-height: 300px;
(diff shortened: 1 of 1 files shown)


danehansen:
What was the issue? It appears to work correctly on my iPhone as is. :fish:

jo-asakura:
If you're browsing with window zoomed in (150% and above) then submenus (About/Help) don't scroll since they have hardcoded height and no scroll ability.

jo-asakura:
I would say it's rare case.

PR #114, opened 2015-04-27
So meta even this acronym

ajacksified:
:eyeglasses: @jo-asakura 

Notes:

Putting together the title in the router is kind of messy, since data is loaded in the view. I plan on refactoring data gathering *back* to the router, which cleans this *and* a lot of other inconsistencies. Since title/meta is really only important for server rendering, and it mostly works on the client, this is good enough to cover use cases until I get a chance for the big refactor.

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -37,8 +37,13 @@ function modifyContext (ctx) {
   return ctx;
 }
 
-function initialize(bindLinks) {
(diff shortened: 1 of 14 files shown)


ajacksified:
Edit didn't go through; also :eyeglasses: @danehansen

jo-asakura:
:fish:

PR #116, opened 2015-04-29
Implement compact view

danehansen:

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -33,6 +33,8 @@ function modifyContext (ctx) {
   ctx.token = this.getState('token');
   ctx.user = this.getState('user');
   ctx.useCache = true;
+  var compact = this.getState('compact');
+  ctx.compact = compact === 'true' || compact === true;
 
   return ctx;
 }
--- a/src/server.jsx
+++ b/src/server.jsx
@@ -156,6 +156,8 @@ class Server {
 
     var user = this.cookies.get('user');
     this.token = this.cookies.get('token');
+    var compact = this.cookies.get('compact');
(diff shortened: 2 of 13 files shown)


assets/js/client.es6.js
```diff
@@ -33,6 +33,8 @@ function modifyContext (ctx) {
   ctx.token = this.getState('token');
   ctx.user = this.getState('user');
   ctx.useCache = true;
+  var compact = this.getState('compact');
```
jo-asakura: If it could be string and boolean then maybe: `var compact = this.getState('compact').toString() === 'true';`?

src/server.jsx
```diff
@@ -156,6 +156,8 @@ class Server {
 
     var user = this.cookies.get('user');
     this.token = this.cookies.get('token');
+    var compact = this.cookies.get('compact');
```
jo-asakura: If you choose to update client side then this should be updated to: `var compact = this.getState('compact').toString() === 'true';`.

src/views/components/Listing.jsx
```diff
@@ -369,14 +384,16 @@ class Listing extends React.Component {
   }
 
   expand(e) {
-    if (e.target && e.target.tagName === 'A' && e.target.href) {
-      return true;
-    } else {
-      e.preventDefault();
+    if(!this.props.compact) {
```
jo-asakura: I think you should move `compact` view detection inside of else statement because first `if` condition checks for links and allows user to navigate using clicked link.

inket:
It seems the compact setting is not being remembered between page views. The cookie is always reset to `false`, and when forced to `true` it is ignored by the views.

PR #122, opened 2015-05-07
Replace XXX icon with NSFW text

danehansen:
nsfw_list
nswf_compact

:eyeglasses: @ajacksified


ajacksified:
:fish:

Looks good everywhere except really old android, where images don't show up anyway; that should be fixed with the next design pass.

PR #125, opened 2015-05-07
Add theme-color attribute for Android 5.0+

russjr08:
Starting with Chrome 39 (for Android 5.0+ and up), it accepts a meta attribute (theme-color) which tells it what to color the address bar with. It also takes the 700 variant of the color and uses it for the status bar to show a darker version of the address bar.

This can be whatever color Reddit prefers it to be (for branding and such), but I think the presence of any color makes the page feel "more at home" for users who are using Android Lollipop.

### DIFF
--- a/src/views/layouts/DefaultLayout.jsx
+++ b/src/views/layouts/DefaultLayout.jsx
@@ -55,6 +55,7 @@ class DefaultLayout extends React.Component {
           { canonical }
 
           <meta name='viewport' content='width=device-width, initial-scale=1.0' />
+          <meta name='theme-color' content='#EEEEEE' />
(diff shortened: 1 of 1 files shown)


russjr08:
Here's a preview of what this looks like for those who do not have an Android device running Lollipop.


ysm98h1

ajacksified:
Looks great! How about using `#336699`, the link color?

russjr08:
@ajacksified Looks pretty good!

ysm98h1

russjr08:
If this is the color we want to go with, I'll update my commit. I can also preview any other color options if need be.

Either way, just let me know :)

ajacksified:
Looks great! Thanks.

ajacksified:
Thanks!

PR #126, opened 2015-05-07
Add submodules to getting started instructions

ArrestedDevelopment:
Alternatively, the cloning instructions could say to use `git clone --recursive`


ajacksified:
Thanks! Been forgetting to add this.

PR #128, opened 2015-05-08
Add 403 endpoint

JordanMilne:
:eyeglasses: @ajacksified 

Sending a 400 here will make crawlers think that they've encountered
a temporary error, and they'll keep now-private resources in the search
results, continually trying to recrawl it. Sending a 403 status code should(?) fix it.


src/routes.jsx
```diff
@@ -514,6 +514,26 @@ function routes(app) {
     this.props = props;
   });
 
+  app.router.get('/403', function * () {
```
ajacksified: Rather than a new route, we can set up the `/\/4\d\d/` path regex below using a capture (`/\/(4\d\d)/`) and use `ctx.captures` to get the code. Then, set the status and title based on that code.

ajacksified:
:nail_care:

JordanMilne:
:haircut: The `5xx` endpoint was pretty similar, so I handled it in the new endpoint as well

ajacksified:
:fish:

ajacksified:
https://github.com/reddit/reddit-mobile/commit/a2d70010930febd3e04dee5ee82f3c3757a40e1d

PR #130, opened 2015-05-09
Auto-refresh expired tokens

ajacksified:
Uses the refresh token to be able to keep a user logged in indefinitely.

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -47,12 +47,33 @@ function setTitle(props={}) {
   }
 }
 
+function refreshToken (app, tries=0) {
+  if (tries >= 3) {
+    window.location = '/logout';
+  }
+
+  $.get('/oauth2/refresh').done(function(token) {
+    app.setState('token', token.token);
+    app.setState('tokenExpires', token.tokenExpires);
+
+    var now = new Date();
+    var expires = new Date(token.tokenExpires);
+
+    window.setTimeout(function() {
+      refreshToken(app);
+    }, (expires - now) * .9);
+  }).fail(function() {
+    window.setTimeout(function() {
+      refreshToken(app, tries + 1);
+    }, 1000 * 10);
+  });
+}
+
 function initialize(bindLinks) {
   $(function() {
     var plugin;
     var p;
     var $body = $('body');
-
 
     config.mountPoint = document.getElementById('app-container');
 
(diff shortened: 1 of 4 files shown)


src/server.jsx
```diff
@@ -194,6 +196,29 @@ class Server {
...
+      if (!expires) {
+        yield next;
+        return;
+      }
+
+      expires = new Date(expires);
+
+      if (now > expires) {
+        var rToken = this.cookies.get('refreshToken');
+        var token = yield app.refreshToken(this, rToken);
```
ajacksified: Needs try/catch to kill the session if the refresh token request fails.

danehansen: what is "function * (next)"? not familiar with that. otherwise looks fine.

assets/js/client.es6.js
```diff
@@ -47,13 +47,34 @@ function setTitle(props={}) {
   }
 }
 
+function refreshToken (app, tries=0) {
```
danehansen: oh god i didn't realize we could do default values in es6 thats rad

curioussavage:
New syntax is a still a little foreign to me but It just took an extra read through.

So the token has to be refreshed every hour?

ajacksified:
Correct, it expires after an hour, so we have to re-get the token before it expires.

Also :eyeglasses: @JordanMilne just in case I missed something.

src/oauth.es6.js
```diff
@@ -7,26 +7,30 @@ import uuid from 'uuid';
...
-    maxAge: 1000 * 3000,
-  };
+  function setTokenCookie(ctx, token) {
+    ctx.cookies.set('token', token.token.access_token, longCookieOptions);
+    ctx.cookies.set('tokenExpires', token.token.expires_at.toString(), longCookieOptions);
 
-  if (app.getConfig('cookieDomain')) {
-    cookieOptions.domain = app.getConfig('cookieDomain');
+    if (token.token.refresh_token) {
+      ctx.cookies.set('refreshToken', token.token.refresh_token, longCookieOptions);
```
JordanMilne: Can `refreshToken` be made `HttpOnly` if it's only read / written by `/oauth/refresh`? That'd allow us to minimize the impact of XSSes.

JordanMilne:
:fish: Looks good, just one question

src/oauth.es6.js
ajacksified: Yup! longCookieOptions has `httpOnly` set to true.

JordanMilne: doh, missed that!

PR #134, opened 2015-05-12
Add support for Android/iOS 'Add to Homescreen'

russjr08:
Adds support for both iOS (Safari) and Android's (Chrome) "Add To Homescreen" functionality by removing the address bar on both platforms.

This is how it looks on iOS: http://i.imgur.com/TocYEqV.png

This is how it looks on Android: http://i.imgur.com/wLtzfBl.png


ajacksified:
Thanks! This is a good idea. Let's add more manifest data:

* `icons` that link to the icon assets in [the layout](https://github.com/russjr08/reddit-mobile/blob/add-homescreen/src/views/layouts/DefaultLayout.jsx#L64-70); that may require the `manifest.json` being a server route, like [/robots.txt](https://github.com/russjr08/reddit-mobile/blob/add-homescreen/src/serverRoutes.es6.js#L5), so you can load `app.config.assetPath`.
* `lang` should be set to `en-US` for now, in the manifest
* `short_name` should be set to `reddit`
* `orientation` should be `portrait`
* `start_url` should be `/`
* `scope` should be set to `app.config.origin`

russjr08:
@ajacksified Thanks! However, are we sure we want this to be portrait only? Tablets would be affected by that, and most tablets that are > 7" usually stick in Landscape by default (I have a Galaxy Tab 4 10.1" for example, and it's definitely used in Landscape by default).

Just food for thought, but I'll add it in for now.

ajacksified:
Ah, that's true; we probably shouldn't lock it down. (I misread some documentation earlier that said it was the *default* orientation, but now I see that it locks it down.) Disregard `orientation`!

russjr08:
@ajacksified Ah, got it.
I'm pretty new to React JS, so I'm not sure to get the manifest file to have the assetPaths in them. For example, is this
    `${assetPath}/favicon/64x64.png`
JSX syntax, or does gulp automatically read it and replace that?

Here's what I've got so far http://pastebin.com/nc1uknxq, but that seems messy and that there should be a better way to do that?

ajacksified:
You can make it even easier - `this.body` can be a regular javascript object, and the result will be stringified. (This is actually a koa thing, this won't even touch the React side of things.)

russjr08:
@ajacksified That worked perfectly! 

This is the end result, and I think it looks quite nice! http://imgur.com/mXLQAX3,MEMIRaN#0 (There's two images in that gallery)

russjr08:
Should I squash these commits? Some repositories like the commits to be squashed, others don't. Not sure what reddit's stance on squashing is.

ajacksified:
Yup, squishing is good.

Thanks! This looks awesome.

russjr08:
@ajacksified The rebase on my end seems to have gone bad. I'm going to open a new branch and create a new PR...

PR #138, opened 2015-05-12
CommentBox: move hidden inputs so we can use + selector.

curioussavage:
Fixes issue with android browser which doesnt support the
~ css selector.

:eyeglasses: @ajacksified @danehansen 

Not sure what you guys think. So this fixes the issue with the android browser not showing the button since it doesn't support the `~` selector. But it seems to introduce some strange behavior with the textarea which now only reverts to the smaller size when you tap in another focusable element. I looked at other solutions, no luck. But having it stay large doesn't really affect usability so I thought I would just leave it. thoughts?

### DIFF
--- a/assets/less/components/comment.less
+++ b/assets/less/components/comment.less
@@ -195,8 +195,8 @@
       height: 10em;
     }
 
-    &:focus ~ .btn-post,
-    &.has-content ~ .btn-post {
+    &:focus + .btn-post,
+    &.has-content + .btn-post {
       opacity: 1;
       visibility: visible;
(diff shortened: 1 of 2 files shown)


ajacksified:
I dislike the resize anyway; I find it annoying, and it's difficult to get working properly on desktop. I'm all for removing it, if it makes things simpler.

curioussavage:
:haircut: 

removed it is then.

ajacksified:
:fish:

ajacksified:
(merged)

PR #140, opened 2015-05-13
Facets

curioussavage:
:eyeglasses:  @ajacksified 

Everybody seemed to be ok with how search was when I showed it, so here is this pr with working subreddit search. 

Will want some help to separate out the requests, I think I could do most of it myself making sure the url and server rendering stuff all works fine is mostly what I'm not sure about. So maybe Ill start a new branch?

https://github.com/reddit/snoode/pull/20

### DIFF
--- a/assets/less/components/search.less
+++ b/assets/less/components/search.less
@@ -40,14 +40,11 @@
       padding-top: 0;
     }
 
-    .subreddits-list {
-      list-style: disc inside none;
-    }
   }
 
   .subreddits-list {
     list-style: none;
-    margin: 0;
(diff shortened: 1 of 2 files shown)


ajacksified:
Sure - so this to turn off facets, then another pass to break it into two separate requests?

:fish:

curioussavage:
yes. turn off facets and add the type param so we get subreddits back too.

ajacksified:
(merged)

curioussavage:
5a23020d74f3637ba7ae2362ae0852eac36307ee

PR #136, opened 2015-05-12
Refresh SideNav

danehansen:
:eyeglasses: @ajacksified

PR #142, opened 2015-05-13
Refactor loading data; get user info on request if missing

ajacksified:
Move loading data onto the request cycle so that data can be updated. Also refactor gathering data for pages into a series of promises, which allows for multiple simultaneous API requests, which should help out performance significantly.


ajacksified:
This is a prerequisite for #136; this loads subscription data into the context so it can be used to build the subscriptions list in the sidebar.

danehansen:
:fish:

PR #143, opened 2015-05-14
Send server render times to tracking server

ajacksified:
This will post the initial server render timings to statsd (via hivemind.)

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -21,6 +21,8 @@ import constants from '../../src/constants';
 import routes from '../../src/routes';
 import TweenLite from 'gsap';
 
+import getTimes from '../../src/lib/timing';
+
 // A few es5 sanity checks
 if (!Object.create || !Array.prototype.map || !Object.freeze) {
   $.getScript(window.bootstrap.assetPath + '/js/es5-shims.js', function(){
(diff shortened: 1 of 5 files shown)


danehansen:
:fish:

assets/js/client.es6.js
```diff
@@ -69,7 +71,10 @@ function refreshToken (app, tries=0) {
   });
 }
 
+
+
```
curioussavage: nit: extra whitespace here

src/routes.jsx
```diff
@@ -434,12 +438,12 @@ function routes(app) {
     this.props = props;
   }
 
-  app.router.get('/comments/:listingId', commentsPage);
-  app.router.get('/comments/:listingId/:listingTitle', commentsPage);
-  app.router.get('/r/:subreddit/comments/:listingId', commentsPage);
-  app.router.get('/r/:subreddit/comments/:listingId/:listingTitle', commentsPage);
+  app.router.get('comments.listingid', '/comments/:listingId', commentsPage);
```
curioussavage: this is for organizing routes in stats/graphite?

curioussavage:
mostly just curious, but looks good to me too!

src/routes.jsx
ajacksified: Yup. As it happens, we can also generate urls like this: https://github.com/alexmingoia/koa-router#named-routes

PR #145, opened 2015-05-15
Prevent body from scrolling while SideNav is open

danehansen:


ajacksified:
Instead of intercepting the scroll event, can you do it through css? `height: 100%; overflow: hidden;` to the `main` element while the sidenav is open seems to work, on desktop at least.

ajacksified:
:fish:

PR #155, opened 2015-05-21
Unpluginify

ajacksified:
:eyeglasses: @danehansen and @curioussavage 

The existing factory/mutator system is cumbersome. The original goal was to allow users to enable/disable features, such as ads and gold, depending on their install needs; however, the extra work required  in the vast majority of cases outweighs the niceness of breaking these out into plugins. Users will still retain the ability to enable/disable features, but through configuration rather than plugins.

The extra boilerplate needed to build templates, and the fact that the plugins are often *very* closely tied to the markup of individual pages, makes the codebase difficult to work with.

In the future, a simple plugin system should be added that at least allows adding extra routes, if not necessarily mutating individual components. If the need arises to break out plugins, time and care should be put in to build a cleaner, less invasive plugin system.


ajacksified:
This also fixes ads.

ajacksified:
Also :no_bell: @madbook and @umbrae because we had a lot of discussions about this early on in mweb development.

zeantsoi:
_this also fixes ads_... :thumbsup:

umbrae:
Props for thinking long term and choosing to not live with something you've determined is suboptimal! I'm glad we tried it. :+1:

curioussavage:
It looks like the images the images for list and compact view for the ad arent being fetched

ajacksified:
@curioussavage maybe a weird internet issue on your end?

screen shot 2015-05-21 at 10 53 48 am
screen shot 2015-05-21 at 10 53 54 am

curioussavage:
Only for the Ad. the thumbnail div still has the `unloaded` css class too

ajacksified:
Oh, missed that, I'll check it out. Thanks!

ajacksified:
:haircut: - fixed ads. https://github.com/ajacksified/reddit-mobile/commit/570dc526afb4712156f2051f2442de6b85076afb

curioussavage:
thats a :fish:  for me

ajacksified:
woo

danehansen:
i like :+1:

madbook:
:+1:

PR #156, opened 2015-05-21
Date-based referrer logic for "new user" setting

ajacksified:
If the user came from the desktop site after visiting the desktop for the first time within the last 5 minutes, treat them as a new user for experiments.

:eyeglasses: @umbrae

### DIFF
--- a/src/server.jsx
+++ b/src/server.jsx
@@ -136,6 +136,17 @@ class Server {
       if (this.cookies.get('loid')) {
         this.loid = this.cookies.get('loid');
         this.loidcreated = this.cookies.get('loidcreated');
+
+        // If user came from desktop, and is a new user, treat them as new for
+        // experiments.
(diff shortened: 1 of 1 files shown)


src/server.jsx
```diff
@@ -136,6 +136,17 @@ class Server {
       if (this.cookies.get('loid')) {
         this.loid = this.cookies.get('loid');
         this.loidcreated = this.cookies.get('loidcreated');
+
+        // If user came from desktop, and is a new user, treat them as new for
+        // experiments.
+        if (this.query.ref_source === 'desktop') {
+          let created = new Date(this.cookies.get('loidcreated'));
```
umbrae: Couldn't you just reuse `this.loidcreated` here since it's assigned above?

umbrae:
:fish: one tiny improvement suggested but really just cleanliness, feel free to use that or not.

PR #159, opened 2015-05-21
Convert reddit.com session to m.reddit.com session

ajacksified:
:eyeglasses: @kemitche and @JordanMilne and @curioussavage

### DIFF
--- a/src/server.jsx
+++ b/src/server.jsx
@@ -57,6 +57,16 @@ function setExperiment(app, ctx, id, value) {
   ctx.experiments.push({ id, value });
 }
 
+function skipAuth(app, url) {
+  return (
+    url.indexOf('/logout') === 0 ||
+    url.indexOf('/login') === 0 ||
+    url.indexOf('/register') === 0 ||
+    url.indexOf('/oauth2') === 0 ||
+    url.indexOf('/timings') === 0
+  );
+}
+
 class Server {
   constructor (config) {
     // Intantiate a new App instance (React middleware)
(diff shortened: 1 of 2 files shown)


src/server.jsx
```diff
@@ -293,6 +303,41 @@ class Server {
...
+      if (skipAuth(app, this.url)) {
+        yield next;
+        return;
+      }
+
+       let session = this.cookies.get('reddit_session');
+
+      if (!this.token &&
+          !this.cookies.get('token') &&
+          this.cookies.get('reddit_sesion')) {
```
kemitche: sesion?

src/routes.jsx
```diff
@@ -228,6 +228,7 @@ function routes(app) {
     var page;
     var sort = this.query.sort || 'hot';
     var ctx = this;
+    console.log('index', ctx.token);
```
JordanMilne: nit: is this still needed?

JordanMilne:
How does this handle logging out? Should we just disable that button? Also, should we detect a logout / account switch on .reddit.com and handle it accordingly?

ajacksified:
Log out will only delete the mweb token - it won't delete the reddit desktop session. Probably should, I'll update.

This is mostly to handle an influx of users coming in from the desktop view; I'm not attempting to handle people going back and forth. I guess I could set a second cookie, like `original_session`, to see if that matches and re-log if not? I'll think about that and possibly handle it in a second pass.

JordanMilne:
IIRC the session cookie has the base10 user id at the start it so you can parse the base36 user id out of the current user's `fullname` and compare the two, but this looks good for now :fish:

PR #161, opened 2015-05-22
Add smart "go to" field in side nav

ajacksified:
It amazes! It astounds! It's the...

screen shot 2015-05-21 at 11 05 46 pm

Magical go-to input box!

It handles input such as:
* r/whatever and /r/whatever
* u/whatever and /u/whatever
* unprefixed text - it tries to load a subreddit, and if that fails, it sends the user to search. Amazing!
* If a matching subreddit isn't found, it sends the user to search. Astounding!

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -168,6 +168,16 @@ function initialize(bindLinks) {
     }
   };
 
+  app.redirect = function(url) {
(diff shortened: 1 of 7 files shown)


ajacksified:
Oh, and it fixes things so it loads more than 25 subs.

PR #163, opened 2015-05-22
Improve healthcheck to hit reddit API

ajacksified:

cc @jdubs because I'm modifying the health check


curioussavage:
looks good to me. :fish:

ajacksified:
Closing - this will just lead to redundant alerts if mweb is down because desktop web is down.

PR #164, opened 2015-05-26
Refactor cookie get/set, use client cookies on client

ajacksified:
Fixes an issue where compact settings were forgotten.


curioussavage:
This is so nice!  

Code looks good to me.

In listing.jsx the compact prop is not being passed to the Listing component. the link is in list view when you go to the comments page.

:haircut:

ajacksified:
Thanks - I think that's intentional; since it's preexisting behavior, I'll confirm, and if it's behaving incorrectly, I'll make a separate PR.

PR #169, opened 2015-05-28
Clicking the report form shouldn't close the dropdown

ajacksified:


danehansen:
:fish:

PR #172, opened 2015-05-29
Refactor build

dwick:
- Add system notifications for start/success/fail (less and js)
- Fix js builds crashing gulp on error
- Add watch for less
- Break out larger tasks into seperate files under `buildTasks`
- Fix live reload

@ajacksified


ajacksified:
:fish::fish::fish::fish:

PR #187, opened 2015-06-01
Proper error tracing on API error

ajacksified:
This makes sure that the error logs get good contextual information. It also makes it so that when an auth call fails, it doesn't log the user out.


curioussavage:
looks good to me. :fish:

PR #190, opened 2015-06-01
Refactor Listing component

danehansen:
:eyeglasses: @ajacksified


src/views/components/ListingContent.jsx
```diff
@@ -0,0 +1,479 @@
...
+          url: _adjustUrlToAppProtocol(preview, config),
+          fallbackUrl: preview,
+        })
+      );
+    }
+
+    return this._renderPlaceholder();
+  }
+
+    _renderText(html, collapsed) {
```
JordanMilne: Can this be renamed to something that makes it obvious that it expects HTML, like `_renderSelfTextHTML`?

src/views/components/ListingContent.jsx
```diff
@@ -0,0 +1,479 @@
...
+                onError={ _handleIFrameError.bind(this, fallbackUrl) }
+                frameBorder='0'
+                allowFullScreen=''
+                sandbox='allow-scripts allow-forms allow-same-origin' />
+        </div>
+      );
+    }
+
+    //dont know exactly what case this is for or a better name...
+    _renderDangerVideo(html) {
```
JordanMilne: `renderVideoEmbedHTML`?

src/views/components/Listing.jsx
```diff
@@ -1,64 +1,15 @@
...
-      mp4: url.replace(/\.gif/, '.mp4'),
-      poster: url.replace(/\.gif/, 'h.jpg'),
-    };
-  } else if (url.indexOf('gfycat') > 8) {
-    var gfy = gfyRegex.exec(url);
-
-    if (gfy.length === 2) {
-      var id = gfy[1];
-      return {
-        iframe: (config.https || config.httpsProxy ? 'https://' : 'http://') + 'gfycat.com/ifr/' + id,
```
ajacksified: Should be able to just use `_adjustUrlToAppProtocol` here instead of the ternary

danehansen: like this? iframe: _adjustUrlToAppProtocol(url, config),

ajacksified: yup

danehansen:
:nail_care: @ajacksified

src/views/components/ListingContent.jsx
```diff
@@ -0,0 +1,441 @@
...
+    } else {
+      var options = {
+        onClick: data.onClick,
+        noRoute: data.noRoute,
+      };
+      return this._renderImage(data.url, data.href, options);
+    }
+  }
+
+  _renderText(html, collapsed) {
```
JordanMilne: nit: The function name should make it clear that this actually accepts HTML

danehansen: theres also a function to render "dangerous html" called _renderHTML, and this one, even though it takes html, is specifically for text so there will be no imagery or videos. the only html going in here is <p>, <h1>, etc...

danehansen: great. my quoted markup turns into actual markup on here.

JordanMilne: Right, but the main thing here is to make it clear at the call site that you're going to do something potentially unsafe (in this case, passing the `html` parameter to `dangerouslySetInnerHTML`.)

A silly example, but something like `_renderText(listing.selftext)` looks innocuous, but really it'll create an XSS hole because untrusted, unescaped input will be passed into an HTML sink. An example of a good name is React's `dangerouslySetInnerHTML`, which makes it clear that the caller needs to think extra hard about what they're passing in.

If this is just meant to render selftext HTML, maybe `_renderSelfTextHTML` or `_renderMarkdownHTML`?

ajacksified:
Everything appears to be working fine; I'll test on more devices, but the initial pass looks good. A few thoughts:

* I found an example of a really long image, if you want to try out some kind of cropping:  `/r/space/comments/38hysk/this_is_cool_everything_you_ever_wanted_to_know/`
* Is it possible to implement grey placeholders on the server render? They appear properly on lazyloading, which is awesome, but it would help out slow phones / connections.
* The hitbox for upvote should include the score
* "Visited" subreddit links appear blue; they should stay dark gray

:nail_care:

limmmster:
Some product feedback:
-add r/ to subreddit name
-lowercase subreddits across both views
-hide subreddit icon in compact view
-fonts for subreddit and title a tad smaller

danehansen:
:haircut: @ajacksified

ajacksified:
Just a few compact view changes:

* Reduce compact title font size to 14px;
* Reduce top-padding of comment buttons (comments, voting) to 0 when in compact view
* The play button in compact view replaces the thumbnail with videos / rich content. They should be a link to the content, unless you're implementing the expand action this PR

:nail_care:

limmmster:
https://reddit.atlassian.net/browse/MOBILEWEB-165

danehansen:
@ajacksified :haircut:

danehansen:
or :nail_care:  or whatever

ajacksified:
:fish:

PR #198, opened 2015-06-04
Fix redirect on expired tokens

ajacksified:

The wrong oauth client id was being used for converted sessions, so refreshing the token was failing and redirecting users to the home page.


curioussavage:
lgtm :fish:

PR #201, opened 2015-06-08
Remove unnecessary passing of random seed

danehansen:
@ajacksified :eyeglasses:

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -231,8 +231,6 @@ function routes(app) {
 
     props.app = app;
     props.api = app.api;
-    props.seed = ctx.seed;
-    props.random = ctx.random;
 
     if (ctx.token) {
       props.token = ctx.token;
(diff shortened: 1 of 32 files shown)


src/routes.jsx
```diff
@@ -232,7 +232,7 @@ function routes(app) {
     props.app = app;
     props.api = app.api;
     props.seed = ctx.seed;
-    props.random = ctx.random;
+    global.random = ctx.random;
```
ajacksified: Instead of redefining it every time `defaultProps` is called (every route), could it maybe be defined in `client.es6.js` and `server.jsx`, or even in `app-mixin` (if on mixin, pass the seed in as a config value)?

Might make sense to move other common props, like `api`, `app`, and `config` to globals as well, treated as read-only singletons.

ajacksified:
:nail_care:

danehansen:
:haircut: @ajacksified

ajacksified:
`global.random` also needs to be set on the server.

ajacksified:
:nail_care:

danehansen:
:haircut: @ajacksified 
p.s. snuck a quick css fix in there too

ajacksified:
:fish:

PR #209, opened 2015-06-11
Fix animated gif play icon bug

danehansen:


ajacksified:
As a side effect, the blue placeholder now appears before images are loaded, instead of the grey box. Is that intentional?

:nail_care:

danehansen:
@ajacksified woops. good catch. :haircut:

ajacksified:
:fish: 

Hooray for nuking more listingcontent code.

PR #210, opened 2015-06-11
Put back accidentally deleted ListingDropdown items

danehansen:


ajacksified:
Hide works, but does not remove the element from the dom - throws an error instead. Save looks good.

:nail_care:

danehansen:
:haircut: @ajacksified

ajacksified:
:fish:, pending whatever your 403 issue is

PR #213, opened 2015-06-11
Disable compactTest when finished.

ajacksified:
Sneak in an Android 2.3 fix too.


curioussavage:
:fish:

PR #220, opened 2015-06-16
Ad: pass in loaded state to rendered link

curioussavage:
:eyeglasses: @ajacksified or @danehansen


danehansen:
as long as we are ok with loading that ad in before its scrolled into view, then :fish:

curioussavage:
looks like dave is working on a broader fix with his dfp work

PR #225, opened 2015-06-23
Update score when voting on comments

ajacksified:
Score wasn't being updated when upvoting/downvoting comments. This fixes that and some issues with the localScore calculation in Vote.jsx.


danehansen:
:fish:

PR #226, opened 2015-06-23
Render tracking pixel on client with referrer domain

ajacksified:
Avoid cache issues by only rendering the tracking pixel on the client side. Also append the referrer's domain.


src/views/components/TrackingPixel.jsx
```diff
@@ -1,15 +1,34 @@
...
+      loaded: false,
+    };
   }
 
-  shouldComponentUpdate (nextProps, nextState) {
-    return false;
+  componentDidMount () {
+    var state = {
+      loaded: true,
+    }
```
curioussavage: nit: semicolon

PR #229, opened 2015-06-24
Fix ad images not always loading due to race condition

dwick:
:eyeglasses: @curioussavage 

Can you test this on staging please?

### DIFF
--- a/src/views/components/ListingContent.jsx
+++ b/src/views/components/ListingContent.jsx
@@ -217,7 +217,11 @@ class ListingContent extends React.Component {
         }
       }
       if (src) {
-        style.backgroundImage = 'url(' + src + ')';
+        var a = document.createElement('a');
+        a.href = src;
+        a.protocol = location.protocol;
+
+        style.backgroundImage = 'url(' + a.href + ')';
       }
     }
     var aspectRatio = this._aspectRatio();
--- a/src/views/components/ListingList.jsx
+++ b/src/views/components/ListingList.jsx
@@ -15,16 +15,16 @@ class ListingList extends React.Component {
       adLocation: Math.min(AD_LOCATION, props.listings.length)
     };
 
-    this._scroll = this._scroll.bind(this);
+    this._lazyLoad = this._lazyLoad.bind(this);
     this._resize = this._resize.bind(this);
   }
 
   componentDidMount() {
     this._hasListeners = true;
-    this.props.app.on(constants.SCROLL, this._scroll);
-    this.props.app.on(constants.RESIZE, this._scroll);
+    this.props.app.on(constants.SCROLL, this._lazyLoad);
+    this.props.app.on(constants.RESIZE, this._lazyLoad);
     this.props.app.on(constants.RESIZE, this._resize);
-    this._scroll();
+    this._lazyLoad();
     this._resize();
   }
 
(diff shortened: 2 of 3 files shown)


dwick:
:eyeglasses: @ajacksified for `ListingList` changes

src/views/components/ListingList.jsx
```diff
@@ -33,28 +33,34 @@ class ListingList extends React.Component {
     this.props.app.off(constants.RESIZE, this._resize);
   }
 
+  _getWindowHeight () {
+    return window.innerHeight * 2;
```
ajacksified: I like moving it to a function, but calling it `windowheight` and returning `windowheight*2` doesn't align.

src/views/components/ListingList.jsx
```diff
@@ -124,14 +135,13 @@ class ListingList extends React.Component {
...
 
     // If ads are enabled, splice an ad into the listings.
-    if (this.props.showAds && listings.length) {
-      ad = this.buildAd();
-      listings.splice(this.state.adLocation, 0, ad);
+    if (props.showAds && listings.length) {
+      listings.splice(this.state.adLocation, 0, this.buildAd());
     }
 
-    this.length = listings.length;
```
danehansen: i created an instance variable of length here so that length doesn't have to be recalculated every iteration of the for loop each time _scroll is called

dwick: http://stackoverflow.com/a/5752991/704286

danehansen: that is good news!

curioussavage:
I can push it up to staging but so far locally I'm not seeing ad images

curioussavage:
thumbnail is getting deleted when you get a response and build the link but in `componentDidMount` in Ad.jsx when you turn the json version back into a link instance there is no `thumbnail` or `mobile_ad_url` prop so it ends up with the default img

curioussavage:
:nail_care:

curioussavage:
is there a reason why we are converting models to json before returning them and the just instantiating them again? I have wondered that while looking at other things. @ajacksified ?

ajacksified:
@curioussavage when snoode was first created, it assumed we could use [proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy); since it can't, we turn it into JSON before returning them, otherwise all of the views have to use `model.get('propertyName')` instead of `model.propertyName`, which is annoying. So, it's turned into JSON first to make it less annoying to use.

src/views/components/ListingContent.jsx
```diff
@@ -215,7 +215,11 @@ class ListingContent extends React.Component {
         }
       }
       if (src) {
-        style.backgroundImage = 'url(' + src + ')';
+        var a = document.createElement('a');
+        a.href = src;
+        a.protocol = location.protocol;
```
dwick: images have a protocol and it's not always https

dwick: can break this into a separate commit if that's wanted

dwick:
:haircut: this is up on staging too https://m.staging.wireddit.com

curioussavage:
:fish:

ajacksified:
:fish:

PR #230, opened 2015-06-24
Remove cruft

dwick:
Second block was never hit.  Also, no need to add the source image
to the `find` call if that is what is fallen back to if no good
alternatives exist.

:eyeglasses: @ajacksified


ajacksified:
:fish:

dwick:
6340b32a67707cdb5ab0f977006b0de7499ab9e2

PR #231, opened 2015-06-24
Editing: implement editing for comments and self posts

curioussavage:
:eyeglasses: @danehansen @ajacksified 

I think this also fixed links on the coments page being switched to compact when the user toggles while on that page.


ajacksified:
:fish: post-snoode-bump!

ajacksified:
adding a link to https://github.com/reddit/snoode/pull/35 so github notifies both

PR #233, opened 2015-06-24
Enabled compact view image expanding

danehansen:


ajacksified:
:fish:, although the sidenav-li stuff probably should be part as a separate pr, or at least a separate commit.

PR #234, opened 2015-06-24
Restore comment transition

danehansen:
:eyeglasses: @ajacksified I stumbled across a hack to prevent a browser from repainting everything below an element that has a height change. Try this out to see if its smooth enough for you. Seems to be a pretty significant difference.


ajacksified:
It's _decent_ on my relatively new Android. Have you tested it on the older test devices?

PR #235, opened 2015-06-25
Adjust max subreddit title display length

curioussavage:
:eyeglasses: @danehansen  or @ajacksified 

had to adjust this some more since some subs were not being caught due to the minimum length


danehansen:
i think we can go back and do a fade out fix for this, like what we do elsewhere on the post, but for now, :fish:

PR #236, opened 2015-06-25
Replace subreddit shortening with css solition

danehansen:
:eyeglasses: @curioussavage

### DIFF
--- a/src/views/components/TopNav.jsx
+++ b/src/views/components/TopNav.jsx
@@ -16,14 +16,8 @@ import InfoIcon from '../components/icons/InfoIcon';
 import TextIcon from '../components/icons/TextIcon';
 import SnooIcon from '../components/icons/SnooIcon';
 
-function shorten(text, len) {
-  if (text.length > 9) {
-    text = text.substr(2, Math.min(8, len-2)) + '…';
-  } else {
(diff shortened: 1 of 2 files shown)


src/views/components/TopNav.jsx
```diff
@@ -87,7 +81,7 @@ class TopNav extends React.Component {
     if (subredditName) {
       var breadcrumbLink = '/' + this.state.subredditName;
       var breadcrumbContents = subredditName;
-      currentSub = '?subreddit=' + subredditName;
+      currentSub = '?subreddit=' + this.state.subredditName;
```
curioussavage: currentSub shouldn't have the `r/` prefix.

maybe we could just get the substring inline and axe `shorten` line 79 should be removed too.

curioussavage:
:haircut:

curioussavage:
uggh.. :nail_care:

danehansen:
@curioussavage :haircut: Good catch.

curioussavage:
:fish:

curioussavage:
actually sorry one little nit. on `shorten` shouldn't take a second argument.

curioussavage:
re-:fish:

PR #237, opened 2015-06-26
Create rainbow snoo

danehansen:


ajacksified:
:fish: - confirm with Tavish that it should be on all subs and not just the
homepage

PR #238, opened 2015-06-26
Fix nsfw image bug

danehansen:

### DIFF
--- a/assets/less/components/ListingContent.less
+++ b/assets/less/components/ListingContent.less
@@ -91,6 +91,10 @@
     }
   }
 
+    .ListingContent-image-img {
+      width: 100%;
+    }
+
     .ListingContent-nsfw {
       .display-flex;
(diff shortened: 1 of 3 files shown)


ajacksified:
Two issues I found:

* Viewing comments shouldn't expand an nsfw image; it should still show the preview (https://m.reddit.com/r/ajacksified/comments/30ye35/nsfw_an_nsfw_post/)
* If you open an nsfw image that links directly to an image (it has no `placeholder` object), it uses the 16x9 aspect ratio, which could cut off the image. It may be best to not try to embed an image if we don't have an aspect ratio, rather than trying to embed the source directly.

:nail_care: - feel free to fix the second as a separate PR if you want

danehansen:
:haircut: @ajacksified

ajacksified:
:fish:

PR #240, opened 2015-06-30
Remove rainbow snoo

danehansen:


ajacksified:
:fish: - might end up needing some way to abstract out snoo styles eventually, but this works

PR #243, opened 2015-06-30
Dfp

dwick:

Might need to do some more with tagging, but this should work for the mean time.

### DIFF
--- a/src/config.es6.js
+++ b/src/config.es6.js
@@ -24,7 +24,7 @@ var config = {
 
   loginPath: process.env.LOGIN_PATH || '/oauth2/login',
 
-  adsPath: process.env.ADS_PATH ||  'https://www.reddit.com/api/request_promo.json',
+  adsSlot: process.env.ADS_PATH || '/32173961/mw_card_test_1',
 };
 
 export default config;
--- a/src/config.es6.js
+++ b/src/config.es6.js
@@ -24,7 +24,7 @@ var config = {
 
   loginPath: process.env.LOGIN_PATH || '/oauth2/login',
 
-  adsSlot: process.env.ADS_PATH || '/32173961/mw_card_test_1',
+  adsSlot: process.env.ADS_SLOT || '/32173961/mw_card_test_1',
 };
 
 export default config;
--- a/src/lib/frames.es6.js
+++ b/src/lib/frames.es6.js
@@ -1,11 +1,9 @@
 'use strict';
 
-var WILDCARD = '*';
-var ALLOW_WILDCARD = '.*';
-var RE_WILDCARD = /\*/;
-var DEFAULT_MESSAGE_NAMESPACE = '.postMessage';
-var DEFAULT_POSTMESSAGE_OPTIONS = {
-  targetOrigin: WILDCARD,
+const ALLOW_WILDCARD = '.*';
+const DEFAULT_MESSAGE_NAMESPACE = '.postMessage';
+const DEFAULT_POSTMESSAGE_OPTIONS = {
+  targetOrigin: '*',
 };
 
 var allowedOrigins = [ALLOW_WILDCARD];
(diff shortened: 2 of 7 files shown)


src/lib/frames.es6.js
```diff
@@ -0,0 +1,210 @@
...
+    customEvent.source = e.source;
+
+    global.dispatchEvent(customEvent);
+  } catch (x) {}
+}
+
+
+
+function compileOriginRegExp(origins) {
+  return new RegExp('^http(s)?:\\/\\/' + origins.join('|'), 'i');
```
JordanMilne: Should this have a `$` anchor at the end of the regex as well?

dwick: good catch

src/config.es6.js
```diff
@@ -24,7 +24,7 @@ var config = {
 
   loginPath: process.env.LOGIN_PATH || '/oauth2/login',
 
-  adsPath: process.env.ADS_PATH ||  'https://www.reddit.com/api/request_promo.json',
+  adsSlot: process.env.ADS_PATH || '/32173961/mw_card_test_1',
```
ajacksified: intentional?

src/lib/frames.es6.js
```diff
@@ -0,0 +1,210 @@
+'use strict';
+
+var WILDCARD = '*';
+var ALLOW_WILDCARD = '.*';
+var RE_WILDCARD = /\*/;
+var DEFAULT_MESSAGE_NAMESPACE = '.postMessage';
+var DEFAULT_POSTMESSAGE_OPTIONS = {
```
ajacksified: might as well make all these `const` instead of `var`

src/lib/frames.es6.js
```diff
@@ -0,0 +1,210 @@
...
+   */
+  receiveMessage: function (source, type, callback, context) {
+    if (typeof source === 'string') {
+      context = callback;
+      callback = type;
+      type = source;
+      source = null;
+    }
+
+    if (!/\..+$/.test(type)) {
```
ajacksified: all of the other regexes are variables; put this one in a const too?

src/views/components/Ad.jsx
```diff
@@ -4,20 +4,92 @@ import superagent from 'superagent';
...
+        googletag.pubads().enableSingleRequest();
+        googletag.enableServices();
+      });
+
+      googletag.cmd.push(function() {
+        googletag.display(adId);
+      });
+
+      setTimeout(() => {
+        reject('timeout');
```
ajacksified: can you be more specific? `'ad creative load timeout'` or something.

ajacksified:
A few nits / questions, but otherwise good to go. :nail_care:

src/views/components/Ad.jsx
```diff
@@ -4,20 +4,92 @@ import superagent from 'superagent';
...
+        '/api/dfp/link.json';
+    var headers = {};
+
+    if (props.token) {
+      headers.authorization = 'bearer ' + props.token;
+    };
+
+    return new Promise((resolve, reject) => {
+      superagent
+        .post(url)
```
JordanMilne: If the `websafe_json` encoding will mess things up here you should add a `raw_json=1` parameter to the URL or POSTDATA.

dwick: :+1:

src/lib/frames.es6.js
dwick: i removed the regexes that weren't reused from the constants

dwick:
:haircut:

ajacksified:
Looks good; :fish: pending testing on staging and DFP-readiness on the reddit.com side

dwick:
not fantastic enough.

ajacksified:
No, @dwick. The problem was that it was *too* fantastic.

PR #244, opened 2015-07-01
Nuke compacttest; unset `false` compact cookie

ajacksified:
Compact should be either `true` or `undefined` to make CF logic easier.


curioussavage:
:fish:

Issue #247, opened 2015-07-02
Loading performance


paulirish:
Been meaning to do a performance audit of the new mobile site. I'm going to drop it right here into the ticket. I may add comments as I go along.

## Initial Test: Loading.

Test setup: 2013 Moto X. On 4G (HSPA+). Chrome Dev Channel.

1. Load reddit.com 
2. Tap on the big blue link to load the mobile site. 
3. Wait until the site feels loaded.

Expectation: Site should load and give me the frontpage items and images in < 5 seconds.
Result: Site came up with items and image placeholders within 3 seconds. It took *45 seconds** to show images.

Here's the overview of Timeline and and network activity.

Timeline. (click to zoom)
![image](https://cloud.githubusercontent.com/assets/39191/8490159/43ff74c6-20d9-11e5-879b-2218becd6a31.png)

Network waterfall, annotated.
![image](https://cloud.githubusercontent.com/assets/39191/8490198/023372bc-20da-11e5-95eb-5d08243a6a07.png)

…

paulirish:
During all that time, the page is scrolling just fine. It feels probably interactive, and as a user I'm wondering why the images are taking so long. 

However if during that time i click on a link to read comments I will wait up to 16 seconds for that tap to be handled.

screenshot of the kind of thing i'm looking at:
<a href="https://cloud.githubusercontent.com/assets/39191/8490267/d92b9f7e-20da-11e5-9384-0248c7d74db4.png"><img src="https://cloud.githubusercontent.com/assets/39191/8490267/d92b9f7e-20da-11e5-9384-0248c7d74db4.png" height=400></a>

So.. the next question is.. **what's the holdup?**
Let's find out. :)

paulirish:
Basically, a megabyte of javascript. 
Indeed, client.min.js is 1.1MB of minified code.

While scrolling is responsive, the main thread is jam packed, having a party.

Here's a look at what happens from start to finish of that 45 seconds:

![image](https://cloud.githubusercontent.com/assets/39191/8490300/4331eb62-20db-11e5-8d83-7b3743cefee9.png)

We can break up all that work into 4 sections for better inspection: 
![image](https://cloud.githubusercontent.com/assets/39191/8490466/445b1e3e-20de-11e5-8924-e7e4dfb90972.png)


From a quick glance here's what each act contains:

* **Act 1: Skeleton** (4 seconds): Initial HTML, css, JS downloaded. Skeleton DOM is built. Everything is laid out. 
[...]

PR #249, opened 2015-07-06
Performance

ajacksified:
First pass at performance optimizations; low-hanging-fruit edition. Things are still slow, but they're significantly faster.

Notes are in the individual commits. highlights:

* Replaced most of the `snooicon`s with css, and many other icons;
* Fixed some bugs causing re-rendering;
* Not rendering large or common hidden elements until they're made visible;

Before and after graph porn:


ajacksified:
Notifying #247. (This PR does _not_ address all the issues in that PR, just many of the easy ones.)

curioussavage:
Can we deploy this to staging?

ajacksified:
Sure can, I'll deploy to staging this morning.

curioussavage:
Here is a timeline from my machine for kicks. Time for scripting/rendering seems opposite. weird. this is on desktop chrome :/ no android device around.

performance

ajacksified:
Opposite? Seems the same to me (of course, everything is smaller.) My times are from my Nexus 6.

curioussavage:
ah. yeah I was reading it wrong. Turns out reading the unit is important :P

src/views/components/Ad.jsx
```diff
@@ -16,7 +20,7 @@ class Ad extends React.Component {
       unavailable: false,
     };
 
-    this._scroll = _.throttle(this._scroll.bind(this), 100);
+    this._scroll = throttle(this._scroll.bind(this), 100);
```
danehansen: We can set this listener to the app's scroll emitter, which is already throttled at 100 rather than implementing a throttle again here.

src/views/components/BaseComponent.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+class BaseComponent extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return (nextProps !== this.props || nextState !== this.state);
+  }
```
danehansen: Does this always work? Like if a new state or props comes through is it always a new object even though only one or so props are changed?

src/views/components/ListingDropdown.jsx
```diff
@@ -72,8 +66,8 @@ class ListingDropdown extends React.Component {
 
       reportLink = (
         <li className='Dropdown-li'>
-          <MobileButton className='Dropdown-button' onClick={ this._onReportClick }>
-            <FlagIcon/>
+          <MobileButton touch={ props.touch } className='Dropdown-button' onClick={ this._onReportClick }>
```
danehansen: This seems cumbersome to pass the touch property into everything through props. Is this really more performant than having a reference to Utils.touch() inside of the components that use it? And at the very least we could throw something like that on globals.

danehansen:
Overall, I dont think ripping out animations and SVGs is correct the solution to immediately jump to. We could have a plain text site with no css and it would look somewhat as fun as the regular reddit.com and we are still left with an extremely slow API and large image load. There are plenty of things we can do to optimize based on the awesome findings of these guys over the last few days without killing everything thats fun in our site. I will be back in tomorrow to help with this.

src/views/pages/submit.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import _ from 'lodash';
+import throttle from 'lodash/function/throttle';
```
danehansen: I'm not doubting that this is more efficient but I am wondering why. It would become a hassle on files if we were using many methods from a library having to import each one. If all these methods still live on the one instance of that library, isn't it all basically the same? And our minification should be leaving out any unused methods on those libraries, right?

src/views/components/BaseComponent.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+class BaseComponent extends React.Component {
```
danehansen: This is awesome, we probably should have had this from the beginning and seems like we could put more cool stuff in here. I dont know what yet though.

src/views/components/Ad.jsx
```diff
@@ -1,13 +1,17 @@
-import _ from 'lodash';
+import throttle from 'lodash/function/throttle';
+import extend from 'lodash/object/assign';
```
dwick: Could be replaced with `Object.assign`

ajacksified:
I didn't kill everything fun in the site - I left most of the visible animations in place; the logo, the loading icon, the menu icon, some of the things that change on click (like save and hide.) However, the single largest impact to the load times, and performance on phones, is the huge amount of SVGs being loaded (like the list of 100 animated snoo heads for the subscription list.)  Animations like the comments or settings icon don't really make sense, when you don't see it because your finger's over it, and it just redirects to a new page immediately anyway.

The profiling shows that loading the JS, including all of the animations before this PR, adds 40 seconds onto the ~4 second api / image load time; the majority of work is react mount / re-rendering. These fixes cut it by more than half. We should focus on what gives us the most impact.

I agree that we should keep the site fun, but it also must remain practical. Use animation where it makes a difference, but animating every icon isn't feasible for mobile devices. We should focus on getting load times back to a reasonable number, and then figure out how we can add some of this back, by doing things like progressive mounting on scroll.

src/views/components/Ad.jsx
ajacksified: Yes, I didn't bother refactoring this out yet because it's used in a few other places, here and in dependencies, anyway.

src/views/components/BaseComponent.jsx
ajacksified: Yeah, I think we could move a *lot* of the boilerplate here as we find common methods in components. Probably makes sense to break out a `BasePageComponent` that handles data loading, too.

src/views/pages/submit.jsx
ajacksified: Browserify can be a bit smarter about pulling in the exact code it needs this way; if it ends up being too cumbersome, we can revisit it, but I suspect that we can use ES6 features to replace a lot of our lodash use anyway (other than throttle/debounce.)

src/views/components/ListingDropdown.jsx
ajacksified: The components were calling out to Utils.touch on mount, which in cases like the icons, was happening several hundred times per page render. This would be a good use for globals.

src/views/components/Ad.jsx
ajacksified: :+1:

src/views/components/BaseComponent.jsx
ajacksified: This works for all of the testing I've done; and you can override it for a component if you need to do a deep comparison for some reason.

src/views/components/ListingDropdown.jsx
danehansen: Should we make globals, or an alternative to it, that would be imported at the tops of files? Reason for this would be that we can have things like app on it, which we do not want to be actually global for security reasons, be able to be easily accessed from anywhere without having to cascade it down through many components' props. Downside is a single import statement at the top of any file that would want to access something from this imported global object.

danehansen:
I understand all that, that it's a balance. But I would rather have some of this broken out. Such as replacing all the svgs. I have some of my own ideas I would like to implement to attempt to streamline this. And taking the compact view transition out has nothing to do with the kind of optimizing this PR is dealing with and I think that is a good feature.

ajacksified:
Removing the transition was part of an unfinished attempt at figuring out how to replace gsap / tweenlite, which is about 1/3 of our total javascript; I'll back that out and we can figure that out later.

src/views/components/ListingDropdown.jsx
ajacksified: Yeah, I think that at this point, we could clean up a *lot* of code by having some kind of app `import` that exports a singleton, so we no longer have to pass down `api`, `app`, and the big list of configuration options in `routes.jsx`. It would, like you said, require the import statement at the top of files, but I think that's much nicer than cascading `app` to every child component.

spladug: I'm sorry for jumping in where I'm eminently unqualified, but I'm curious about moving in the direction of globals to save passing commonly-used objects around; a lot of devs on the r2 codebase have specifically called out that aspect of r2/Pylons as a negative feature (especially for testability) and are actively trying to move away from it.  What do you think of the downsides of going global/singleton on the commonly used objects?

(I'm happy to take this to another venue -- don't mean to hijack your PR comments -- sorry!)

danehansen:
these guys reported the greensock stuff at about 100kb, not 1/3 of our 1.1mb. looking over which aspects of the greensock library i am utilizing, their combined minified size is closer to 90kb, which is not too different than 100kb, but its different enough that it concerns me that more than necessary files are being packaged into our build or that if it is reminifying the already minified files that its not doing as good of a job.

src/views/components/SideNav.jsx
```diff
@@ -95,25 +98,25 @@ class SideNav extends React.Component {
...
+          <MobileButton touch={ this.props.touch } className='SideNav-button' onClick={this._onTwirlyClick.bind(this, 'user')}>
             <TwirlyIcon altered={isUser}/>
             <span className='SideNav-text'>{ user.name }</span>
           </MobileButton>
           <ul ref='user' className='SideNav-ul list-unstyled'>
             <li className='SideNav-li'>
-              <MobileButton className='SideNav-button' href={ `/u/${user.name}` }>
-                <SnooIcon/>
+              <MobileButton touch={ this.props.touch } className='SideNav-button' href={ `/u/${user.name}` }>
+                <span className='icon-snoo-circled icon-large' />
```
curioussavage: These icons are right up to the text. Needs margin between

ajacksified:
:haircut:

curioussavage:
Don't see anything else. :fish:  for me.

src/views/components/ListingDropdown.jsx
```diff
@@ -3,19 +3,13 @@ import constants from '../../constants';
 
 import { models } from 'snoode';
 
+import BaseComponent from './BaseComponent';
 import SeashellsDropdown from '../components/SeashellsDropdown';
-import CommentIcon from '../components/icons/CommentIcon';
 import MobileButton from '../components/MobileButton';
-import SnooIcon from '../components/icons/SnooIcon';
-import InfoIcon from '../components/icons/InfoIcon';
-import FlagIcon from '../components/icons/FlagIcon';
```
danehansen: I would say probably leave FlagIcon in here, since it doesn't link to a different page. On the inverse of that, the settings one could probably go since it causes the component to immediately dismount.

danehansen:
Should we merge in my globals object PR first so that we aren't merging in all these touch props just to immediately remove them later?

ajacksified:
Yup, good idea.

src/views/components/ListingDropdown.jsx
ajacksified: done.

dwick:
:tada:

PR #250, opened 2015-07-06
Replace env variables with strings

dwick:
Example:

```js
// This:
if (process.env.NODE_ENV === 'development') {
  console.log('dev!');
}

// Turns into this:
if ('production' === 'development') {
  console.log('dev!');
}
```

Uglify is able to strip the latter via dead code removal, but not
the former.

See https://github.com/reddit/reddit-mobile/issues/247#issuecomment-118551305 and https://facebook.github.io/react/downloads.html#npm for more info.

:eyeglasses: @ajacksified

### DIFF
--- a/buildTasks/js.js
+++ b/buildTasks/js.js
@@ -13,6 +13,7 @@ var streamqueue = require('streamqueue');
 var browserify = require('browserify');
 var watchify = require('watchify');
 var babelify = require('babelify');
+var envify = require('envify/custom');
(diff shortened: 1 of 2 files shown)


dwick:
results in a little over 50kb of savings

ajacksified:
:fish:

jdalton:
Holy :hankey:! 50kb in savings!

PR #252, opened 2015-07-06
Pass oauth origins to client

dwick:
:eyeglasses: @ajacksified this allows for non-default api endpoints (using a local dev version of reddit)


ajacksified:
:fish:

PR #257, opened 2015-07-09
Remove manifest.json

ajacksified:
It doesn't play well with tabs and can be awkward to use for a very external-link-heavy application like this.

Jira 188

:eyeglasses: @danehansen

### DIFF
--- a/src/serverRoutes.es6.js
+++ b/src/serverRoutes.es6.js
@@ -31,54 +31,6 @@ Allow: /
     `;
   });
 
-  router.get('/manifest.json', function * () {
-    this.body = {
-      name: 'reddit: the front page of the internet',
-      icons: [
-        {
-          src: `${app.config.assetPath}/favicon/64x64.png`,
-          sizes: '64x64',
-          type: 'image/png'
-        },
-        {
-          src: `${app.config.assetPath}/favicon/76x76.png`,
-          sizes: '76x76',
-          type: 'image/png'
-        },
-        {
-          src: `${app.config.assetPath}/favicon/120x120.png`,
-          sizes: '120x120',
-          type: 'image/png'
-        },
-        {
-          src: `${app.config.assetPath}/favicon/128x128.png`,
-          sizes: '128x128',
-          type: 'image/png'
-        },
-        {
-          src: `${app.config.assetPath}/favicon/152x152.png`,
-          sizes: '152x152',
-          type: 'image/png'
-        },
-        {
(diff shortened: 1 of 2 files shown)


danehansen:
:fish:

tomitm:
@ajacksified Could you go into detail on what sorts of issues you ran into with this?

As far as I'm aware, using a `standalone` should open a link in a new browser tab, while still allowing you to switch back to Reddit as if it were an app. Did you find that really awkward? How was it different from a native app behaviour?

Also, keep in mind that a lack of a manifest will prevent Chrome (and other browsers eventually) from [suggesting](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android) users to add to homescreen. However, the criteria doesn't actually require you to set `standalone` if it's problematic for Reddit, so rather than fully removing it, you could either omit `display` or use the value `browser` or maybe even `minimal-ui`.

ajacksified:
@tomitm - it was awkward, precisely because it acted like a native app, despite the user expecting it to work like a web app. It also (iirc, this was a while ago) hid browser functionality like the URL bar (frequently used to copy the current URL for sharing) and other browser tools. This makes sense, since the purpose of the manifest is to make it appy... But it breaks how users expect to use a web app. We have an  Android and iOS app in beta now that have been designed to surface sharing natively, but it's lower priority here because that's basically reimplementing what the browser already provides.

If there's a way to build a manifest so that it doesn't hide browser ui, then maybe it would make sense; I just haven't gone back to look yet.

tomitm:
Interesting! I tend to deliberately add to homescreen expecting an app-like behaviour, not a browser-UI, hence my initial confusion.

The lack of URL for sharing is a really good point, though. I realise that's something critical for Reddit, as I heavily do that myself.

There's two ways around that:

1. Forego the app-like experience, using either `browser` or `minimal-ui` as your `display` values
2. Use the [`display-mode`](https://developers.google.com/web/updates/2015/10/display-mode) CSS media query to know when you're in `standalone` mode (vs being in the browser) so you can show some sharing UI (like you would in a mobile app), like you would in a native app.

It would be nice if Chrome let you pick though, like it does on desktop with the "Open as window" option. I just tested `minimal-ui` and I'm not sure if Chrome implements it yet, as it seems to behave just like `browser`. If you're wondering, [the spec](https://www.w3.org/TR/appmanifest/#display-modes) seems to indicate it should behave like `fullscreen` (which probably isn't preferable) but providing some browser UI.

PR #265, opened 2015-07-13
Messaging; all except compose

ajacksified:

### DIFF
--- a/src/views/components/MessagePreview.jsx
+++ b/src/views/components/MessagePreview.jsx
@@ -1,0 +1,242 @@
+import React from 'react';
+import moment from 'moment';
+import process from 'reddit-text-js';
+import { models } from 'snoode';
+
+import Inbox from '../components/Inbox';
+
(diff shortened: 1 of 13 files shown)


src/views/components/MessagePreview.jsx
```diff
@@ -0,0 +1,242 @@
...
+
+    this.state = {
+      showReply: false,
+    };
+
+    this._onReplyClick = this._onReplyClick.bind(this);
+    this._onReplySubmit = this._onReplySubmit.bind(this);
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
```
curioussavage: probably goes without saying but with the baseComponent #249  this should be removed.

ajacksified: Yup, whichever one gets in first will need to make some changes after the rebase

curioussavage:
:fish:

PR #267, opened 2015-07-14
Use a small cookie to do cache poisoning detection.

ajacksified:
:eyeglasses: @JordanMilne


assets/js/client.es6.js
```diff
@@ -45,6 +46,86 @@ function onLoad(fn) {
...
+    });
+  }
+}
+
+function checkCanaryCookie(config) {
+  var canary = cookies.get(CANARY_COOKIE_NAME);
+
+  // There was a canary from the page, which sould only appear for logged-in
+  // users. So check it against our canary to make sure they match.
+  if (config.canary && config.canary !== canary) {
```
JordanMilne: need an `&& canary` here too, the canary cookie might not be set yet.

assets/js/client.es6.js
```diff
@@ -45,6 +46,86 @@ function onLoad(fn) {
...
+}
+
+var csrfRegex = /csrf-token" content="([^"]+)/;
+
+function logCachePoisoning(config) {
+  superagent
+    .get(window.location.href)
+    .end(function (err, res) {
+      var details = {
+        poisoner_name: config.user.name,
```
JordanMilne: need `victim_canary` here too, the server double checks that they're actually different

assets/js/client.es6.js
```diff
@@ -45,6 +46,86 @@ function onLoad(fn) {
...
+        .send(details)
+        .set({
+          'X-Loggit': true,
+        })
+        .end(function(){});
+    });
+}
+
+function checkCachePoisoning(config) {
+  if (!checkCanaryCookie(config)) {
```
JordanMilne: Does this make sure that this is a logged-in page? We don't care if loggedout pages end up in caches

src/routes.jsx
```diff
@@ -218,6 +218,7 @@ function routes(app) {
       origin: app.getConfig('origin'),
       propertyId: app.getConfig('googleAnalyticsId'),
       userAgent: ctx.userAgent,
+      canary: ctx.canary,
```
JordanMilne: Same as above, does this only put this on the page if logged in?

src/routes.jsx
```diff
@@ -218,6 +218,7 @@ function routes(app) {
       origin: app.getConfig('origin'),
       propertyId: app.getConfig('googleAnalyticsId'),
       userAgent: ctx.userAgent,
```
JordanMilne: You'll also need a `cachePolicy` attr to differentiate between different cache policies used on the same app. 'loggedin_mweb' if loggedin and 'loggedout_mweb' if loggedout are what I'm relying on for now.

src/routes.jsx
ajacksified: with https://github.com/ajacksified/reddit-mobile/commit/ec6c8407e228f14a03b4f4b735e3a28abb041b53 - yes.

assets/js/client.es6.js
ajacksified: `config.canary` will be empty if you're logged in, so it won't run the check (`if (canary && config.canary && config.canary !== canary)` will fail)

assets/js/client.es6.js
```diff
@@ -45,6 +46,87 @@ function onLoad(fn) {
...
+
+var csrfRegex = /csrf-token" content="([^"]+)/;
+
+function logCachePoisoning(config) {
+  superagent
+    .get(window.location.href)
+    .end(function (err, res) {
+      var details = {
+        poisoner_name: config.user.name,
+        poisoner_id: config.user.id,
```
JordanMilne: You'll need to MAC some of this data, include it in the page, and then send it off with any reports, see  how `make_poisoning_report_mac` is used in r2

src/routes.jsx
JordanMilne: Oh, and also a `renderTime` property with a JS timestamp of when the page was rendered. That way we know how long the page has been sitting around in the cache.

JordanMilne:
Definitely a lot cleaner than what I was doing! Just a couple things :nail_care:

ajacksified:
:haircut:

src/views/layouts/DefaultLayout.jsx
```diff
@@ -1,8 +1,38 @@
...
+  hmac.setEncoding('hex');
+  hmac.write(data);
+  hmac.end();
+
+  return hmac.read();
+}
+
+function generateCanaryMac(props, secret) {
+  var macData = [
+    props.user.name,
```
JordanMilne: the canary itself should be the first thing in `macData`

ajacksified:
:haircut:

JordanMilne:
:fish: lgtm, but deploy's blocked until I fix the CORS issues so you can send the report :sleeping:

JordanMilne:
We _should_ be good to go ahead with this now

PR #271, opened 2015-07-15
Pass loid on call to request_promo

zeantsoi:
:eyeglasses: @ajacksified @dwick 

Passes the value of the `loid` cookie on the call to `request_promo`. If the cookie is unset, [the validator logic](https://github.com/reddit/reddit-plugin-adzerk/pull/68/files#diff-8ff7bdcffc2cdf19a06e2d539306378aR619) will accept a value of `undefined` and default the value to `None` on the server side.

This PR should not be merged or rebased until [reddit-public #3889](https://github.com/reddit/reddit-public/pull/3889) is deployed around 7/16.

### DIFF
--- a/src/views/components/Ad.jsx
+++ b/src/views/components/Ad.jsx
@@ -175,6 +175,7 @@ class Ad extends BaseComponent {
 Ad.propTypes = {
   afterLoad: React.PropTypes.func.isRequired,
   compact: React.PropTypes.bool.isRequired,
+  user: React.PropTypes.object,
 };
 
 export default Ad;
--- a/src/views/components/Ad.jsx
+++ b/src/views/components/Ad.jsx
@@ -60,14 +60,19 @@ class Ad extends BaseComponent {
       srnames = ' reddit.com';
     }
 
+    var postData = {
+      srnames: srnames,
+      is_mobile_web: true,
+    };
+    // If user is not logged in, send the loid in the promo request
+    if (!this.props.user) {
+      postData['loid'] = globals().app.state.loid;
+    }
+
     return new Promise((resolve, reject) => {
       superagent.post(this.props.adsPath)
         .type('form')
-        .send({
-          srnames: srnames,
-          is_mobile_web: true,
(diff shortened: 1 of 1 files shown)


ajacksified:
The `loid` should only be passed if the user is not logged in. :nail_care:

zeantsoi:
In the event that `loid` is `undefined`, the validation will do the handling server side.

Do you suggest we add conditional logic so that the key-value pair is not sent at all?

ajacksified:
Correct; `props.user` should be passed to `<Ad>`, and if `user` is defined, `loid` should not be sent at all.

zeantsoi:
:haircut:

src/views/components/Ad.jsx
```diff
@@ -60,13 +60,19 @@ class Ad extends BaseComponent {
       srnames = ' reddit.com';
     }
 
+    var postData = {
+      srnames: srnames,
+      is_mobile_web: true,
+    };
+    // If user is not logged in, send the loid in the promo request
+    if (!this.props.user) {
+      postData['loid'] = globals().app.state.loid;
```
ajacksified: use `postData.loid` instead of `['loid']` - and you'll need to send `props.user` into `<Ad` where it's used.

ajacksified:
:nail_care:

zeantsoi:
:haircut: 

Dunno what to say re: passing the `user` prop... it's already accessible from within the component.

dwick:
Can you add it to `propTypes` if it is indeed there/required?

see: https://github.com/zeantsoi/reddit-mobile/blob/4a403d7cf1f1343317cad2e204706ed6b4e4aa7a/src/views/components/Ad.jsx#L175

dwick:
:haircut:

zeantsoi:
:haircut:

zeantsoi:
:eyeglasses: @danehansen

danehansen:
:fish:

zeantsoi:
057b9deb818b18476da07cd8bb1efdc7ad7e2947

PR #273, opened 2015-07-17
Fix for subreddit about page

curioussavage:
:eyeglasses: @ajacksified 

Quick fix for this page, an undefined `props` var was being referenced.


ajacksified:
:fish:

PR #274, opened 2015-07-17
Ad.jsx: Send auth token for logged in users

dwick:
:eyeglasses: @ajacksified @zeantsoi 

Missing bearer token, this fixes votes and frequency capping for logged in users.


src/config.es6.js
```diff
@@ -24,7 +24,7 @@ var config = {
 
   loginPath: process.env.LOGIN_PATH || '/oauth2/login',
 
-  adsPath: process.env.ADS_PATH ||  'https://www.reddit.com/api/request_promo.json',
+  adsPath: process.env.ADS_PATH ||  '/api/request_promo.json',
```
zeantsoi: Do we know that switching up to a relative path won't have any issues elsewhere in the code, [like here](https://github.com/reddit/reddit-mobile/blob/master/src/views/pages/index.jsx#L187)?

zeantsoi:
One comment, else :fish:

src/config.es6.js
dwick: that's fine

dwick:
:haircut: @ajacksified trying to get this out asap

dwick:
da8d6477635021c2dcc94910689c1197956d17b6

PR #275, opened 2015-07-21
Message compose

curioussavage:
:eyeglasses: @danehansen @dwick

### DIFF
--- a/assets/less/components/messageCompose.less
+++ b/assets/less/components/messageCompose.less
@@ -1,0 +1,118 @@
+@double-top-nav: (@top-nav-height * 2);
+
+.messageCompose {
+  .flex-column-layout;
+  margin: 0 auto;
+  max-width: 1000px;
+  height: calc(100% ~'-' @double-top-nav);
+  min-height: 200px;
+
+  &__select {
+    -webkit-appearance: none;
+    height: 32px;
+    padding: 6px 12px;
+    font-size: 13px;
+    line-height: 1.42857143;
+    color: #505453;
+    background-color: #fff;
+    background-image: none;
+    .themeify({
+      border: @theme-alt-border;
+      background-color: @theme-comment-body-color;
+      color: @theme-body-text-color;
+    });
+
+    &:focus {
+      outline: none;
+    }
+  }
+
+  &__footer-btn {
+    .btn-primary;
+    .pull-right;
+    display: inline-block;
+    margin: 5px;
+    font-weight: normal;
+    text-align: center;
+    vertical-align: middle;
+    touch-action: manipulation;
+    cursor: pointer;
+    background-image: none;
+    border: 1px solid transparent;
+    white-space: nowrap;
+    padding: 6px 12px;
+    font-size: 13px;
+    line-height: 1.42857143;
+    height: @top-nav-height;
+    width: 39%;
+    border-radius: none;
+  }
+
+  &__input {
+    .themeify({
+      border-bottom: @theme-border;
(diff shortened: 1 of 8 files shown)


src/views/components/Modal.jsx
```diff
@@ -10,7 +10,7 @@ class Modal extends BaseComponent {
     }
   }
 
-  componentWillRecieveProps (nextProps) {
+  componentWillReceiveProps (nextProps) {
```
dwick: separate commit for this

src/views/pages/messageCompose.jsx
```diff
@@ -1,24 +1,273 @@
...
+
+  sendMessage (e) {
+    e.preventDefault();
+    // etc etc
+
+    let text = this.state.text;
+    let recipient = this.state.recipient;
+    let subject = this.state.subject;
+    // TODO: let from_sr = this.refs.subject.getDomNode().value;
+    var errors = this._validateFields(recipient, subject, text);
```
dwick: i would think this logic should belong to the `Message` model

curioussavage: I'm inclined to leave this separate for now. The validate function on the model is just meant to return a bool, this method is just there so I can highlight fields that need to be filled. needs to return the names of the fields. bad naming maybe?

curioussavage:
:haircut:

src/views/pages/messageCompose.jsx
dwick: looks to me as if it returns `true` if valid or a list of invalid fields if invalid: https://github.com/reddit/snoode/blob/31595436ac25e4fe758d3a149b6a593e8ae13e7e/src/models/base.es6.js#L66

dwick:
:nail_care:

src/views/pages/messageCompose.jsx
curioussavage: oops. thanks for pointing that out. will do then

curioussavage:
:eyeglasses: @ajacksified 

this one is ready now

ajacksified:
Needs some design changes to bring it in line with submit; :nail_care:

curioussavage:
messagecompose-change
messagecompose-old

curioussavage:
@librarianavenger @tdohz   want some feedback on this form. Never had any design attention so I just built it to look like the submit page. 

The bottom is how it looks like right now, with the exception that my snooboots theme has dark buttons (need to fix that). 

The image above is another possible layout (not styled well right now) with the from select box on the bottom. Can I get some input on this?

I would really like to get this out, the branch has been sitting here done for a while now. Design tweaks are only holdup

dwick:
:running:

curioussavage:
I restyled this view per advice from jack last week. Also refactored the css to fit with new style
messagecompose

curioussavage:
:haircut:

schwers:
note: the select for username / subreddits shouldn't be a select if the user isn't moderating any subredits

assets/less/components/messageCompose.less
```diff
@@ -0,0 +1,107 @@
...
+    &:focus {
+      outline: none;
+    }
+  }
+
+  &__from-wrap {
+    padding-left: 10px;
+    padding-top: 15px;
+    display: inline-block;
+    width: calc(~'100% - 170px');
```
schwers: double check layout is okay-ish when calc isn't supported (eg android 4.3)

curioussavage:
closing this as it has sat here for ages and niranjan is working on a new messages ui

PR #276, opened 2015-07-21
Increase z-index on top nav, fixes search page.

curioussavage:
:eyeglasses: @danehansen 
fix for [MOBILEWEB-226](https://reddit.atlassian.net/browse/MOBILEWEB-226)

The z index being set on the Listings is hiding the nav on search page. Should this be set higher since we will be doing infinite scroll? or is there a better fix? looked at differences between this and index page can't tell what keeps it from happening there.

### DIFF
--- a/assets/less/components/search.less
+++ b/assets/less/components/search.less
@@ -32,11 +32,11 @@
     border-bottom: @gray-border;
     padding-bottom: 10px;
 
-    & + .listings-container {
+    & + .listing-container {
       padding-top: 10px;
     }
 
-    &.hidden + .listings-container {
+    &.hidden + .listing-container {
       padding-top: 0;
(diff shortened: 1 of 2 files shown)


curioussavage:
shoot just realized the sideNav is broken on that page too..

danehansen:
A better solution here is like on the home page where we put position: relative and z-index:1 on the parent container of the Listings. That way, the z-index of no matter how high we go once implementing infinite scroll will be irrelevant. This is probably my bad for not catching this on the search page when I implemented the z-index hack on the home page.

curioussavage:
:nail_care: 

ok the listing-container class was listings-container in the search component.

danehansen:
:fish:

PR #277, opened 2015-07-21
Impression tracking

dwick:
Impressions are only being checked on scroll/resize.  This also checks immediately after the ad is loaded.

The problem before is that I could refresh the page over and over and never have the impression counted so long as I didn't scroll.

:eyeglasses: @curioussavage @zeantsoi


curioussavage:
:fish: lgtm

PR #281, opened 2015-07-23
remove onSubmit from post submit form

curioussavage:
:eyeglasses: @danehansen 

pressing enter while choosing a subreddit would submit the form. this was really no point to having this there anyways


danehansen:
:fish:

PR #285, opened 2015-07-27
add GA events for message reply and pageview

curioussavage:
:eyeglasses: @danehansen 

not sure if GA is the thing to use for pageviews but to my knowledge we don't have anything else right now


danehansen:
:fish:

PR #288, opened 2015-07-28
Don't display scores that are hidden

dwick:
I was primarily concerned with ads always showing 0 due to being heavily downvoted, but figured we might as well hit all the cases .com does.

Here's some screenshots:

screen shot 2015-07-28 at 11 42 21 am
screen shot 2015-07-28 at 11 51 39 am

:eyeglasses: @danehansen @limmmster


dwick:
related: https://github.com/reddit/reddit-public/pull/3965

danehansen:
should we just leave out the voting buttons and score altogether? seems better to me than to have pointless vote buttons and "score hidden" text. @limmmster @dwick

dwick:
@danehansen just because you dont see the score doesn't mean there isn't one and votes still do matter internally.  also, sometimes scores are only hidden for a period of time.

danehansen:
:fish:

dwick:
https://github.com/reddit/reddit-mobile/commit/fc1c91be9fc53d64c3a877a8298f2b1be3639efb

PR #292, opened 2015-08-05
Remove globals

curioussavage:
:eyeglasses: @dwick @JordanMilne 

I wouldn't mind removing all the globals but for now I just removed all of the globals that seemed like they would be related to individual requests.


JordanMilne:
:fish:

dwick:
:fish:

PR #294, opened 2015-08-06
Fix ad response encoding

dwick:
:eyeglasses: @JordanMilne 

on https://m.staging.wireddit.com/

### DIFF
--- a/src/views/components/Ad.jsx
+++ b/src/views/components/Ad.jsx
@@ -67,6 +67,7 @@ class Ad extends BaseComponent {
     var postData = {
       srnames: srnames,
       is_mobile_web: true,
+      raw_json: '1',
     };
 
     // If user is not logged in, send the loid in the promo request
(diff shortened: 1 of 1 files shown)


src/views/components/Ad.jsx
```diff
@@ -77,7 +78,12 @@ class Ad extends BaseComponent {
     }
 
     return new Promise((resolve, reject) => {
-      superagent.post(origin + this.props.adsPath)
+      var adUrl = url.format({
+        host: origin,
+        path: this.props.adsPath,
+        query: { raw_json: true },
```
JordanMilne: nit: raw_json can be either in the postData or the query params. It'd be more idiomatic to put it in the postData here

JordanMilne: Also, `raw_json` must be `=== "1"`

JordanMilne:
:nail_care:

dwick:
:haircut:

JordanMilne:
:fish:

PR #295, opened 2015-08-06
icons: replace react wrapped svgs with icons

curioussavage:
This includes: post, search, hamburger, vote
and seashell icons.

:eyeglasses: @dwick


assets/less/components/TopNav.less
```diff
@@ -1,3 +1,5 @@
+@hamburger-width: 45px;
```
dwick: this is the same as `@top-nav-height` also, variables should be in the variables file if this is actually needed for some reason

dwick:
remove the components too and cleanup the css!

dwick:
:nail_care:

curioussavage:
:haircut:  should be better now

assets/less/components/Dropdown.less
```diff
@@ -9,6 +9,11 @@
     line-height: 20px;
   }
 
+  .icon-seashells {
```
dwick: `.SeashellIcon` above is no longer used. 

Also, i'm not sure how @ajacksified feels about this, but I would generally just add a less presentational class for making small adjustments like you're doing here and remove the nesting.

Example:

```html
<span class="icon-seashells Dropdown-toggle"></span>
```

```css
.Dropdown {
  // ...
}
.Dropdown-toggle {
  display: inline-block;
  font-size: 2em;
}
```

dwick:
I believe the css for `.SeashellIcon`, `.UpvoteIcon`, `.SearchIcon`, `.PostIcon`, `.HamburgerIcon` is still here.   Can you remove that too?

:nail_care:

curioussavage:
:haircut:  that should be the last of the refs to the svg icons that were removed.

assets/less/components/TopNav.less
```diff
@@ -1,3 +1,4 @@
+
```
ajacksified: nuke newline

ajacksified:
I do rather like the animated up/downvote icons, which differ from the rest in that you actually see them animate when clicking; what's the performance impact of leaving them in place?

ajacksified:
:nail_care:

ajacksified:
:fish: after talking in person about icons

PR #298, opened 2015-08-10
Add mountTime to rum timings

ajacksified:
Track the time it takes React to mount and send it to our rum endpoint.

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #299, opened 2015-08-11
Promises

curioussavage:
:eyeglasses: @ajacksified 

Not done yet, but want input before I continue. 

I was looking into promises and read that if you use the reject handler then errors thrown by the resolve handler itself will not be caught.

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -186,7 +186,7 @@ function routes(app) {
         });
 
         resolve(data);
-      }, function(error) {
+      }).catch(function(error) {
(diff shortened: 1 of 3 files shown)


src/routes.jsx
```diff
@@ -155,7 +155,7 @@ function routes(app) {
           try {
```
curioussavage: is the `try` `catch` still necessary with using catch?

ajacksified:
Closing this - the refactoring I'm doing should cover it.

PR #300, opened 2015-08-20
Use props.loid rather than app.state.loid

ajacksified:
State shouldn't be in global app state, or globals at all. This fixes loids; the rest will be fixed soon.

:eyeglasses: @zeantsoi @curioussavage


zeantsoi:
This seems to work for me locally. However, I needed to clear my cookies in order for the value of the `loid` cookie to be passed... think an old cookie value might have been cached.

:fish:

curioussavage:
:fish: for me too

PR #306, opened 2015-09-01
Perf followup

curioussavage:
:eyeglasses: @ajacksified 

Any other small fixes I should sneak into this?


ajacksified:
This should do it! :fish:

PR #307, opened 2015-09-02
SideNav: fix login and logout links

curioussavage:
:eyeglasses: @ajacksified 

Looks like one more thing slipped through.

I am assuming this data-no-route attr is something the koa router looks for?


ajacksified:
:fish: - it's used here: https://github.com/reddit/reddit-mobile/blob/master/assets/js/client.es6.js#L262

ajacksified:
whoops, wrong button.

PR #308, opened 2015-09-02
Don't register routes twice

ajacksified:
:eyeglasses: @curioussavage 

We're already registering in app-mixin, so this is unnecessary (and wasting cpu.)

https://github.com/ajacksified/reddit-mobile/blob/ee68de25914a44dae566c58365a38d522f7a7852/src/app-mixin.es6.js#L21


curioussavage:
:fish:

PR #309, opened 2015-09-03
zoom fix: remove ios zoom fix from project

curioussavage:
:eyeglasses: @ajacksified 

should be the last of the zoom fix stuff

### DIFF
--- a/src/views/layouts/DefaultLayout.jsx
+++ b/src/views/layouts/DefaultLayout.jsx
@@ -72,7 +72,7 @@ class DefaultLayout extends BaseComponent {
           <link href={ baseCSS } rel='stylesheet' />
           { canonical }
 
-          <meta name='viewport' content='width=device-width, initial-scale=1.0' />
+          <meta name='viewport' content='width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no' />
           <meta name='theme-color' content='#336699' />
           <meta name='apple-mobile-web-app-capable' content='yes' />
           <meta id='csrf-token-meta-tag' name='csrf-token' content={ this.props.ctx.csrf } />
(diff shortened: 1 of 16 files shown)


src/views/layouts/DefaultLayout.jsx
```diff
@@ -72,7 +72,7 @@ class DefaultLayout extends BaseComponent {
           <link href={ baseCSS } rel='stylesheet' />
           { canonical }
 
-          <meta name='viewport' content='width=device-width, initial-scale=1.0' />
+          <meta name='viewport' content='width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no' />
```
curioussavage: this is my proposed solution. Every major mobile site I looked at had this.

ajacksified: The downside is that this prevents users from being able to zoom into the high-res images in the listing view. I've seen some solutions add that only when an input/textarea is focused, which allows for both.

curioussavage: If you think its fine I could have the defaultLayout component listen for that and add it

ajacksified: DefaultLayout doesn't get rendered client-side, so that won't do anything - you'll need to manually update the dom.

curioussavage: I tried adding/removing the new meta tag info on `focusin focusout` events but the browser zooms in for a second before it works.

Also tried upping the font size using the css focus pseudo selector but that doesnt seem to work either. 

Not sure what else to try at this point

curioussavage:
@ajacksified any more thoughts about preventing the zoom ? I mentioned to tavish and he thought we should just disable zoom. Although one very simple fix that I am fan of is just increasing the font size to 16 px for inputs and textarea fields.

ajacksified:
:fish:

PR #310, opened 2015-09-09
Rewrite redirects using `param` instead of regexes for koa-router

ajacksified:
:eyeglasses: @curioussavage 

Also notifying @kjoconnor and @umbrae 

The new version of koa-router doesn't support regexes anymore, which means we need to use `param` filters, as seen below. I also moved redirects above `routes` for higher priority and added `return`s so that multiple route-matching functions don't end up being called.


curioussavage:
:fish:

PR #312, opened 2015-09-17
Update the routes to match subreddit urls less aggressively

ajacksified:
Match /r/:subreddit/hot, etc directly instead of /r/:subreddit:sort,
which was catching /r/:subreddit/search and such.

:eyeglasses: @curioussavage


ajacksified:
Addendum: this should eventually redirect the other way around, but this fixes subreddit search in the meantime.

curioussavage:
:fish:

PR #313, opened 2015-09-17
Remove popstate safety check

ajacksified:
Previously, the 'popstate' event listener checked to see if your url had
changed. This is no longer necessary now that events are not bound until
_after_ the app render, and may fix some issues with prev/next browser
cache reloading.

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #314, opened 2015-09-22
Swap "propertyId" for "googleAnalyticsId"

ajacksified:
Accidentally using the wrong property to load GA.


curioussavage:
:fish:

PR #316, opened 2015-09-22
Build options with deepassign instead of nested assigns

ajacksified:
:eyeglasses: @dwick


dwick:
lodash supports this as well :fish:

ajacksified:
Oh, may as well use lodash then. I'll update.

PR #317, opened 2015-09-24
Add "compact" prop to tracking pixel

ajacksified:


umbrae:
Based on https://github.com/ajacksified/reddit-mobile/blob/fix-pixel-compact-tracking/src/lib/TrackingPixel.es6.js#L26 it looks like this already existed and we're just adding it back in?

If so, I'm guessing that means that whatever is consuming the pixel is prepared to take in this data? If so, :fish:

PR #318, opened 2015-09-24
Client oauth refresh improvements

ajacksified:
* Don't route the app if we're in the middle of refreshing the token,
cutting down on unnecessary auth errors
* Show a loading screen while oauth is being refreshed, if you try to
change the route in the middle, so things don't freeze
* Fixes a current bug where if you're away for a while, then come back,
you're taken to a login screen

To test:  set `expires = Date.now()` right after `var expires = new Date(app.getState('tokenExpires'));` in `client.es6.js`, watch for the token to change on the next api request. woo :tada:

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -1,5 +1,6 @@
 import 'babel/polyfill';
 
+import React from 'react';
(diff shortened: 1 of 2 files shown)


curioussavage:
:fish:

PR #319, opened 2015-09-25
TopNav: refactor component

curioussavage:
Remove event for topnav change, relies on
new props instead. Fixes subscription toggle bug
and multi name also.

:eyeglasses: @ajacksified 

fixes following bugs in prod:

* no subreddit title in comments page
* multi name is not displayed in the topnav
* subscribing is broken

### DIFF
--- a/src/constants.es6.js
+++ b/src/constants.es6.js
@@ -1,7 +1,6 @@
 export default {
   SIDE_NAV_TOGGLE: 'sideNavToggle',
   TOP_NAV_HAMBURGER_CLICK: 'topNavHamburgerClick',
-  TOP_NAV_CHANGE: 'topNavSubredditChange',
   VOTE: 'vote',
   DROPDOWN_OPEN: 'dropdownOpen',
   COMPACT_TOGGLE: 'compactToggle',
--- a/src/views/components/TopNav.jsx
+++ b/src/views/components/TopNav.jsx
@@ -14,68 +14,75 @@ class TopNav extends BaseComponent {
     super(props);
 
     this.state = {
-      title: props.topNavTitle,
-      link: props.topNavLink,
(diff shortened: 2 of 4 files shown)


src/constants.es6.js
```diff
@@ -1,7 +1,6 @@
 export default {
   SIDE_NAV_TOGGLE: 'sideNavToggle',
   TOP_NAV_HAMBURGER_CLICK: 'topNavHamburgerClick',
-  TOP_NAV_CHANGE: 'topNavSubredditChange',
```
curioussavage: Seemed more straightforward to just rely on props since only time the title changes is during a route where the bodyLayout component is updated anyways.

src/views/components/TopNav.jsx
```diff
@@ -163,7 +166,7 @@ class TopNav extends BaseComponent {
       });
 
       this.setState({
-        subreddit: Object.assign(subreddit, {
+        subreddit: Object.assign({}, state.subreddit, {
```
curioussavage: This was screwing up `componentDidMount` on the `baseComponent`. Since a reference to state was being modified the nextState and this.state were already equal.  The React docs warn about this exact bug [here](https://facebook.github.io/react/docs/advanced-performance.html). It's possible we might have other bugs related to this and even if not it would be easy to cause again.

Docs also point out how expensive doing deep comparisons are for perf. With how often this method is running in our app this might be a net loss for perf. Based on that It would probably be a good idea to consider removing `baseComponent`.

ajacksified: :fish:

curioussavage:
noticed a small problem and decided to refactor some more

curioussavage:
:nail_care:

curioussavage:
:haircut:

ajacksified:
:fish:

PR #320, opened 2015-09-28
Use apiOptions from props for user info

ajacksified:
Fixes, among other things, API_PASS_THROUGH_HEADERS config. Probably means we should find a better method for calling out to the API as part of `app` or something eventually.

:eyeglasses: @curioussavage or @umbrae.


umbrae:
:fish: 

Yeah, by the look of it it looks really easy to accidentally bypass `buildAPIOptions` (and I'm not super clear on the difference between that and `app.api.buildOptions` and when you'd want one vs the other, or if they should be merged or what).

Should we put in a ticket to refactor this?

PR #323, opened 2015-09-29
Use subreddit name as

ajacksified:
:eyeglasses: @curioussavage 

https://github.com/reddit/snoode/pull/47

### DIFF
--- a/src/serverRoutes.es6.js
+++ b/src/serverRoutes.es6.js
@@ -103,6 +103,19 @@ Allow: /
       this.redirect(this.headers.referer || '/');
     });
   });
+
+  app.router.get('/routes', function* () {
+    this.body = app.router.stack.routes
+      .filter(function(r) {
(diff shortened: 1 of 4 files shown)


curioussavage:
Ill need to wait for this one to rebase, I touched some of the same files

src/serverRoutes.es6.js
```diff
@@ -103,6 +103,19 @@ Allow: /
       this.redirect(this.headers.referer || '/');
     });
   });
+
+  app.router.get('/routes', function* () {
```
curioussavage: when hitting this in the browser I get a routeError for /favicon.ico and a TypeError because app is undefined in horse-react. server.es6.js 45:25  not sure if that is really an issue because this is intended for another server?

ajacksified: Not an issue (and the 404 PR fixed this anyway)

curioussavage:
:fish:

PR #325, opened 2015-10-01
Edit fix

curioussavage:
:eyeglasses: @ajacksified 

fix for editing in prod. I did some refactoring as well. decided I should just get it out before going further

### DIFF
--- a/src/app-mixin.jsx
+++ b/src/app-mixin.jsx
@@ -10,6 +10,7 @@ import { v1 as V1Api } from 'snoode';
 
 import BodyLayout from './views/layouts/BodyLayout';
 import ErrorPage from './views/pages/error';
+import Loading from './views/components/Loading';
 
 const errorMsgMap = {
   '404': 'Sorry, that page doesn\'t seem to exist.',
--- a/src/views/components/Comment.jsx
+++ b/src/views/components/Comment.jsx
@@ -22,19 +22,24 @@ class Comment extends BaseComponent {
     super(props);
 
     this.state = {
-      comment: this.props.comment,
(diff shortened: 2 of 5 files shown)


src/app-mixin.jsx
```diff
@@ -10,6 +10,7 @@ import { v1 as V1Api } from 'snoode';
 
 import BodyLayout from './views/layouts/BodyLayout';
 import ErrorPage from './views/pages/error';
+import Loading from './views/components/Loading';
```
curioussavage: my token was not refreshing and it was calling the loading method in app-mixin Once I added this component it put the loading screen up but just sat there.

src/views/components/Comment.jsx
```diff
@@ -134,136 +117,116 @@ class Comment extends BaseComponent {
...
 
-  preventPropagation (e) {
-    e.stopPropagation();
   }
 
   render () {
-    if (this.state.hidden) {
-      return (<div />);
-    }
+    let {app, apiOptions, token, nestingLevel, permalinkBase, user, op, highlight, subredditName} = this.props;
```
curioussavage: I kinda went crazy with the destructuring, personally I don't mind the long line, but I can break it up if you think it would be nicer

ajacksified: Yeah, break it up; destructuring is a good idea.

ajacksified:
One note: comment dropdowns say 'edit post' / 'delete post' instead of 'edit comment' / 'delete comment'

ajacksified:
:nail_care:

curioussavage:
:haircut:

ajacksified:
:fish:

PR #326, opened 2015-10-01
IndexPage: change key to use title

curioussavage:
this fixes route changes between multis and home

:eyeglasses: @ajacksified


ajacksified:
:fish:

PR #330, opened 2015-10-05
Comments: add load more comments using permalinks

curioussavage:
:eyeglasses: @ajacksified 

works fine, only concern of mine is that it can be a little disorienting when comment tree is large, its easy to lose your place


ajacksified:
The href should point you to the `context` link, but the javascript case should ajax in the children and just append to the current tree. Sorry if I wasn't clear on that! :nail_care:

curioussavage:
oops, misunderstood that

PR #335, opened 2015-10-09
Comment: add missing await keyword

curioussavage:
Also use async function for delete for consistency

:eyeglasses: @ajacksified 

looks like I missed an `await`  this is breaking editing comments right now.


ajacksified:
:fish:

PR #336, opened 2015-10-13
Testing: set up mocha and chai, add some example tests

curioussavage:
:eyeglasses: @ajacksified 

Ok I added mocha and chai/sinon along with jsdom. Not really tied to that if you want to do something else, but it is working fine.

I could probably improve the tests but I just wanted to get it set up with some examples in the repo so anyone can start adding tests.


ajacksified:
Looks like a great start! :fish:

PR #337, opened 2015-10-14
Remove commentsmap now that "show more" works

ajacksified:
:eyeglasses: @curioussavage 

Also @tdohz and @spez for notification.

Before: we'd set comments to collapse under a certain score, to show less elements for mobile users. Now, we're going to use the desktop algorithm now that "load more comments" works. Eventually, it would be nice to have the algorithm accept a "score" that we can set to lower for mobile apps. As a side effect, comments load faster now, hooray.

Before: left, After: right
screen shot 2015-10-14 at 4 48 39 pm


spez:
:fish:

tdohz:
Looks good. Can I see this on staging (also a nice excuse to check out staging)?

ajacksified:
@tdohz it's up on staging now

PR #338, opened 2015-10-15
Comment minor improvements

ajacksified:
On the ferry this morning, I made some small tweaks:

* Change `[score hidden]` to `[?]` to save space
* Change `Load more comments (57 replies)` to `Load 47 replies`
* Reduce horizontal padding for nested comments by a couple pixels per nesting level
* Increase vertical padding slightly for individual comments
* Reduce vertical padding after comment box
* Increased hit area on text (which shows upvote/downvote/reply/etc)
* Fixed bug where the color won't change on vote icons until you click something else, after voting

New (left) vs Old (right)

screen shot 2015-10-15 at 10 24 31 am
screen shot 2015-10-15 at 10 19 50 am

:eyeglasses: @curioussavage, @tdohz

### DIFF
--- a/assets/less/components/comment.less
+++ b/assets/less/components/comment.less
@@ -50,8 +50,7 @@
 }
 
 .comment-article {
-  .transition(.2s all ease);
-  padding-left: .5em;
+  padding-left: 3px;
 
   &.comment-highlighted {
     background-color: darken(@brand-primary-bg-lightest, 2.5%);
@@ -66,15 +65,23 @@
   margin-left: .4em;
 }
 
+.comment-loadmore {
+  margin-bottom: 1em;
+
+  a:active, a:hover {
+    text-decoration: none;
+  }
+}
+
 .linkbar.comment-title-list {
   width: 100%;
 
   &.linkbar-compact li.twirly {
-    left: -14px;
-    width: 19px;
+    left: -10px;
+    width: 12px;
     padding-bottom: 0;
     margin-right: 0;
-    height: 17px;
+    height: 18px;
 
     &:before {
       display: inline-block;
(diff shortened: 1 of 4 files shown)


umbrae:
I think I don't love [?] because it feels like a help button? Not sure.

librarianavenger:
The comment icon + count really helps. 
Gotta agree that the [?] doesn't help. When i first saw it, i thought it was the "before" image and was glad it was gone. Do we need to put anything there at all when the score isn't available?

tdohz:
Pulling in @librarianavenger for comment. I agree that [?] is a little ambiguous. Maybe just [hidden]? 

The other changes look OK, although the repeated comment bubble icon is a little distracting.

If we're making minor tweaks anyway, I would love to see some sort of separator between the timestamp and score, maybe a bullet.

ajacksified:
@tdohz / @librarianavenger:

screen shot 2015-10-15 at 12 12 08 pm

tdohz:
I like baby bullet as separator, but big bullet next to baby bullet looks weird (I know it's what we use on desktop but in a vertical context).

Maybe [hidden]? @librarianavenger thoughts?

ajacksified:
Used the bullet because `[hidden]` takes up a lot of space once you're several levels deep, and that isn't a problem that will be solved before we do an actual design pass at this; figured that consistency with desktop was better than `[score hidden]`, even if it isn't perfect to use in the long term. Happy to change to whatever @librarianavenger prefers.

librarianavenger:
Baby bullet is great. Why do we need something if the data isn't available? I'd leave it empty.

tdohz:
I'm OK with empty for now.

ajacksified:
Alrighty, I'll go with empty.

PR #339, opened 2015-10-15
Request ALL the scopes

ajacksified:
In particular, lacking scopes for subscribe, privatemessages, report, and edit were missing and breaking functionality for some users.

We may want to come up with a strategy for nuking login cookies after this goes live to help everyone along.

:eyeglasses: @curioussavage @umbrae


curioussavage:
:fish:

PR #342, opened 2015-10-16
Scopes cookie

ajacksified:

This makes it so that if we change oauth scopes in the future, we can invalidate existing login cookies. (Like https://github.com/reddit/reddit-mobile/commit/73c9b73d2b8b36f060e9b844cbf33472e6474374)

### DIFF
--- a/src/server.jsx
+++ b/src/server.jsx
@@ -341,9 +341,14 @@ class Server {
 
   checkToken (app) {
     return function * (next) {
-
       var now = new Date();
       var expires = this.cookies.get('tokenExpires');
+
+      if (this.cookies.get('tokenScopes') !== app.oauthScopes) {
+        app.nukeTokens(this);
+        yield next;
+        return;
+      }
 
       if (!expires) {
         yield next;
@@ -372,10 +377,9 @@ class Server {
             console.log(e, e.stack);
           }
 
-          this.cookies.set('tokenExpires');
-          this.cookies.set('token');
-          this.cookies.set('refreshToken');
+          app.nukeTokens(this);
           this.redirect('/');
+
           return;
         }
       }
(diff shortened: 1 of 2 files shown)


curioussavage:
:fish:

src/server.jsx
```diff
@@ -341,16 +341,16 @@ class Server {
...
         yield next;
         return;
       }
 
-      if (skipAuth(app, this.url)) {
+      var now = new Date();
+      var expires = this.cookies.get('tokenExpires');
+
+      if (this.cookies.get('tokenScopes') !== app.oauthScopes) {
+        app.nukeTokens(this);
```
umbrae: Q: Does this mean that if we ever alter scopes for the app that every user will be logged out and forced to reauthenticate, or does it mean that we'll nuke the cookies and then regenerate a new token silently?

umbrae:
:nail_care: for a Q

src/server.jsx
ajacksified: Yes, they will be forced to reauthenticate - oauth flows don't allow for users to silently 'accept' new scopes, which would defeat a primary security model for oauth.

umbrae:
Jack and I talked about this, and we came up with a middle ground:

1. Use this for now.
2. Create a ticket to make a special scope, "all", that we will use for mobile web, which will allow for all functionality for first party clients.
3. Once we have that built, we can update this to only use the "all" scope and not worry about forward compatibility.

:fish: on that

kjoconnor:
We have that scope, it's `*`

umbrae:
Yay! I thought we might have it, but I hadn't known for sure. @ajacksified can you test that?

Thank you @kjoconnor !

kjoconnor:
I tried poking around in the code originally to see how we grant access to requesting that scope, but couldn't find it easily. I'm hoping it already uses one of the ways we whitelist clients, but if not I can check into it a little more deeply.

PR #343, opened 2015-10-16
Comment box should not show up if you're not logged in.

madbook:
for [**MOBILEWEB-285**](https://reddit.atlassian.net/browse/MOBILEWEB-285)

Some general questions, since I'm new to reddit-mobile world:

e516708 - Any strong opinions on doing conditional components inline like this vs setting to a variable above? This seems easier to read IMO, but it seems like you've been generally doing the latter.

b70b832 - Better way to do this?

Is checking `token` like this the correct way to determine logged-in vs logged out? Poking around this seemed to be what is being used elsewhere.


tdohz:
Can you push to mweb staging when you get a chance?

ajacksified:
Previously, if you made a comment and you weren't logged in, it would store the comment, send you to the login page, and redirect you. As part of this PR, the code that restores comments from localstorage after login should also be removed (@curioussavage can give you more details, but it appears in areas like https://github.com/reddit/reddit-mobile/search?utf8=%E2%9C%93&q=localstorage).

Also, the `reply` button on comments should redirect you to login, like the voting buttons do.

ajacksified:
:nail_care:

madbook:
@ajacksified 9070171 cleans up the saved reply stuff you mentioned, the reply button redirect was already in there (though implemented incorrectly, fixed up in bdb0145).

5b63686 includes a bunch of cleanup and refactoring to use the latest 'n greatest babel stuff.  It includes a bunch of formatting tweaks which are a matter of my own preference/opinion and up for debate.  We should probably add an extension to our styleguide for es6+ & jsx stuff since we're going to be using it more and more.  I can also leave this commit off completely if it's too... controversial?

4188082 fixes some weirdness introduced by removing the savedReply stuff - specifically the textarea losing anything you've typed if you toggle the reply button.  This still happens if you toggle the menu completely or collapse & expand the comment.  I can fix those cases up as well, but I wanted to avoid creeping to much out of the `CommentBox` component.

:haircut:

src/views/components/CommentBox.jsx
```diff
@@ -1,107 +1,134 @@
...
+    apiOptions: PropTypes.object.isRequired,
+    app: PropTypes.object.isRequired,
+    ctx: PropTypes.object.isRequired,
+    onSubmit: PropTypes.func.isRequired,
+    thingId: PropTypes.string.isRequired,
+    token: PropTypes.string,
+    visible: PropTypes.boolean,
+  };
+
+  static get defaultProps() {
```
ajacksified: :no_bell: why does this need to be a getter instead of a static property?

ajacksified:
Should we show a "please log in to leave a comment" link or something in place of the comment box, @tdohz / @librarianavenger?

src/views/components/CommentBox.jsx
madbook: ah, boo to me.  even though it's _directly adjacent_ to the other change I made, I reverted to an older syntax.  before there was class property support, you could still mimic the behavior by using a static getter.

librarianavenger:
Absolutely - In an ideal world, the user would be able to log in / sign up as part of the save/post step. There's a really cool example here -  
disqus-login

It might be better to disable the submit button and put "please log in" above the disabled button.

librarianavenger:
Many places, like Medium, just pop open the login/signup page when the user clicks into the comment box.

src/lib/savedReply.es6.js
```diff
@@ -1,52 +0,0 @@
-'use strict';
```
schwers: Does this intentionally remove saved replies? Nice to have if you tap a link mid commenting.

schwers:
@madbook and @nramadas Can you coordinate if this is still needed with the [new comments pr](https://github.com/reddit/reddit-mobile/pull/457). 

For login/redirect stuff [this patch](https://github.com/reddit/reddit-mobile/pull/462) should clean a lot of that up

ajacksified:
Closing due to redesign.

PR #344, opened 2015-10-16
Editing comments: reuse original comments replies

curioussavage:
After editing we get the comment without replies,
which would cause the children to vanish at next
render

https://reddit.atlassian.net/browse/MOBILEWEB-290
but in the wild

:eyeglasses: @madbook @ajacksified 

I think this is better than just merging the new body html/text into the comment because we get updates to votes.  

Even better might be to render the route again on changes like this so we get that unidirectional data fluxy flow goodness.

### DIFF
--- a/src/views/components/Comment.jsx
+++ b/src/views/components/Comment.jsx
@@ -154,6 +154,7 @@ class Comment extends BaseComponent {
     try {
       let data = await app.api.comments.patch(options);
       if (data) {
+        data.replies = oldComment.replies;
         this.setState({
           comment: data,
           editing: false,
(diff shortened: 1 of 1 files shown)


ajacksified:
Rendering the route again is overkill for an edit - I think, in this case, it's best to just edit the comment, and let the user do something to explicitly reload the page, or else it would be distracting if everything updated. 

:fish:

curioussavage:
I agree then, rendering again could alter the collapsed state of comments among other things which would be distracting.

madbook:
:fish:

ajacksified:
https://github.com/reddit/reddit-mobile/commit/adba3592c8c2ae2ff5e01572c3077178327de2ee

PR #347, opened 2015-10-19
Add new `icon-inline` class to fix icon color in links

ajacksified:
:eyeglasses: @curioussavage 

To be used when an icon is used inline with text, rather than as a nav / heading / etc.


curioussavage:
:fish:

PR #348, opened 2015-10-19
Global error logging; first pass, sends to statsd.

ajacksified:
Together with https://github.com/reddit/hivemind/pull/21 and https://github.com/reddit/horse-react/pull/6 (both merged now)

This first pass sets us up to log to statsd (hivemind) and gets us ready for more complicated logging (winston, probably.) This also changes the `debug` config to `debugLevel`.

Note: error functions had to be defined three times; as early as possible for uncaught exceptions, and again with more detail within the app, which should give us decent traces.

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -1,5 +1,22 @@
 import 'babel/polyfill';
 
+import errorLog from '../../src/lib/errorLog';
+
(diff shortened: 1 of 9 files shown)


curioussavage:
:fish: after rebase and making sure there are no checks for `app.config.debug left`

PR #349, opened 2015-10-19
Ignore CSRF on /timings

ajacksified:
:eyeglasses: @curioussavage, and also @JordanMilne for visibility 

Don't use an invalid, cloudflare'd CSRF token on timings.


ajacksified:
(Updated after speaking to @JordanMilne in slack)

JordanMilne:
:fish:

PR #350, opened 2015-10-19
Revert token-nuke and get all scopes by *

ajacksified:
:eyeglasses: @umbrae 

Thanks for the pointer, @kjoconnor


umbrae:
:fish: yay

PR #351, opened 2015-10-19
Version gate node version

ajacksified:
:eyeglasses: @umbrae

### DIFF
--- a/version.js
+++ b/version.js
@@ -1,0 +1,6 @@
+var semver = require('semver');
+var packageJSON = require('./package.json');
+
+if (!semver.satisfies(process.version, packageJSON.engines.node)) {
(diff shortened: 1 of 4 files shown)


version.js
```diff
@@ -0,0 +1,6 @@
+var semver = require('semver');
+var packageJSON = require('./package.json');
+
+if (!semver.satisfies(process.version, packageJSON.engines.node)) {
```
umbrae: This will make it bomb out on iojs, right? If so, is there a reason we need `iojs` in `package.json`'s `engines`?

umbrae:
:nail_care: just a small q!

ajacksified:
Oh, derp

ajacksified:
:haircut:

umbrae:
:fish:

PR #353, opened 2015-10-20
Fix hidden

curioussavage:
:eyeglasses: @madbook @ajacksified 

other than all the es5 to es6 cleanup noteworthy items are:

userSaved: I refactored it a little while testing and had to refactor listing to get it working again.


curioussavage:
I think we pretty much picked let, I do see that I forgot to use it a few times though, will update.

ajacksified:
:fish:

PR #354, opened 2015-10-21
SideNav: don't rely on datacache for user or prefs

curioussavage:
a few small changes to SideNav for readability too

:eyeglasses: @ajacksified 

sidenav: Doesn't rely on the datacache for subscriptions (the pr moving cache to snoode broke subscriptions in the menu), user and subscriptions are passed as props

Also pulled the data for the about reddit links into another file, saved about 70 lines :)


curioussavage:
:eyeglasses: @madbook if you have a sec

ajacksified:
Whoops, :fish:

PR #356, opened 2015-10-22
ListingList: get rid of extra resize

curioussavage:
Each Listing was already calling its resize method.
Also fixes ad thumbnail aspect ratios

:eyeglasses: @ajacksified @madbook 

both the listinglist and listing components were listening to the resize event and firing the resize methods on the listings.

The change to listingContent also fixes ad image aspect ratio, that really needs some more refactoring but I wanted to get that out since there are paid ads running and it looks crappy.

### DIFF
--- a/src/views/components/ListingContent.jsx
+++ b/src/views/components/ListingContent.jsx
@@ -3,11 +3,13 @@ import MyMath from '../../lib/danehansen/utils/MyMath';
 import mobilify from '../../lib/mobilify';
 import propTypes from '../../propTypes';
 
+const PropTypes = React.PropTypes;
+
 import BaseComponent from './BaseComponent';
 
-var _gifMatch = /\.(?:gif)/gi;
-var _gfyRegex = /https?:\/\/(?:.+)\.gfycat.com\/(.+)\.gif/;
-var _httpsRegex = /^https:\/\//;
+const _gifMatch = /\.(?:gif)/gi;
+const _gfyRegex = /https?:\/\/(?:.+)\.gfycat.com\/(.+)\.gif/;
+const _httpsRegex = /^https:\/\//;
 const _DEFAULT_ASPECT_RATIO = 16 / 9;
 
 function _gifToHTML5(url) {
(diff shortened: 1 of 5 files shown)


ajacksified:
:fish:

src/views/components/ListingContent.jsx
```diff
@@ -465,26 +465,21 @@ class ListingContent extends BaseComponent {
...
-      return props.single ? ratio : _limitAspectRatio(ratio);
-    } catch (err) {
+    let { listing, single } = this.props;
+
+    if (listing.media) {
+      let oembed = listing.media.oembed;
+      let ratio = oembed.width / oembed.height;
+      return single ? ratio : _limitAspectRatio(ratio);
+    } else if (listing.preview) {
+      let source = listing.preview.images[0].source ||
```
madbook: is it possible for `listing.preview.images` to be empty? if so this will throw an error.

src/views/components/ListingList.jsx
```diff
@@ -203,15 +185,17 @@ class ListingList extends BaseComponent {
   }
 }
 
+const { bool, number, arrayOf, oneOfType } = React.PropTypes;
```
madbook: let's decide on how we want to format propTypes.  Personally, I'm not really a fan of aliasing the really generic propType objects off of the `PropTypes` object.  I'm also a fan of moving the whole definition into the `class` block, e.g.

```javascript
class ListingList extends BaseComponent {
  ...

  static propTypes = {
    // them propsTypes
  }
}
```

Both of these are style preferences, so not terribly important, but we should agree on something since more people are going to be working on it. @ajacksified thoughts/preferences?

madbook:
:nail_care: two questions

src/views/components/ListingContent.jsx
curioussavage: the only case I know of is for ads. but in the constructor for the Listing model we add it there.

src/views/components/ListingList.jsx
ajacksified: `static` is preferable.

curioussavage:
:haircut:

ajacksified:
:fish:

madbook:
can you update https://reddit.atlassian.net/browse/MOBILEWEB-319 and link this PR to it?

nit: In the future, we should try to avoid tacking extra things on to PRs like this, it makes it harder to hunt down where changes cames from, and its easier to track things between github and jira when there's a 1-1 mapping of PRs to jira tickets.

PR #357, opened 2015-10-22
SideNav: call close method on route start event

curioussavage:
:eyeglasses: @madbook 

this one slipped through in the sideNav refactor. I had searched the project for the `route:start` event and didn't find it so I assumed (incorrectly) that this did nothing. Right now when you click on a subreddit in sidenav it doesnt collapse the menu


ajacksified:
:fish:  - @curioussavage, it's called from horse: https://github.com/reddit/horse/blob/master/src/app.es6.js#L35

madbook:
:fish:

PR #358, opened 2015-10-27
Server

curioussavage:
:eyeglasses: @ajacksified @madbook 

No significant code changes here. Just made a few quick changes to how things are organized. separates the server only stuff into its own directory


src/server/config.es6.js
```diff
@@ -0,0 +1,61 @@
+
```
ajacksified: extra newline :scream:

ajacksified:
The moving around looks good. If you feel like it, take a pass and swap `var`s for `const` and `let`s. Small style fix, and :fish: as current

ajacksified:
:fish:

PR #359, opened 2015-10-27
Sub about

curioussavage:
:eyeglasses: @ajacksified or @madbook 

the index, comment and about pages all had some shared stuff, I put that into `getSubreddit` to be a little more dry.

this fixes:
* broken about page
* fake subreddits like r/mod
* subreddit menu wasn't showing up
* changes the key for the current subreddit to singular 'subreddit' to avoid confusion.
* in the indexPage route handler I removed the `ctx` variable, in this and other routes it was being used with `this`  both referring to the same context. This has been very confusing to me in the past. If it sounds good, for clarity, I would like to remove use of a `ctx` variable everywhere in all the route handlers so it only appears in functions that are passed a context as an argument.

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -350,14 +350,14 @@ function routes(app) {
     });
 
     let subredditOpts = buildAPIOptions(ctx, {
-      id: props.subredditName.toLowerCase(),
+      id: ctx.props.subredditName.toLowerCase(),
     });
 
-    this.props.data.set('subreddit', app.api.subreddits.get(subredditOpts));
+    ctx.props.data.set('subreddit', app.api.subreddits.get(subredditOpts));
 
-    var key = this.key;
+    var key = ctx.key;
 
-    this.body = function(props) {
+    ctx.body = function(props) {
       return (
         <BodyLayout {...props}>
           <SubredditAboutPage {...props} key={ key } track='subreddit' />
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -124,6 +124,29 @@ function buildProps(ctx, app) {
   return ctx.props;
 }
 
+function getSubreddit(ctx, app) {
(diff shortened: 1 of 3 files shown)


src/routes.jsx
```diff
@@ -124,6 +124,31 @@ function buildProps(ctx, app) {
...
+    Object.assign(ctx.props, {
+      subredditName: subreddit,
+      topNavLink: `/r/${subreddit}`,
+      title: `r/${subreddit}`,
+      metaDescription: `r/${subreddit} at reddit.com`,
+    });
+
+    let fakeSubs = /all|mod|friends/i;
+    if (subreddit.indexOf('+') === -1 &&
+     subreddit.match(fakeSubs) === null
```
ajacksified: Instead of a regex, use `const fakeSubs = ['all','mod','friends']; fakeSubs.includes(subredit)`. Also, the regex you're using will match _any_ subreddit that has 'all', 'mod', or 'friends' anywhere within it.

src/routes.jsx
```diff
@@ -204,43 +229,36 @@ function routes(app) {
       buildProps(this, app);
       yield next;
    });
+  router.use(function * (next) {
```
ajacksified: let's put newlines before and after this, since it's multiline.

ajacksified:
Multireddits are broken in this branch, but not in prod - not sure if something here caused that.

ajacksified:
:nail_care:

src/routes.jsx
curioussavage: oops. ok thanks for noticing this

ajacksified: Also: `random`, `randnsfw`, and `myrandom` (we don't implement them yet, but they should be added to the list for the future.)

curioussavage:
:haircut:

ajacksified:
:fish:

PR #360, opened 2015-10-28
Event tracking

ajacksified:
:eyeglasses: @umbrae @curioussavage 

Refactors event tracking a bit, and adds in the event-tracker to send data to the events collector.

PR #361, opened 2015-10-28
Implement events

ajacksified:
Depends on https://github.com/reddit/reddit-mobile/pull/360 (which is has been rebased on top of.) Relevant commits are everything beginning with `[events]`.


ajacksified:
:eyeglasses: @umbrae @curioussavage

umbrae:
I think we'll need to get these events vetted by data before it goes through to prod and poke around with them on the test endpoint. Maybe @drunken-economist or @kaitaan can help us do that?

PR #363, opened 2015-10-29
Add request url to errorlog, and fix client-side error pages

ajacksified:
* add requestUrl to the error log, so we can see _where_ errors are happening on the server
* also emit a body replacement event on the client; because everything's async, errors were just showing loading spinners forever
* add props to the context on client-side error handling, so we can render error pages properly

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #366, opened 2015-10-30
ListingContent: use second resolution provided for ads case

curioussavage:
the first image resolution is for a thumbnail with
a different aspect ratio

:eyeglasses: @ajacksified 

turns out this was all that is necessary. tested with compact/ listing view. thumbnail is unaffected

### DIFF
--- a/src/views/components/ListingContent.jsx
+++ b/src/views/components/ListingContent.jsx
@@ -474,8 +474,11 @@ class ListingContent extends BaseComponent {
       let ratio = oembed.width / oembed.height;
       return single ? ratio : _limitAspectRatio(ratio);
     } else if (listing.preview && (listing.preview.images || []).length) {
-      let source = listing.preview.images[0].source || 
-        listing.preview.images[0].resolutions[0];
+      let images = listing.preview.images;
+      // we use the second item in resolutions if it exists because for ads
+      // the first item is a thumbnail with 1x1 aspect ratio.
+      let source = images[0].source || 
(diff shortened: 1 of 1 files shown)


umbrae:
Could we have a comment here indicating why it's necessary? I think most folks coming across this later would have no idea why we'd choose one over the other.

curioussavage:
will do

dwick:
can we not pass the image used in rendering to this? seems hacky

curioussavage:
I think I get what you mean. off the top of my head I have no idea. this file is in need of a major refactor so its difficult to tell what does what. I would rather put that on a list of changes for another pass, especially since this is breaking prod

dwick:
let me take a look

curioussavage:
up on staging here: https://www.cloudflarestagingformobilereddit.com/comments/3qhxub/experience_authentic_mlb_in_perfect_inning_15/

dwick:
i think if you changed `_previewImageUrl` to return the preview image object you could do it pretty easily

curioussavage:
yeah, sort of, its getting called in a few places

curioussavage:
:eyeglasses: @madbook

madbook:
this aspect ratio stuff is super weird and should get yanked, but if this is urgent then :+1: to doing the minimal to fix it for ads

:fish:

curioussavage:
After experimenting with images that don't have the aspect ratio stuff. will likely remove the aspect ratio stuff completely unless @ajacksified remembers a reason why its there. images look fine without it as far as madbook and I could see

PR #368, opened 2015-10-31
Align icon-post in Firefox

Gheoan:
### Firefox
Before: 
screen shot 2015-10-31 at 17 55 32
After:
 screen shot 2015-10-31 at 17 56 10
### Chrome
Icon is aligned correctly with and without this PR.


Gheoan:
Fixed by fbbe0b1

PR #369, opened 2015-10-31
Fix missing domain bug on listing footer

ajacksified:
:eyeglasses: @spladug


spladug:
:fish: i'll take your word for it

PR #371, opened 2015-11-02
Fix no-data error page rendering

ajacksified:
:eyeglasses: @curioussavage 

Specifically for: https://github.com/reddit/snoode/pull/56

This helps renders on error pages for API timeouts, which won't have data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "reddit-mobile",
-  "version": "3.0.3",
+  "version": "3.0.4",
   "description": "",
   "main": "src/app.es6.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "semver": "^5.0.3",
     "simple-oauth2": "0.2.1",
     "snoode": "1.11.4",
-    "superagent": "1.3.0",
+    "superagent": "1.4.0",
```
curioussavage: so why don't we bump snoode to 1.4?

curioussavage:
:nail_care:

package.json
ajacksified: Whoops, reverting.

curioussavage:
:fish:

PR #372, opened 2015-11-03
Session conversion does not like * scope

ajacksified:
The post to password flow works (`/api/fp/1/auth/access_token`) but the post to session conversion / modhash list (`api/v1/authorize`) isn't happy.

:eyeglasses: @umbrae

### DIFF
--- a/src/server/oauth.es6.js
+++ b/src/server/oauth.es6.js
@@ -5,7 +5,7 @@ import uuid from 'uuid';
 import url from 'url';
 import querystring from 'querystring';
 
-const SCOPES = '*';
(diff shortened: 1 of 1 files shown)


umbrae:
Can we add TEMP to this commit msg to indicate it's temporary while we look into fixing `authorize`? Additionally a bit more context on the issue and why we're doing this may be helpful for reddit historians later (I.E. Ourselves in 2 months when we're digging through commits for some reason.)

umbrae:
:fish: on the code itself barring commit msg improvements though

ajacksified:
@umbrae updated 4 u

umbrae:
A+ would read again

PR #373, opened 2015-11-03
Throw a 400 for missing user agent

ajacksified:
:eyeglasses: @curioussavage


curioussavage:
can you explain why? :nail_care:

spladug:
There are a non-trivial number of users who use extensions/settings to deliberately remove their user agent for privacy reasons.

ajacksified:
@spladug well then, I'll make everything work with an empty string instead.

PR #375, opened 2015-11-04
Refactor loading sub/userdata

ajacksified:
Health and other routes were loading subreddit and user data, which is
unnecessary, and just slows everything down.

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #376, opened 2015-11-04
Server console

ajacksified:
:eyeglasses: @curioussavage, @umbrae 

This adds:
* snoode api request logging (via https://git.io/vlxC1)
* active open requests logging stub (so we can track this in statsd later)
* a developer console, enableable with the `--console` flag (`node index.js --console`)

If the console is up, it will write logs to `reddit-mobile.log` in your project directory, clearing it out each time you start up the console.

http://gfycat.com/NegligibleAdorableBat

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,8 @@
   },
   "devDependencies": {
     "babelify": "6.1.2",
+    "blessed": "^0.1.81",
+    "blessed-contrib": "^2.5.1",
     "browserify": "10.2.4",
     "buffer": "3.4.3",
     "chai": "^3.3.0",
@@ -50,7 +52,8 @@
     "envify": "3.4.0",
     "exorcist": "0.4.0",
     "fastclick": "1.0.6",
-    "glob": "5.0.10",
+    "filesize": "^3.1.3",
+    "glob": "^5.0.15",
     "gulp": "3.9.0",
     "gulp-buffer": "0.0.2",
     "gulp-concat": "2.6.0",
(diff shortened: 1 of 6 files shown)


src/server/console.es6.js
```diff
@@ -0,0 +1,411 @@
...
+    });
+
+    this.emitter.on('logRequest', (data) => {
+      apiLog.log(data);
+    });
+
+    return apiLog;
+  }
+
+  buildMap (grid) {
```
curioussavage: this is a nice touch here :laughing:

ajacksified: Gotta have blinky `X`s on a map, or it isn't a real dashboard.

src/server/index.jsx
```diff
@@ -134,6 +136,13 @@ class Server {
         secureProxy: config.httpsProxy
     };
 
+    let that = this;
+    server.use(function * (next) {
+      that.activeRequests++;
+      yield next;
+      that.activeRequests--;
```
curioussavage: Can you tell me how this works?  when does line 143 get called?  (fuzzy understanding of generators here.)

package.json
```diff
@@ -8,7 +8,8 @@
     "build": "gulp",
     "watch": "gulp watch",
     "dev": "gulp dev",
-    "test": "mocha test/index.js"
+    "test": "mocha test/index.js",
+    "console": "babel-node ./src/server/console.es6.js"
```
curioussavage: when would I use this?  should this be running index.jsx?

curioussavage:
:nail_care:  pretty much a fish, I just have questions, maybe add some inline documentation for the next guy?

package.json
ajacksified: Ah, yeah, leftovers.

src/server/index.jsx
ajacksified: 143 gets called after all the other generators have finished yielding (basically, at the end of a response, just before it's sent to a browser.)

ajacksified:
:haircut:

curioussavage:
:fish:

PR #377, opened 2015-11-04
Interstial: fixes interstial for authenticated users

curioussavage:
:eyeglasses:  @ajacksified 

renamed userPrefs to preferences in an earlier commit (routes.jsx), missed all the other spots :p


ajacksified:
:fish:

PR #378, opened 2015-11-05
Add statsd monitoring of servers

ajacksified:
:eyeglasses: @umbrae, @spladug

This adds:
* Total requests
* Request timings
* Total active requests
* Response code

Buffered to a 100ms timeout by default, configurable.

Questions:
* Should I sample?

### DIFF
--- a/src/server/index.jsx
+++ b/src/server/index.jsx
@@ -12,6 +12,7 @@ import csrf from 'koa-csrf';
 import compress from 'koa-compress';
 
 import session from 'koa-session';
+import StatsdClient from 'statsd-client';
 
 import { ServerReactApp } from 'horse-react';
 import mixin from '../app-mixin';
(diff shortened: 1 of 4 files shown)


src/server/index.jsx
```diff
@@ -136,11 +137,43 @@ class Server {
...
+
     let that = this;
     server.use(function * (next) {
+      let statsd = that.statsd;
+
       that.activeRequests++;
+
+      if (statsd) {
+        statsd.increment('request');
+        statsd.increment('activeRequests');
```
spladug: Counters aren't really meant for this sorta thing; if any request takes longer than the flush interval (10s) then it'll be lost in the flush and you could end up with an inaccurate (even negative!) request count next interval.  I'd say you could potentially use gauges, but I'm pretty wary of them too since a worker dying mid-request would leave the gauge permanently out of whack (and [tallier doesn't support them yet anyway](https://github.com/reddit/tallier/pull/10)). 

It might work better to keep a global counter in your worker process and just flush a straight up count every 10 seconds.

src/server/index.jsx
```diff
@@ -136,11 +137,43 @@ class Server {
...
+      }
+
       yield next;
       that.activeRequests--;
+
+      if (statsd) {
```
spladug: Having to check this everywhere is a bummer, would it be possible to make a client that just bitbuckets any metrics if no host is configured? so downstream code needn't care about if stats are configured or not.

spladug:
:nail_care:

src/server/index.jsx
ajacksified: Yeah, that'd be cleaner. In fact, the client may do that already, I'll take a look.

src/server/index.jsx
ajacksified: Alright - I've got something similar in place for the dev console anyway, so I'll instantiate a statsd client on the master process and flush it out from there, like I do to the console.

ajacksified:
:haircut:

ajacksified:
@spladug I went for 1s rather than 10s resolution on activerequests; should I go with 10 instead?

spladug:
@ajacksified yes, graphite and statsd both have a minimum resolution of 10s.

ajacksified:
@spladug done

spladug:
looks fine to me, but i'm dumb at es6  :fish:

umbrae:
:fish: lgtm I think!

PR #379, opened 2015-11-06
Fix infinite loop

ajacksified:
Fix infinite loop caused by an `undefined` dataCache value.

:eyeglasses: @curioussavage


curioussavage:
:fish:  yes please.

PR #380, opened 2015-11-09
Still decrement requests even if middleware fails

ajacksified:
:eyeglasses: @curioussavage 

If middleware (such as route handlers) broke on an uncaught exception, it didn't decrement the active requests value, giving us a potentially incorrect `active` count.


curioussavage:
:fish:

PR #381, opened 2015-11-09
503s

ajacksified:
:eyeglasses: @curioussavage 

First commit unrelated, let me know if I should just break it out instead. To be used in conjunction with https://github.com/reddit/snoode/pull/60.


curioussavage:
:fish:  :+1:  it looks much cleaner

PR #383, opened 2015-11-10
Routes: abstract construction of body function into a method

curioussavage:
:eyeglasses: @ajacksified

### DIFF
--- a/src/views/pages/userProfile.jsx
+++ b/src/views/pages/userProfile.jsx
@@ -9,13 +9,13 @@ import UserProfile from '../components/UserProfile';
 
 class UserProfilePage extends BasePage {
   render() {
-    if (!this.state.data || !this.state.data.userProfile) {
+    if (!this.state.data || !this.state.data.user) {
       return (
         <Loading />
       );
     }
 
-    var userProfile = this.state.data.userProfile;
+    var userProfile = this.state.data.user;
     var name = this.props.userName;
 
     return (


src/routes.jsx
```diff
@@ -188,6 +188,17 @@ function userData(ctx, app) {
   }
 }
 
+function makeBody(props, key, track, Component, SecondComponent=null) { 
+  return () => {
+    return (
+          <BodyLayout { ...props }>
```
ajacksified: de-indent a bit

src/routes.jsx
```diff
@@ -188,6 +188,17 @@ function userData(ctx, app) {
...
+            { SecondComponent }
+            <Component { ...props } key={ key } track={track} />
```
ajacksified: Why don't we just set a `key` in buildProps to a guid, instead of passing in a key every time? We always want a route to re-render anyways.

src/routes.jsx
```diff
@@ -188,6 +188,17 @@ function userData(ctx, app) {
   }
 }
 
+function makeBody(props, key, track, Component, SecondComponent=null) {
```
ajacksified: Is there a way to make `Component` an array instead of sending in two parameters? That would allow us to order, and add `n` pre- or post- components.

ajacksified:
Good idea, :nail_care:

src/routes.jsx
curioussavage: oh cool. good idea!

src/routes.jsx
curioussavage: if we always want the route to render again is there any reason to even add a key?

ajacksified: Yes, if you don't have a key, it won't re-render the page, even with new props.

curioussavage:
:haircut:

src/routes.jsx
```diff
@@ -396,19 +378,17 @@ function routes(app) {
...
-        <li className='TextSubNav-li' active={activityActive}>
-          <a className={`TextSubNav-a ${activityActive}`} href={`/u/${userName}/activity`}>Activity</a>
-        </li>
-        <li className='TextSubNav-li' active={gildActive}>
-          <a className={`TextSubNav-a ${gildActive}`} href={`/u/${userName}/gild`}>Give gold</a>
-        </li>
-      </TextSubNav>
-    );
+    return [
+      <li className='TextSubNav-li' active={aboutActive}>
```
ajacksified: while you're at it, these need `key`s.

src/routes.jsx
```diff
@@ -269,15 +280,7 @@ function routes(app) {
...
-    var key = this.key;
-
-    this.body = function(props) {
-      return (
-        <BodyLayout { ...props }>
-          <IndexPage { ...props } key={ key } track='listings' />
-        </BodyLayout>
-      );
-    };
+    this.body = makeBody(this.props, [{c: IndexPage, props: this.props}]);
```
ajacksified: Tuples might be cleaner: `[[IndexPage, this.props], [...]]`. You can then destructure within makeBody using `const [component, props] = comp`. What do you think?

curioussavage: yeah, if you don't mind nested arrays then I would prefer that too.

ajacksified: :+1:

curioussavage:
:haircut:

ajacksified:
:fish:

PR #385, opened 2015-11-10
Comment: make deeply nested flair scrollable.

curioussavage:
:eyeglasses: @ajacksified 

forgot to solve the actual issue. Which is very rare with the new layout for flair. (only affected comments 5  levels deep and up with max length flair)


ajacksified:
:fish:

PR #386, opened 2015-11-11
React 14x

ajacksified:
:eyeglasses: @curioussavage 

Woo

First commit: get things working, by moving `getDOMNode`.
Second commit: remove addons warnings, and in fact, remove all the addons since they're completely unnecessary on mobile anyways.
Third commit: remove dead code, as react handles touch events natively, and it wasn't being used anyways.


ajacksified:
depends on https://github.com/reddit/horse-react/pull/9

ajacksified:
@curioussavage I'm having trouble running tests here _and_ on master - it doesn't appear to like `static proptypes` inside of classes, which is weird, because babel should be totally fine with that. Any ideas?

ajacksified:
If the code changes look good, I'll make a pass at design/tiffany/steve to confirm that removing animations is :ok_hand:

curioussavage:
oh. oops, need to whitelist property initializers and the other es7 stuff in test/index.js or go to stage 2.

ajacksified:
Yeah, stage 2 is probably best. As an aside, the babel 6 upgrade is a deep
rabbit hole.

curioussavage:
yay!! stateless function components!

curioussavage:
if we aren't animating SVGs anymore can we just have the component return an image tag linking to them?

curioussavage:
could we just make a babelrc file or put babel options in the package.json, then we shouldn't have to specify the options in several places.

curioussavage:
:nail_care: code looks good. only thing would be tests (dont run on stage 2
)

ajacksified:
@curioussavage -

* Good point on the SVGs, I'll update.
* I agree on babelrc - but let's do that on a babel upgrade PR instead.
* I'll fix test stage

ajacksified:
On second thought, I'l leave the SVG changes to another PR. :haircut:

curioussavage:
:fish:  sounds good to me

ajacksified:
@tdohz and @spez: this PR removes the fading transitions on menus, and the slide-down animation on the sidebar. They didn't seem necessary on mobile, but before I merge this you may want to review.

Old - New

animations

spez:
:fish:

PR #387, opened 2015-11-11
Fix word break on listing titles

ajacksified:
:eyeglasses: @curioussavage 

Fixes word break so that it breaks in the middle of long words or at spaces, but not in the middle of short words.


curioussavage:
:fish:  definitely better

PR #389, opened 2015-11-13
Filter out 429s and timeouts from logs

ajacksified:
:eyeglasses: @umbrae and/or @spladug


umbrae:
:fish: Small whitespace issue but otherwise looks good

PR #390, opened 2015-11-13
Add timeouts for superagent requests

ajacksified:
:eyeglasses: @umbrae, @curioussavage 

This covers all of our superagent requests. Bad hivemind and oauth requests will hang for 4 minutes each, without timeouts set.

Next step: fork superagent and allow setting a default timeout. Also, implement response timeouts in the http server that will send back a timeout if the server's taking to long (although we still need to kill the open http requests using superagent timeouts.)


curioussavage:
:fish: lgtm

PR #391, opened 2015-11-16
Don't set ref on CommentPreviews

ajacksified:
You cannot set a `ref` on a stateless component - this was throwing
warnings. This commit also fixes the ref to `ul` on TextSubNav, which
was throwing errors on the same page.

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #392, opened 2015-11-17
Fixes test warnings, and adds pre-commit and pre-push hooks

ajacksified:
:eyeglasses: @curioussavage 

Next step, to add a git pre-commit hook to auto-run tests on commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,4 +80,6 @@ Getting Up and Running
     4. Repeat recursively if you need to work on a dependency's dependency
     5. Re-run the build and restart your server (`npm run watch` *will* watch
       linked files)
+8. Commit hooks - symlink `hooks` into `.git/hooks` (`ln -s -f ../hooks .git/hooks`), which
+  will run some safety checks before committing and pushing code.
 
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -5,7 +5,7 @@ exec 1>&2
 exec < /dev/tty
 
 # Check if a console.log was added
-consoleregexp='\+*console.log'
+consoleregexp='\+*console\.log'
 
 # CHECK
 if test $(git diff --cached | grep $consoleregexp | wc -l) != 0
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,22 @@
+#!/bin/sh
+# Redirect output to stderr.
+exec 1>&2
+# enable user input
+exec < /dev/tty
+
+# Check if a console.log was added
+consoleregexp='\+*console.log'
+
+# CHECK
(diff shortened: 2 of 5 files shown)


curioussavage:
:fish:

ajacksified:
:haircut: - added git hooks that'll:

* yell at you if you add console.logs (we may want to have this run the linter, too)
* automatically run tests before pushing to `master` or `staging`

ajacksified:
Also :eyeglasses: @umbrae

README.md
```diff
@@ -80,4 +80,6 @@ Getting Up and Running
     4. Repeat recursively if you need to work on a dependency's dependency
     5. Re-run the build and restart your server (`npm run watch` *will* watch
       linked files)
+8. Commit hooks - copy the contents of `./hooks/` into `./.git/hooks`, which
```
umbrae: Should we have this be a symlink instead, so that if we make updates it will still be respected?

hooks/pre-commit
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# Redirect output to stderr.
+exec 1>&2
+# enable user input
+exec < /dev/tty
+
+# Check if a console.log was added
+consoleregexp='\+*console.log'
```
umbrae: nit: that dot would need escaping if we want to be perfect about it:

```shell
➜  ~  ls | grep bla..py
blah.py
```

hooks/pre-push
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# Runs tests before pushing to master or staging
+
+BRANCH=`git rev-parse --abbrev-ref HEAD`
+PUSH_COMMAND=`ps -ocommand= -p $PPID`
+PROTECTED_BRANCHES="^(master|staging)"
+
+if [[ "$BRANCH" =~ $PROTECTED_BRANCHES ]]; then
+  eval npm test
```
umbrae: Maybe we should have an echo here saying what is happening in case it bombs in a really unreadable way somehow? (Like if they haven't set up npm at all?). Maybe "Running pre-push tests with `npm test` (skip with --no-verify)"?

umbrae:
:nail_care: cool

hooks/pre-push
ajacksified: If they haven't set up npm at all, they didn't follow the readme, didn't install any dependencies, can't run the web server, and shouldn't be pushing _anyways_. I don't want to build a skip in, since this is only for `master` and `staging` anyways.

ajacksified:
:haircut:

hooks/pre-push
umbrae: good points

umbrae:
:fish:

PR #393, opened 2015-11-17
Set defaultConfig as constant in server

ajacksified:
Only get the env vars at startup time; it's unnecessary to get them
every request.

:eyeglasses: @curioussavage


curioussavage:
:fish:

PR #394, opened 2015-11-18
Configurable (and increased) API timeouts

ajacksified:
Put default API timeout (which cancels the request and returns a 504 if exceeded) into constants so that it's defined in one place. (One exception: hivemind requests, which we're okay with failing if they take
over a few seconds.)

Also increase API timeout to 10s to accomodate slow requests, like the google crawler. It was set to 5s previously when we thought that api requests were causing the request queue to back up, which does not appear to be the case.

:eyeglasses: @curioussavage

### DIFF
--- a/src/app-mixin.jsx
+++ b/src/app-mixin.jsx
@@ -12,6 +12,7 @@ import ErrorPage from './views/pages/error';
 import Loading from './views/components/Loading';
 
 import errorLog from './lib/errorLog';
+import constants from './constants';
 
 function logError(err, ctx, config) {
   var userAgent;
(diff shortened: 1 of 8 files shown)


src/app-mixin.jsx
```diff
@@ -71,6 +72,7 @@ function mixin (App) {
       let app = this;
       // Set up two APIs (until we get non-authed oauth working).
       this.api = new V1Api({
+        timeout: constants.DEFAULT_API_TIMEOUT,
```
curioussavage: is there a PR for snoode? or does it already accept this arg?

curioussavage:
:nail_care:

src/app-mixin.jsx
ajacksified: There is - https://github.com/reddit/snoode/pull/62

ajacksified:
:haircut:

curioussavage:
:fish:

PR #396, opened 2015-11-18
Infobar

curioussavage:
:eyeglasses: @ajacksified 

copy might change for the EU cookie message


assets/js/client.es6.js
```diff
@@ -97,10 +97,14 @@ function modifyContext (ctx) {
...
   let app = this;
 
+  const EUCookie = parseInt(cookies.get('EUCookieNotice')) || 0;
+
   ctx = Object.assign({}, baseCtx, ctx, {
     dataCache: app.getState('dataCache') || {},
     compact: (cookies.get('compact') || '').toString() === 'true',
     showOver18Interstitial: (cookies.get('over18') || 'false').toString() === 'false',
+    showEUCookieMessage: (EUCookie < 3) && this.getState('showEUCookieMessage'),
+    hideGlobalMessage: !!cookies.get('hide_global_message'),
```
ajacksified: should this change when the message changes? wouldn't this hide _all_ global messages instead of just the specific message the user is trying to hide?

assets/less/components/InfoBar.less
```diff
@@ -0,0 +1,28 @@
...
+  .panel {
+    margin-botom: 0;
+  }
+}
+
+.infobar-text {
+  margin: 3px 10px 0 0;
+}
+
+.infobar-html {
```
ajacksified: nuke if empty

assets/less/components/InfoBar.less
```diff
@@ -0,0 +1,28 @@
...
+  margin-top: @top-nav-height;
+}
+
+.infobar {
+  .shadow;
+  border-radius: 3px;
+  padding: 5px 8px 8px;
+  background-color: @white;
+  
+  .panel {
```
ajacksified: an `infobar-panel` class would be preferable to nesting

assets/less/components/InfoBar.less
```diff
@@ -0,0 +1,28 @@
...
+    margin-botom: 0;
```
ajacksified: `bottom`

src/constants.es6.js
```diff
@@ -10,4 +10,11 @@ export default {
   SCROLL: 'scroll',
   ICON_SHRUNK_SIZE: 16,
   CACHEABLE_COOKIES: ['compact'],
+  HIDE_GLOBAL_MESSAGE: 'hideGlobalMessage',
+  EU_COOKIE_NOTICE_SEEN: 'euCookieNoticeSeen',
+  NEW_INFOBAR_MESSAGE: 'newInfoBarMessage',
+  messageTypes: {
+    GLOBAL: 'global',
+    EU_COOKIE: 'euCookie',
+  }
```
ajacksified: needs `,`

src/routes.jsx
```diff
@@ -253,6 +271,11 @@ function routes(app) {
   router.use(loadData);
   router.use(defaultLayout);
 
+  router.use(function * (next) {
```
ajacksified: if you make globalMessage a generator, you can just `router.use(globalMessage)`, and use `this` instead of passing in `ctx`

src/server/index.jsx
```diff
@@ -368,13 +369,18 @@ class Server {
...
       this.userAgent = this.headers['user-agent'];
+      this.country = this.headers['cf-ipcountry'];
 
       if (!this.token) {
         this.token = this.cookies.get('token');
         this.tokenExpires = this.cookies.get('tokenExpires');
       }
 
-      this.showOver18Interstitial = !this.cookies.get('over18') || this.cookies.get('over18') === 'false';
+      let EUCookie = parseInt(this.cookies.get('EUCookieNotice')) || 0 ;
```
ajacksified: `const` rather than `let`

src/server/index.jsx
```diff
@@ -368,13 +369,18 @@ class Server {
...
+      this.showEUCookieMessage = (EUCookie < 3) &&
```
ajacksified: render this client-side only, because of caching

src/views/components/InfoBAr.jsx
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
```
ajacksified: rename to `InfoBar` instead of `InfoBAr`

src/views/components/InfoBAr.jsx
```diff
@@ -0,0 +1,146 @@
...
+import BaseComponent from './BaseComponent';
+
+import constants from '../../constants';
+
+const PropTypes = React.PropTypes;
+
+const cookie_message = 'Reddit and our partners use cookies to help ' +
+  'personalize Reddit, remember your preferences, measure Reddit\'s ' +
+  'performance, serve you better ads, and have a safer Reddit experience. ' + 
+  'By using Reddit, you agree to our ';
```
ajacksified: `EU_COOKIE_MESSAGE` for consistency

src/views/components/InfoBAr.jsx
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  close() {
+    let { app } = this.props;
+    let message = this.state.messages[0];
+
+    if (message.type === constants.messageTypes.GLOBAL) {
+      app.emit(constants.HIDE_GLOBAL_MESSAGE, new Date(message.expires));
+    } else if (message.type === constants.messageTypes.EU_COOKIE) {
+      app.emit(constants.EU_COOKIE_NOTICE_SEEN, 3);
```
ajacksified: This feels a little hacky; how do you feel about emitting `EU_COOKIE_NOTICE_CLOSE`, which sets a `eu_cookie_notice_close` cookie, instead of checking if it's seen 3 times? (And then, when it _is_ seen three times, also set that close cokie.)

src/views/components/InfoBAr.jsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+    
+    let messages = this.state.messages.slice(1);
+    this.setState({ messages });
+  }
+
+  render() {
+    let messages = this.state.messages;
```
ajacksified: const

src/views/components/InfoBAr.jsx
```diff
@@ -0,0 +1,146 @@
...
+    let message = messages[0];
+    let btnText = '&times';
```
ajacksified: const

src/views/layouts/BodyLayout.jsx
```diff
@@ -12,11 +14,22 @@ class BodyLayout extends BasePage {
 
   render () {
     let { user, userSubscriptions } = this.state.data;
+    let { app, globalMessage, showEUCookieMessage } = this.props;
```
ajacksified: consts

src/views/pages/index.jsx
```diff
@@ -32,7 +33,7 @@ class IndexPage extends BasePage {
   render() {
     var loading;
     var props = this.props;
-    var { data, compact } = this.state;
+    var { data, compact, showInfoBar} = this.state;
```
ajacksified: const

ajacksified:
:nail_care:

src/views/components/InfoBAr.jsx
curioussavage: Don't know how much this matters, but without storing views in a cookie you would have to see the cookie 3 times in the same session for that to count. with the cookie its 3 times ever

ajacksified: Yeah, that's true. Perhaps slightly better would be to put the `3` into constants, like EU_COOKIE_HIDE_AFTER_VIEWS or something?

src/routes.jsx
curioussavage: but what if we want to test the middleware?

ajacksified: You'd just need to call it as a generator.

curioussavage:
:haircut:

assets/js/client.es6.js
```diff
@@ -421,6 +428,30 @@ function initialize(bindLinks) {
     cookies.set('over18', val)
   });
 
+  app.on(constants.HIDE_GLOBAL_MESSAGE, function(message) {
+    let options = {
+      expires: new Date(message.expires),
+    };
+    cookies.set(message.key, 'globalMessageSeen', options);
```
ajacksified: May want to throw an error if message.key is undefined

src/views/components/InfoBar.jsx
```diff
@@ -0,0 +1,144 @@
...
+  render() {
+    const message = this.state.messages[0];
+    let content;
+
+    if (message) {
+      if (message.type === constants.messageTypes.EU_COOKIE) {
+        content = (
+          <p>
+            { EU_COOKIE_MESSAGE }
+            <a href='https://www.reddit.com/help/privacypolicy'>privacy policy</a>
```
ajacksified: why is this separate from EU_COOKIE_MESSAGE? I don't think we can assume all messages will have the link to privacy policy. you could include this in EU_COOKIE_MESSAGE as markdown and use reddit-text-js to process the markdown and set html.

src/views/components/InfoBar.jsx
```diff
@@ -0,0 +1,144 @@
...
+          </p>
+        );
+      }
+
+      if (message.html) {
+        content = (
+          <div
+            className='infobar-html'
+            dangerouslySetInnerHTML={{__html: message.html}}
```
ajacksified: I think we should only allow markdown - what are your thoughts?

assets/js/client.es6.js
curioussavage: throwing in the config now

curioussavage:
:haircut:

src/views/components/InfoBar.jsx
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  componentDidMount() {
+    let { app, messages } = this.props;
+    let hasEUCookie = this.checkForEUCookie(messages);
+
+    app.on(constants.NEW_INFOBAR_MESSAGE, this.handleNewMessage);
+
+    if (hasEUCookie) {
+      app.on("route:start", this.incrementCookieNoticeSeen);
```
ajacksified: single quotes here and below

ajacksified:
:fish:, with the caveat that we should figure out why page:update is called twice

PR #398, opened 2015-11-19
Fix subredditAboutPage

curioussavage:
It appears that the div with html needs to be self closing.

:eyeglasses: @ajacksified 

weird...


ajacksified:
:fish:

PR #400, opened 2015-11-23
Fix userProfile loading from `user` instead of `userProfile

ajacksified:
:eyeglasses: @curioussavage 

This is currently loading the logged-in user's data instead of the
profile you are attempting to view. This uses the correct property.


curioussavage:
:fish:

PR #401, opened 2015-11-23
Add `key` to components in `makeBody`

ajacksified:
:eyeglasses: @curioussavage 

Default props include a random key, which is passed down to components.
If more than one component child has a key, React throws a warning.


curioussavage:
:fish:

ajacksified:
248251bb3e6e6246b9d6df3ce1bc31ad4dbb27ac

PR #402, opened 2015-11-23
Move `track` prop to getter on page element

ajacksified:
:eyeglasses: @curioussavage 

It's both easier to keep track of, and harder to accidentally mess up, when it's a static property or a getter instead of passed in via props. Since it's page-specific (used in basepage), it makes sense to be here. This also fixes an issue where both bodylayout and the page are calling the tracking pixel.


curioussavage:
could it just be a static prop?

`static track = 'string'`

curioussavage:
:nail_care:

ajacksified:
@curioussavage - I'm using a `getter` instead of `static` so I can just call `this.track` instead of `this.__constructor.track`.

ajacksified:
:haircut:

curioussavage:
:fish:

ajacksified:
fd6f4e0a25ce7dc7ccf83510c6da7b8ebd778835

PR #404, opened 2015-11-24
Content nsfw fixes

curioussavage:
:eyeglasses: @ajacksified

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -51,7 +51,6 @@ var $body = document.body || document.getElementsByTagName('body')[0];
 var $head = document.head || document.getElementsByTagName('head')[0];
 
 var config = defaultConfig();
-const dataCache = window.bootstrap.dataCache;
 
 function loadShim() {
   var shimScript = document.createElement('script');
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -51,6 +51,7 @@ var $body = document.body || document.getElementsByTagName('body')[0];
 var $head = document.head || document.getElementsByTagName('head')[0];
 
 var config = defaultConfig();
+const dataCache = window.bootstrap.dataCache;
 
 function loadShim() {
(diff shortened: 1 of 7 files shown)


assets/js/client.es6.js
```diff
@@ -239,13 +241,13 @@ function initialize(bindLinks) {
...
 
-  if (window.bootstrap.dataCache.user) {
-    app.setState('user', window.bootstrap.dataCache.user);
-    app.setState('preferences', window.bootstrap.dataCache.preferences);
+  if (dataCache.user) {
+    app.setState('user', dataCache.user);
+    app.setState('preferences', dataCache.preferences);
 
-    cookies.set('over18', window.bootstrap.dataCache.preferences.over_18);
+    cookies.set('over18', dataCache.preferences.body.over_18);
```
curioussavage: this was one of the big issues here. the preferences are nested in body.

assets/js/client.es6.js
```diff
@@ -51,6 +51,7 @@ var $body = document.body || document.getElementsByTagName('body')[0];
 var $head = document.head || document.getElementsByTagName('head')[0];
 
 var config = defaultConfig();
+const dataCache = window.bootstrap.dataCache;
```
ajacksified: nuke here, since you defined `dataCache` within `initialize`

ajacksified:
:nail_care:

ajacksified:
Also, it doesn't show NSFW images (good) but it _also_ doesn't show the blurred pixel version (bad); it only shows the blue placeholder.

curioussavage:
do you have an example? I am looking around for a non nsfw sub with nsfw images where there is a preview and I cant find one :(

curioussavage:
Ok, so it will show the blurry images but only if 'make safer for work' is disabled. otherwise those images are not returned in the api response. (maybe the api should return that?)

:haircut:

ajacksified:
:fish:

PR #405, opened 2015-11-24
Remove user activity 'Old' sort

schwers:
https://reddit.atlassian.net/browse/MOBILEWEB-147

Removes the ‘Old’ sort from user activity page. This doesn’t work as
the api doesn’t support the ‘old’ sort.

Now looks like
image

:eyeglasses: @ajacksified

### DIFF
--- a/src/views/pages/userActivity.jsx
+++ b/src/views/pages/userActivity.jsx
@@ -19,18 +19,16 @@ class UserActivityPage extends BasePage {
       );
     }
 
-    var props = this.props;
-    var state = this.state;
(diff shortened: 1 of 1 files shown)


ajacksified:
This change removes "old" from the comments page; it should only be removed from user activity pages. :nail_care:

schwers:
:haircut: changed to use a new list in `SortDropdown.jsx`

ajacksified:
:fish: - and if you care to do some cleanup, those `var`s in `render()` could be `const` instead.

PR #407, opened 2015-12-01
Respect compact preference everywhere. Fix search pagination

schwers:
https://reddit.atlassian.net/browse/MOBILEWEB-329

Creates a new component `ListingContainer` to manage the compact state of `ListingList`. Passes the initial compact state everywhere so search pages, user saved, and user submitted posts respect the preference.

https://reddit.atlassian.net/browse/MOBILEWEB-315
Creates a new component `ListingPaginationButtons` to manage the next|prev buttons on listings. Fixes generating pagination buttons for the search page so they appear.

:eyeglasses: @ajacksified

### DIFF
--- a/src/views/pages/search.jsx
+++ b/src/views/pages/search.jsx
@@ -3,7 +3,7 @@ import constants from '../../constants';
 import querystring from 'querystring';
 
 import BasePage from './BasePage';
-import ListingList from '../components/ListingList';
+import ListingContainer from '../components/ListingContainer';
 import Loading from '../components/Loading';
 import SearchSortSubnav from '../components/SearchSortSubnav';
 import SearchBar from '../components/SearchBar';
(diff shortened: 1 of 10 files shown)


src/views/pages/search.jsx
```diff
@@ -20,7 +20,18 @@ class SearchPage extends BasePage {
...
     }
 
+    this.state.compact = props.compact;
+
     this._lastQueryKey = null;
+    this._onCompactToggle = this._onCompactToggle.bind(this);
+  }
+
+  componentDidMount() {
+      this.props.app.on(constants.COMPACT_TOGGLE, this._onCompactToggle);
```
ajacksified: de-indent

ajacksified: I wonder if we should just have `listinglist` listen to the event instead, so that we don't have to remember to use it in every page.

src/views/pages/search.jsx
```diff
@@ -122,21 +137,22 @@ class SearchPage extends BasePage {
...
-      var listings = searchResults.links || [];
-      var noListResults = listings.length === 0;
-      var noSubResults = subreddits.length === 0;
-      var noResult = noSubResults && noListResults;
-      var subredditResultsOnly = props.subredditName && props.ctx.query.q;
+      const subreddits = searchResults.subreddits || [];
+      const listings = searchResults.links || [];
+      const noListResults = listings.length === 0;
+      const noSubResults = subreddits.length === 0;
+      const noResult = noSubResults && noListResults;
```
ajacksified: Because `const` is block-scoped, it won't be available outside of this `if` statement (unlike `var`, which  hoists it to the top of the function.) This results in a `noResult is undefined` error upon rendering the search page.

ajacksified:
:nail_care:

ajacksified:
:fish:

PR #408, opened 2015-12-02
Dnt gtm

dwick:
:eyeglasses: @ajacksified 

depends on https://github.com/reddit/reddit-public/pull/4448

### DIFF
--- a/src/config.es6.js
+++ b/src/config.es6.js
@@ -38,10 +38,12 @@ function config() {
     reddit: process.env.REDDIT || 'https://www.reddit.com',
 
     googleAnalyticsId: process.env.GOOGLE_ANALYTICS_ID,
+    googleTagManagerId: process.env.GOOGLE_TAG_MANAGER_ID,
 
     loginPath: process.env.LOGIN_PATH || '/oauth2/login',
 
     statsURL: process.env.STATS_URL || 'https://stats.redditmedia.com/',
+    mediaDomain: process.env.MEDIA_DOMAIN || 'www.redditmedia.com',
     adsPath: process.env.ADS_PATH ||  '/api/request_promo.json',
     manifest: {},
 
--- a/src/lib/dnt.es6.js
+++ b/src/lib/dnt.es6.js
@@ -1,0 +1,15 @@
+'use strict';
+
+/* Normalize DNT across browser vendors */
+
+const RE_IE_VERSION = /(?:\b(?:MS)?IE\s+|\bTrident\/7\.0;.*\s+rv:)(\d+(?:\.?\d+)?)/i;
+
+if (global.navigator) {
+  const uaMatches = global.navigator.userAgent.match(RE_IE_VERSION);
+  const ieVersion = uaMatches && uaMatches[1] && parseFloat(uaMatches[1]);
+  const doNotTrack = global.navigator.doNotTrack ||
+    global.doNotTrack ||
+    global.navigator.msDoNotTrack;
+
+  global.DO_NOT_TRACK = /^(yes|1)$/i.test(doNotTrack) && ieVersion !== 10;
+}
--- a/src/lib/dnt.es6.js
+++ b/src/lib/dnt.es6.js
@@ -1,6 +1,7 @@
 'use strict';
 
-/* Normalize DNT across browser vendors */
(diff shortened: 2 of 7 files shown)


src/config.es6.js
```diff
@@ -38,9 +38,11 @@ function config() {
     reddit: process.env.REDDIT || 'https://www.reddit.com',
 
     googleAnalyticsId: process.env.GOOGLE_ANALYTICS_ID,
+    googleTagManagerId: process.env.GOOGLE_TAG_MANAGER_ID,
 
     loginPath: process.env.LOGIN_PATH || '/oauth2/login',
 
+    mediaDomain: process.env.MEDIA_DOMAIN || 'www.redditmedia.com',
```
ajacksified: Can you add a protocol to this, to match the other `DOMAIN` config values? (even if it's just `//` to avoid http/https protocol switching)

src/lib/dnt.es6.js
```diff
@@ -0,0 +1,15 @@
...
+const RE_IE_VERSION = /(?:\b(?:MS)?IE\s+|\bTrident\/7\.0;.*\s+rv:)(\d+(?:\.?\d+)?)/i;
+
+if (global.navigator) {
+  const uaMatches = global.navigator.userAgent.match(RE_IE_VERSION);
+  const ieVersion = uaMatches && uaMatches[1] && parseFloat(uaMatches[1]);
+  const doNotTrack = global.navigator.doNotTrack ||
+    global.doNotTrack ||
+    global.navigator.msDoNotTrack;
+
+  global.DO_NOT_TRACK = /^(yes|1)$/i.test(doNotTrack) && ieVersion !== 10;
```
ajacksified: Please add a comment as to why IE 10 is specifically called out here.

src/views/layouts/DefaultLayout.jsx
```diff
@@ -60,6 +62,27 @@ function DefaultLayout  (props) {
     );
   }
 
+  let gtmTracking;
+
+  if (props.config.googleTagManagerId) {
```
ajacksified: Also check for `mediaDomain`

src/views/layouts/DefaultLayout.jsx
```diff
@@ -60,6 +62,27 @@ function DefaultLayout  (props) {
...
+    let gtmCode = `
```
ajacksified: `const`

src/lib/dnt.es6.js
ajacksified: Instead of using window, could this file return a function that, when called, returns true/false? We could then use it in `client.es6.js` to set a config value that `DefaultLayout` could read, and either include the tracking code or not. That way, we're not relying on globals, and we can make this testable.

ajacksified:
:nail_care:

src/lib/dnt.es6.js
dwick: Since this is only a polyfill (and someday should be removed in favor of the [standard](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack) I'm inclined to want to leave it.  If you want I could use `navigator.doNotTrack` instead of `window.DO_NOT_TRACK` so this can just be removed at a later date?

ajacksified: Yeah, I suppose that, as a shim, it's fair to leave it this way.

src/config.es6.js
dwick: mind if i do the opposite change for `statsDomain` since it's more than a domain?

dwick:
:haircut: and [question](https://github.com/reddit/reddit-mobile/pull/408#discussion_r46498377)

src/config.es6.js
ajacksified: You're right, it's more than a domain... but it also means we end up copying the protocol everywhere it's used. I suppose adding a scheme and renaming both from `DOMAIN` to `URI` is most correct.

dwick: done https://github.com/reddit/reddit-mobile/pull/410

ajacksified:
:fish:

dwick:
:haircut: added tracking: 0cfc73d

ajacksified:
more :fish:es

PR #409, opened 2015-12-02
Eu galletas

curioussavage:
:eyeglasses: @ajacksified 

changes:
* view count for cookie starts at one like desktop.
* moved code to set cookie into infobar so it runs before modify context. plus for simplicity as its not used by any other components.
* added a key to bodylayout 
* use lodash to find/filter messages. (my function had a bug :( )

up on staging now

### DIFF
--- a/src/views/components/InfoBar.jsx
+++ b/src/views/components/InfoBar.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import process from 'reddit-text-js';
+import cookies from 'cookies-js';
+import { find as _find, filter as _filter } from 'lodash/collection';
 
 import BaseComponent from './BaseComponent';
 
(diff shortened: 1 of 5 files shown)


src/views/components/InfoBar.jsx
```diff
@@ -44,28 +48,43 @@ class InfoBar extends BaseComponent {
...
+  incrementCookieNoticeSeen(e, num=1) {
+    const { app } = this.props;
+    const oldCookie = parseInt(cookies.get('EUCookieNotice')) || 0;
+    let options = {};
+
+    let date = new Date();
+    date.setFullYear(date.getFullYear() + 2);
+    options.expires = date;
+
+    var newNum;
```
ajacksified: `let`

ajacksified:
The tiniest of nits; :fish: after that fix.

As an aside, I'm not a big fan of allowing components to manage cookies- since all of our other cookie logic flows through props, rather than accessing the cookies directly; keeping it one place seems cleaner. I don't feel like it's a strong enough argument to hold this up, but we should revisit how we manage cookies in the future.

PR #411, opened 2015-12-04
Added Android Mobile Web App Compatability

TheMightyCraken:
More Mobile Features :)

Also, the Theme Color on Android has now changed and it is Mobile Web App Compatable :)


src/views/layouts/DefaultLayout.jsx
```diff
@@ -68,11 +68,15 @@ function DefaultLayout  (props) {
...
+        <meta name='theme-color' content='#483D8B' />
+        <meta name='apple-mobile-web-app-capable' content="yes" /> 
+        <meta name="mobile-web-app-capable" content="yes" />
+        <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent" />
+        <meta name="apple-mobile-web-app-title" content="Reddit Mobile" />
         <meta id='csrf-token-meta-tag' name='csrf-token' content={ props.ctx.csrf } />
         { metaDescription }
-
+        
+        <link href={ `${assetPath}/favicon/64x64.png` } rel="apple-touch-icon-precomposed"sizes="76x76 />
```
Deimos: Looks like you're missing the closing double-quote here, so this is invalid HTML currently. There should be a space before `sizes` as well.

ajacksified: (Also, in our styleguide, we use single quotes.)

ajacksified:
Thanks for the submission! We originally had this in place, but we found that it actually made the browsing experience _worse_, rather than _better_, because it hides things like the url bar, makes opening things in new tabs more painful, and other small tweaks that take the webbiness out of the website. (It would probably work better in other websites that try to act more like apps.)

We'll close this because of the usability issues, but thanks anyways!

PR #412, opened 2015-12-07
CommentBox: refactor and add error handling

curioussavage:
:eyeglasses: @schwers 

with https://github.com/reddit/snoode/pull/63 which you could look at too if jack already hasn't

changes: 
* displays error message
* post button now disabled when form is empty instead of hidden.

### DIFF
--- a/src/views/components/CommentBox.jsx
+++ b/src/views/components/CommentBox.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+const PropTypes = React.PropTypes;
 
 import { models } from 'snoode';
 import querystring from 'querystring';
@@ -10,9 +11,10 @@ class CommentBox extends BaseComponent {
   constructor(props) {
     super(props);
 
-    this.state = {};
+    this.handleInputChange = this.handleInputChange.bind(this);
+    this.submit = this.submit.bind(this);
+    this.handleError = this.handleError.bind(this);
   }
-
   componentDidMount () {
     this.setState({
       reply: savedReply.get(this.props.thingId),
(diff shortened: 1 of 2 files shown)


src/views/components/CommentBox.jsx
```diff
@@ -37,70 +40,93 @@ class CommentBox extends BaseComponent {
...
+    }
+  }
 
-    this.props.app.emit('comment', comment);
+  handleError (e) {
+    let error;
+    if (e.errors.length) {
+      error = e.errors[0][1];
+    }
+    this.setState({ error });
```
schwers: Does this `e.errors.length` check ever fail? Curious if the `setState` call here is to clear the form or if there should be a generic error message

schwers:
:nail_care: other than the small question

src/views/components/CommentBox.jsx
curioussavage: ah, setting state out of the if block was kind of an oversight. all errors from r2 should be nested arrays (that I know of). Ill move that in.

curioussavage:
:haircut:

src/views/components/CommentBox.jsx
```diff
@@ -37,70 +40,92 @@ class CommentBox extends BaseComponent {
...
+    const { apiOptions, app, token, thingId, onSubmit } = this.props;
+    const { requireLogin, config, api } = app;
+
+    if (!token) {
+      requireLogin(config.loginPath);
 
       return;
     }
 
     var comment = new models.Comment({
```
schwers: :no_bell: Not needed, but this might be a good candidate for `const` cleanup while you're here.

src/views/components/CommentBox.jsx
```diff
@@ -37,70 +40,92 @@ class CommentBox extends BaseComponent {
...
-    this.props.app.emit('comment', comment);
+      onSubmit(comment);
+      app.emit('comment', comment);
+    } catch (err) {
+      this.handleError(err);
+    }
+  }
+
+  handleError (e) {
+    let error;
```
schwers: Last thing: could you move this declaration into the block and make it `const`?

schwers:
:nail_care:

src/views/components/CommentBox.jsx
```diff
@@ -10,7 +11,9 @@ class CommentBox extends BaseComponent {
   constructor(props) {
     super(props);
 
-    this.state = {};
+    this.handleInputChange = this.handleInputChange.bind(this);
```
ajacksified: You could use:

```javascript
handleInputChange = (e) => {
 ...
}
```

syntax instead of binding; a lot of new code's been using that.

src/views/components/CommentBox.jsx
```diff
@@ -37,70 +40,92 @@ class CommentBox extends BaseComponent {
...
     var comment = new models.Comment({
-      thingId: this.props.thingId,
+      thingId: thingId,
       text: text
     });
 
-    var options = this.props.app.api.buildOptions(this.props.apiOptions);
+    let options = api.buildOptions(apiOptions);
 
     options = Object.assign(options, {
```
ajacksified: :no_bell: I found a trick for objects:

`const options = {...api.buildOptions(apiOptions), model: comment }`

Splats (`...`) will pull the props out of buildOptions! I'm thinking we may want to start replacing all the Object.assigns everywhere in favor of this much cleaner-looking syntax, but what do you think?

src/views/components/CommentBox.jsx
```diff
@@ -37,70 +40,92 @@ class CommentBox extends BaseComponent {
...
+      app.emit('comment', comment);
+    } catch (err) {
+      this.handleError(err);
+    }
+  }
+
+  handleError (e) {
+    let error;
+    if (e.errors.length) {
+      error = e.errors[0][1];
```
ajacksified: You should be more specific about your logic here, and do something if `e.errors` or `e.errors[0]` is null or not-an-array. (Maybe set a default "there was a problem" message and app.log it?)

ajacksified:
:nail_care:

src/views/components/CommentBox.jsx
curioussavage: I think its fine if your sure it behaves the same way in all cases.

does it work for cloning?  `const newObj = {...{}, oldObject};`

ajacksified: You'd just write `newObj = {...oldObject}`! Woo

src/views/components/CommentBox.jsx
ajacksified: I've been talking to @schwers and this actually produces some pretty ugly code - hold off on this.

curioussavage:
:haircut:

schwers:
:fish:

PR #413, opened 2015-12-09
Register err

curioussavage:
:eyeglasses:  @schwers

### DIFF
--- a/src/server/oauth.es6.js
+++ b/src/server/oauth.es6.js
@@ -387,7 +387,7 @@ var oauthRoutes = function(app) {
       duration: 'permanent',
     };
 
-    const p = new Promise(function(resolve, reject) {
+    return new Promise(function(resolve, reject) {
       let headers = {
         'User-Agent': ctx.headers['user-agent'],
         'Authorization': basicAuth,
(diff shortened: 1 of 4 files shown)


src/server/oauth.es6.js
```diff
@@ -466,7 +485,8 @@ var oauthRoutes = function(app) {
     }
 
     if (!ctx.body.email && ctx.body.newsletter === 'on') {
-      return ctx.redirect('/register?error=EMAIL_NEWSLETTER&message=please+enter+an+email+to+sign+up+for+the+newsletter');
+      const newsLetterMessage = 'please+enter+an+email+to+sign+up+for+the+newsletter';
+      return ctx.redirect(`/register?error=EMAIL_NEWSLETTER&message=${newsLetterMessage}`);
```
ajacksified: The message should go in the view - look it up in a dictionary of errors in the view, based on the `error` querystring.

ajacksified:
:haircut:

curioussavage:
:haircut:

src/server/oauth.es6.js
```diff
@@ -424,62 +426,71 @@ var oauthRoutes = function(app) {
     return p;
   }
 
+  async function handleLogin(username, passwd, ctx) {
```
ajacksified: Can this be rewritten to return a url to redirect to, rather than causing side effects on ctx?

ajacksified:
:nail_care:

curioussavage:
:haircut: This seems to be working perfectly. please run test it though. 

Promises were not being resolved before.

ajacksified:
:fish:

src/server/oauth.es6.js
```diff
@@ -406,22 +406,63 @@ var oauthRoutes = function(app) {
...
         });
     });
+  }
 
-    return p;
+  async function getLoginRedirect(username, passwd, ctx) {
+    try {
+      const status = await login(username, passwd, ctx);
+      const dest = ctx.body.originalUrl || '';
+
```
schwers: Is there a reason we don't try using referrer to bring you back to the right page in cases where originalURL was not set?

schwers: @curioussavage Talked with Jack about this. It's because it's a `POST` and referrer won't be what we need. We could try doing something fancy in routes/redirects in a separate patch though.

schwers:
:fish:

ajacksified:
https://github.com/reddit/reddit-mobile/commit/874adb810ccb68c81e70280e243313a36d9acd3d

PR #414, opened 2015-12-11
add preferences page

curioussavage:

### DIFF
--- a/assets/less/components/prefs.less
+++ b/assets/less/components/prefs.less
@@ -1,0 +1,39 @@
+
+.prefs-select-wrap, .prefs-number-wrap {
+  padding-left: 10px;
+}
+
+.prefs-form-wrap {
+  padding-bottom: 10px;
+}
+
+// toggle component
+
+.prefs-toggle-circle {
+  .shadow;
+  display: inline-block;
+  border-radius: 30px;
+  height: 25px;
+  width: 25px;
+  background-color: white;
+  position: absolute;
+  top: -4px;
+  left: 0;
+}
+
+.prefs-toggle-bar {
+  display: inline-block;
+  border-radius: 30px;
+  height: 15px;
+  width: 45px;
+  background-color: lightgray;
+  position: relative;
+  vertical-align: middle;
+  float: right;
+}
+
+.prefs-toggle-bar.selected > .prefs-toggle-circle {
+  right: 0;
+  left: 42%;
+  background-color: @blue;
+}
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -26,8 +26,9 @@ import RegisterPage from './views/pages/register';
 import MessagesPage from './views/pages/messages';
 import MessageComposePage from './views/pages/messageCompose';
 import SubmitPage from './views/pages/submit';
-import constants from './constants';
+import PrefsPage from './views/pages/prefs';
 
+import constants from './constants';
 import defaultConfig from './config';
 
 const config = defaultConfig();
@@ -726,6 +727,18 @@ function routes(app) {
     this.props.data.set('messages', app.api.messages.get(listingOpts));
 
     this.body = makeBody(MessagesPage);
+  });
+
+  router.get('preferences', '/prefs', function *() {
+    if (!this.token) {
+      let query = {
+        originalUrl: this.url,
+      };
+
+      return this.redirect('/login?' + querystring.stringify(query));
+    }
+
+    this.body = makeBody(PrefsPage);
   });
 
   router.get('404', '*', function *() {
--- a/src/userPreferencesFields.es6.js
+++ b/src/userPreferencesFields.es6.js
@@ -1,0 +1,247 @@
+export default [
+  {
+    'title': 'media',
+    'type': 'section_title',
+  },
+  {
+    'key': 'media',
+    'type': 'select',
+    'options': [
+      {
+        'key': 'on',
+        'description': 'show thumbnails next to links',
+      },
+      {
+        'key': 'off',
+        'description': 'don\'t show thumbnails next to links',
+      },
+      {
+        'key': 'subreddit',
+        'description': 'show thumbnails based on each subreddit\'s media preferences',
+      },
+    ],
+  },
+  {
+    'depends': 'over_18',
+    'key': 'no_profanity',
+    'description': 'make safe(r) for work',
+  },
+
+
+  {
+    'title': 'link options',
+    'type': 'section_title',
+  },
+  {
+    'key': 'newwindow',
+    'description': 'open links in a new window',
+  },
+
+  {
+    'key': 'organic',
+    'description': 'show the spotlight box on the front page',
+  },
+  {
+    'key': 'show_trending',
+    'description': 'show trending subreddits on the front page',
+  },
+  {
+    'key': 'clickgadget',
+    'description': 'show me links I\'ve recently viewed',
+  },
+  {
+    'key': 'compress',
+    'description': 'compress the link display',
+  },
+  {
+    'key': 'domain_details',
+    'description': 'show additional details in the domain text when available',
+  },
+  {
+    'key': 'hide_ups',
+    'description': 'don\'t show me submissions after I\'ve upvoted them',
+  },
+  {
+    'key': 'hide_downs',
+    'description': 'don\'t show me submissions after I\'ve downvoted them',
+  },
+  {
+    'key': 'numsites',
+    'description': 'number of links to display at once',
+    'type': 'select',
+    'options': [10, 25, 50, 100],
+  },
+  {
+    'key': 'min_link_score',
+    'description': 'minimum score to show links',
+    'type': 'range',
+    'min': -100,
+    'max': 100,
+  },
+
+  {
+    'title': 'comment options',
+    'type': 'section_title',
+  },
+  {
+    'key': 'default_comment_sort',
+    'description': 'default comments sort',
+    'type': 'select',
+    'options': ['best', 'old', 'top', 'q&a', 'controversial', 'new'],
+  },
+  {
+    'key': 'ignore_suggested_sort',
+    'description': 'ignore suggested sorts',
+  },
+  {
+    'key': 'highlight_controversial',
+    'description': 'show a dagger (†) on comments voted controversial',
+  },
+  {
+    'key': 'min_comment_score',
+    'description': 'minimum score to show comments',
+    'type': 'range',
+    'min': -100,
+    'max': 100,
+  },
+  {
+    'key': 'num_comments',
+    'description': 'default number of comments displayed',
+    'type': 'range',
+    'min': 0,
+    'max': 200,
+  },
+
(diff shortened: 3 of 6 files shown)


curioussavage:
@librarianavenger
prefs

curioussavage:
will also need to remove many fields that won't affect mobile web at all

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,237 @@
...
+import userOptions, { goldOptions } from '../../userPreferencesFields';
+
+class Toggle extends BaseComponent {
+  constructor(props) {
+    super(props);
+
+    this.state.status = props.status;
+  }
+
+  onClick = () => {
```
schwers: This should be a normal function bound in the constructor

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,237 @@
...
+    const { onChange, formKey } = this.props;
+    const val = !this.state.status;
+    if (onChange) {
+      onChange(null, val, formKey);
+    }
+    this.setState({status: val});
+  }
+
+  render() {
+    let wrapStyle = {
```
schwers: Can you move these styles to css, and then use a modifier class for the enabled/disabled state?

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,237 @@
...
+      if (res._type === 'Preferences') {
+        this.setState({changes: {}});
+        app.redirect('/prefs');
+      }
+    } catch(e) {
+      app.error(e, this, app, { redirect: false, replaceBody: false });
+    }
+  }
+
+  addChange (e, val, key) {
```
schwers: Style nit: no space before the first paren

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,237 @@
...
+    if (e) {
+      val = e.currentTarget.value.trim();
+      key = e.currentTarget.dataset.key;
+    }
+
+    let oldChanges = {...this.state.changes};
```
schwers: :no_bell: What do you and @ajacksified think of syntax `const changes = {...this.stage.changes, [key]: val}`, It compiles into a little bit more work than this, but it might be arguably cleaner and preserves const. Might not matter if we start using immutable though.

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,237 @@
...
+    };
+
+    if (this.state.status) {
+      delete circleStyle.left;
+      circleStyle.right = '0';
+      wrapStyle.backgroundColor = 'lightgreen';
+    }
+
+    return (
+      <div onClick={ this.onClick } style={ wrapStyle }>
```
schwers: Would you want to try making these checkboxes instead? If it turns into too much work we could stick with div's but I imagine checkboxes are better for accessibility. This repo might be a good example http://abpetkov.github.io/switchery/

src/views/pages/prefs.jsx
curioussavage: I think we are using that elsewhere, so looks good to me

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+    const form = this.buildForm();
+    return (
+      <div className='container'>
+        <div className='row'>
+          <div className='col-xs-12 col-sm-12 prefs-form-wrap'>
+              { form }
+              <button
+                className="btn btn-primary pull-right"
+                onClick={ this.updatePrefs } >Save</button>
```
schwers: What do you think of autosaving the prefs as settings are toggled? I personally didn't notice the save button at first

src/views/pages/prefs.jsx
schwers: Okay cool. I didn't see anywhere where the the key is a variable like this, so that was my biggest concern. For reference here's the two in babel and they output so you can see the extra call to a `_defineProperty`

```
const oldChanges = {foo: 'bar', baz: 'fox' };
const key = 'baz';
const val = 'chuck testa';

const newChanges = { ...oldChanges, [key]: val };
const changesHardcoded = {...oldChanges, baz: val };
```

Compiles to:
```
'use strict';

var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };

function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }

var oldChanges = { foo: 'bar', baz: 'fox' };
var key = 'baz';
var val = 'chuck testa';

var newChanges = _extends({}, oldChanges, _defineProperty({}, key, val));

var changesHardcoded = _extends({}, oldChanges, { baz: val });
```

schwers:
:nail_care: Mostly some nits re: toggle switch, looks good! Will be awesome to have this :)

src/views/pages/prefs.jsx
curioussavage: that repo adds hides the input and uses spans which would be easy to do here as well if that sounds fine

src/views/pages/prefs.jsx
curioussavage: I think thats fine as long as we have a clear visual indicator for the behavior

schwers: Good point. This sounds like a good candidate for the new toasts

src/views/pages/prefs.jsx
schwers: Looked into `Switchery` a little more. I think something like this http://codepen.io/designcouch/pen/sDAvk is actually a better starting point. The labels here work as expected and it doesn't rely on js

src/views/pages/prefs.jsx
librarianavenger: Sorry - still not sure why we have mobile-specific preferences at all? How do you even get here?

schwers: These aren't mobile specific, but general account preferences like the ones on desktop. We don't link to prefs right now in production mobile web because of how we log you in (if we link to desktop you won't be logged in as same user as you are on mobile web, or not logged in at all). So this adds support for those prefs instead of having to log into the desktop site and go to the prefs page manually.

assets/less/components/prefs.less
```diff
@@ -0,0 +1,39 @@
+
```
ajacksified: nuke extra newline

assets/less/components/prefs.less
```diff
@@ -0,0 +1,39 @@
...
+  padding-left: 10px;
+}
+
+.prefs-form-wrap {
+  padding-bottom: 10px;
+}
+
+// toggle component
+
+.prefs-toggle-circle {
```
ajacksified: can these be generic `toggle` classes instead of `prefs-toggle`?

src/routes.jsx
```diff
@@ -728,6 +729,18 @@ function routes(app) {
     this.body = makeBody(MessagesPage);
   });
 
+  router.get('preferences', '/prefs', function *() {
+    if (!this.token) {
+      let query = {
```
ajacksified: `const`

src/routes.jsx
```diff
@@ -728,6 +729,18 @@ function routes(app) {
...
+        originalUrl: this.url,
+      };
+
+      return this.redirect('/login?' + querystring.stringify(query));
```
ajacksified: template string

src/userPreferencesFields.es6.js
```diff
@@ -0,0 +1,247 @@
...
+      },
+    ],
+  },
+  {
+    'depends': 'over_18',
+    'key': 'no_profanity',
+    'description': 'make safe(r) for work',
+  },
+
+
```
ajacksified: normalise newlines in this file

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,217 @@
...
+    const { onChange, formKey } = this.props;
+    const val = !this.state.status;
+    if (onChange) {
+      onChange(null, val, formKey);
+    }
+    this.setState({status: val});
+  }
+
+  render() {
+
```
ajacksified: nuke extra newline

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,217 @@
...
+    let toggleClass;
+    if (this.state.status) {
+      toggleClass = 'selected';
+    }
+
+    return (
+      <div className={ `prefs-toggle-bar ${toggleClass}` } onClick={ this.onClick } >
```
ajacksified: nuke space between `} >`

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,217 @@
...
+    super(props);
+
+    this.state.changes = {};
+
+    this.addChange = this.addChange.bind(this);
+    this.updatePrefs = this.updatePrefs.bind(this);
+  }
+
+  async updatePrefs () {
+    var { app, apiOptions } = this.props;
```
ajacksified: `const`

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  makeBoolField(preferences, pref) {
+    let info;
+    if (pref.description) {
+      info = <p>{ pref.description }</p>;
+    }
+
+    return (
+      <div className='' key={ pref.key } style={ {padding: '10px'} }>
```
ajacksified: spacing should look like `{{ padding: '10px' }}`

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,217 @@
...
+
+  buildForm() {
+    const { preferences, user } = this.state.data;
+    // check if user has gold for gold options
+    let options = userOptions.slice();
+    if (user.is_gold) {
+      options.push(goldOptions);
+    }
+
+    const fields = options.map((pref) => {
```
ajacksified: :no_bell: if `buildField` pulled preferences from `this.state.data`, this could be condensed to `options.map(p => this.buildField(p))`

src/views/pages/prefs.jsx
```diff
@@ -0,0 +1,217 @@
...
+    );
+  }
+
+  makeHeading(title) {
+    return (
+      <h3>{ title }</h3>
+    );
+  }
+
+  buildField(pref, preferences) {
```
ajacksified: Your fields need to have `keys` attached to them (preferably, something reproducible like `type-key`

curioussavage:
woops sorry about all the syntax stuff. Wasn't able to go over this with the linter merged yet.

src/views/pages/prefs.jsx
curioussavage: yes thats true. I did it this why to get ordering. Although If design ends up wanting me to do something like the mobile apps then I will probably just abandon this method entirely.

PR #415, opened 2015-12-14
Add WikiPage

curioussavage:
Also used to display privacy policy

### DIFF
--- a/src/redirects.jsx
+++ b/src/redirects.jsx
@@ -39,6 +39,11 @@ function routes(app) {
   app.router.get('/r/:subreddit/search/:query', function*() {
     return this.redirect(`/r/${this.params.subreddit}/search?q=${this.params.query}`);
   });
+
+  app.router.get('/help/*', function*() {
+    const url = this.params[0];
+    return this.redirect(`/wiki/${url}`);
+  });
 }
 
 export default routes;
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -26,8 +26,9 @@ import RegisterPage from './views/pages/register';
 import MessagesPage from './views/pages/messages';
 import MessageComposePage from './views/pages/messageCompose';
 import SubmitPage from './views/pages/submit';
-import constants from './constants';
+import WikiPage from './views/pages/wikiPage';
 
+import constants from './constants';
 import defaultConfig from './config';
 
 const config = defaultConfig();
@@ -621,14 +622,8 @@ function routes(app) {
     });
   }
 
-  router.get('/goto', function * () {
-    let location = this.query.location.toLowerCase();
-    let token = this.token;
+  function makeOptions(token, app) {
     let apiOptions;
-    let result;
-
-    app.emit('goto', location);
-
     if (token) {
       apiOptions =  {
         origin: app.getConfig('authAPIOrigin'),
@@ -642,7 +637,17 @@ function routes(app) {
       };
     }
 
-    let options = app.api.buildOptions(apiOptions);
+    return app.api.buildOptions(apiOptions);
+  }
+
+  router.get('/goto', function * () {
+    let location = this.query.location.toLowerCase();
+    const token = this.token;
+    let result;
+
+    app.emit('goto', location);
+
+    let options = makeOptions(token, app);
 
     if (this.headers['user-agent']) {
       options.headers['user-agent'] = this.headers['user-agent'];
@@ -727,6 +732,34 @@ function routes(app) {
 
     this.body = makeBody(MessagesPage);
   });
+
+  function * wikiPage() {
+    const path = this.params[0].substr(1) || 'index';
+
+    let options = buildAPIOptions(this, {
+      query: {
+        raw_json: 1,
+      },
+      subreddit: this.params.subreddit || null,
+      path,
+    });
+
+    const wikiGet = new Promise(function(resolve, reject) {
+      app.api.wiki.get(options).then(resolve, function(e) {
+        // handle api crap. otherwise user gets redirected to auth if loggedout and wiki is disabled.
+        if (e.status === 403) {
+          e.status = 404;
+        }
+        reject(e);
+      });
+    });
+
+    this.props.data.set('wikiPage', wikiGet);
+    this.body = makeBody(WikiPage);
+  }
+
+  router.get('wiki', '/wiki*', wikiPage);
+  router.get('wiki.subreddit', '/r/:subreddit/wiki*', wikiPage);
 
   router.get('404', '*', function *() {
     this.props.status = 404;
--- a/src/views/components/SideNav.jsx
+++ b/src/views/components/SideNav.jsx
@@ -203,9 +203,14 @@ class SideNav extends BaseComponent {
         aboutButtons = (
           <ul key='about' className='SideNav-ul list-unstyled'>
             {aboutItems.map((i) => {
+              let url = `${config.reddit}${i.url}`
+              if (i.url.indexOf('/help') !== -1 || i.url.indexOf('/wiki') !== -1) {
+                url = i.url;
+              }
+
               return (
                 <li className='SideNav-li'>
-                  <a className='SideNav-button' href={`${config.reddit}${i.url}`}>
+                  <a className='SideNav-button' href={ url } >
                     { snooIcon }
                     <span className='SideNav-text'>{titleCase(i.title)}</span>
                   </a>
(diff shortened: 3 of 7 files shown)


curioussavage:
wikipage

src/views/components/SideNav.jsx
```diff
@@ -202,9 +202,14 @@ class SideNav extends BaseComponent {
         aboutButtons = (
           <ul key='about' className='SideNav-ul list-unstyled'>
             {aboutItems.map((i) => {
+              let url = `${config.reddit}${i.url}`
```
schwers: :no_bell: We're going to have a mild conflict with https://github.com/reddit/reddit-mobile/pull/420/files. 

What would you and @ajacksified think of not link to config.reddit/* since we'll have more of those pages covered, and redirecting to the ones we don't in redirects.jsx ?

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    const subreddit = this.props.subredditName;
+    const compact = this.props.compact;
+    let content;
+
+    switch (data.wikiPage._type) {
+      case 't3':
+        return (
+          <div className={'container Listing-container' + (compact ? ' compact' : '')} ref='listings'>
```
schwers: You can use a `ListingContainer` and it will track the compact state for you

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+        content = (
+          <div className='wikiPage-html' dangerouslySetInnerHTML={{ __html: process(mobilify(body)) }} />
+        );
+        break;
+      case 'WikiPageSettings':
+        const { pageEditorsList, editingPermissionLevel, listedInPagesIndex } = data.wikiPage;
+        let editors = pageEditorsList.map((editor) => {
+          return (
+            <li>{ editor.name }</li>
```
schwers: Do we have a consensus on when to use `key=...` in these array situations? I'm not sure if this updating a lot so it probably isn't a concern but React will give you warnings. Maybe we can just add a rule to the linter to enforce key's on components in an map/array

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+      );
+    }
+
+    const subreddit = this.props.subredditName;
+    const compact = this.props.compact;
+    let content;
+
+    switch (data.wikiPage._type) {
+      case 't3':
+        return (
```
schwers: Can you move the t3 case out of the switch and before the `let content`? That would make it more clear that this case doesn't rely on setting content and wrapping it

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+      return (
+        <Loading />
+      );
+    }
+
+    const subreddit = this.props.subredditName;
+    const compact = this.props.compact;
+    let content;
+
+    switch (data.wikiPage._type) {
```
schwers: Can you make a new function and call it something like `renderContent` that returns  the content assigned in this switch statement? This function's rather large and I think it would help with readability and seeing where/how the content is wrapped

src/routes.jsx
```diff
@@ -629,7 +624,17 @@ function routes(app) {
       };
     }
 
-    let options = app.api.buildOptions(apiOptions);
+    return app.api.buildOptions(apiOptions);
+  }
+
+  router.get('/goto', function * () {
+    let location = this.query.location.toLowerCase();
+    let token = this.token;
```
schwers: `token` (probably) should be a `const`

src/redirects.jsx
```diff
@@ -39,6 +39,11 @@ function routes(app) {
   app.router.get('/r/:subreddit/search/:query', function*(next) {
     return this.redirect(`/r/${this.params.subreddit}/search?q=${this.params.query}`);
   });
+
+  app.router.get('/help/*', function*(next) {
```
schwers: :no_bell: Why do we link to the /help/ pages instead of the wiki directly?

schwers:
:nail_care: 

Can you try rebasing this as well? I think some of the let/var's have changed after the linter patch

src/redirects.jsx
curioussavage: only to match .com urls where it always links to /help/* which only pulls data from the wiki

src/views/components/SideNav.jsx
curioussavage: thats a good point. I think that sounds better now that the majority will be relative links

curioussavage:
:haircut:

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,149 @@
...
+    const pagesActive = path.indexOf('/wiki/pages') !== -1 ? 'active' : '';
+
+    return (
+      <div className='container wikiPage-wrapper'>
+        <TextSubNav>
+          <li className='TextSubNav-li' active={ revisionActive } >
+            <a
+              className={ `TextSubNav-a ${revisionActive}` }
+              href={ `${urlPrefix}/wiki/revisions` }
+            >Revisions</a>
```
schwers: :no_bell: What would you think of adding a rule to linting which forces a newline and indent in situations like this? (i.e. body content for a multiline tag)

schwers:
It doesn't look like revision pages are working. AFAIK I know this is because of the Snoode patch.

schwers:
:haircut:

schwers:
:nail_care:

schwers:
@librarianavenger How should I be linking to wiki pages in the new topnav design? The current link is in the dropdown in the topbar, but that's replaced in new topnav.

src/redirects.jsx
```diff
@@ -39,6 +39,11 @@ function routes(app) {
   app.router.get('/r/:subreddit/search/:query', function*() {
     return this.redirect(`/r/${this.params.subreddit}/search?q=${this.params.query}`);
   });
+
+  app.router.get('/help/*', function*() {
+    let url = this.params[0];
```
ajacksified: `const`

src/routes.jsx
```diff
@@ -728,6 +733,34 @@ function routes(app) {
     this.body = makeBody(MessagesPage);
   });
 
+  function * wikiPage() {
+    const path = this.params[0].substr(1) || 'index';
+
+    let options = buildAPIOptions(this, {
+      query: {
+        raw_json: 1,
```
ajacksified: :no_bell: We should eventually move this into the API and make _all_ queries `raw_json`.

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,149 @@
...
+import BasePage from './BasePage';
+import Loading from '../components/Loading';
+import ListingContainer from '../components/ListingContainer';
+import TextSubNav from '../components/TextSubNav';
+
+import moment from 'moment';
+
+class WikiPageComp extends BasePage {
+
+  makeContent(props, data) {
```
ajacksified: nuke extra newline above

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,149 @@
...
+    const compact = props.compact;
+    let content;
+
+    switch (data.wikiPage._type) {
+      case 't3':
```
ajacksified: turn these into constants; `PAGETYPES.LISTING, PAGETYPES.WIKI_REVISION`, etc.

src/views/pages/wikiPage.jsx
```diff
@@ -0,0 +1,149 @@
...
+        </TextSubNav>
+         { content }
+      </div>
+    );
+  }
+}
+
+export default WikiPageComp;
+
+
```
ajacksified: nuke newlines

ajacksified:
:nail_care:

curioussavage:
I think the latest snoode change was not pushed, revisions should work now.

ajacksified:
^ disregard failure, was using a broken travis config

ajacksified:
:fish:

curioussavage:
paging @spez since this is UI related. Pretty simple page, but whats there is using the old style until design can touch it.

spez:
what's the change?

ajacksified:
@spez this adds wiki pages for subreddits, and also pulls the privacy policy and terms of service from the desktop site so they can be displayed in mobile web. The screenshot in the PR is an example of what it looks like when we pull the markdown in. It is styled using the current mweb styles.

ajacksified:
@curioussavage if you rebase against latest master, you'll get a green build, btw.

spez:
got it :fish:

PR #419, opened 2015-12-16
Fix indexOf undefined errors when clicking links without an href.

schwers:
Noticed these errors popping up, looks like one of the causes is about pages not setting topNavLink.

Also prevents the error from happening in `client.es6.js`. Alternatively we could catch it and log it to see where else it happens. Thoughts? :eyeglasses: @curioussavage or @ajacksified

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -328,13 +328,31 @@ function initialize(bindLinks) {
     }
   }
 
+  function logMissingHref($link) {
+    const $linkClone = $link.cloneNode(true);
+    const $tmpWrapper = document.createElement('div');
+    $tmpWrapper.appendChild($linkClone);
+    const linkStringified = $tmpWrapper.innerHTML;
+
+    const error = {
+      message: 'A tag missing HREF',
+      linkStringified,
+    };
+
+    const options = {
+      redirect: false,
+      replaceBody: false,
+    };
+
+    app.error(error, app.getState('ctx'), app, options);
+  }
+
   function attachEvents() {
     attachFastClick(document.body);
 
-    if(history && bindLinks) {
-
+    if (history && bindLinks) {
       $body.addEventListener('click', function(e) {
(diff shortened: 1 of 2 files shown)


curioussavage:
:fish: this fix looks fine. No opinion about logging it instead here.

ajacksified:
We should definitely be logging - we shouldn't have links without hrefs.

assets/js/client.es6.js
```diff
@@ -340,7 +340,7 @@ function initialize(bindLinks) {
           }
         }
 
-        var href = $link.getAttribute('href');
+        var href = $link.getAttribute('href') || "";
```
ajacksified: single quotes

ajacksified:
One nit, :nail_care:

assets/js/client.es6.js
umbrae: :no_bell: Is this something that can be added to the linter I wonder?

ajacksified: @umbrae - it's in the linter; this PR was opened before the linter.

umbrae: oh yay!

schwers:
:haircut:

ajacksified:
Just a baby nit, :nail_care:

assets/js/client.es6.js
```diff
@@ -342,8 +360,13 @@ function initialize(bindLinks) {
           }
         }
 
-        var href = $link.getAttribute('href');
-        var currentUrl = app.fullPathName();
+        const href = $link.getAttribute('href') || '';
```
ajacksified: Does this need to be a string check? You could simplify this by just calling href = $link.getAttribute('href') and checking if (href).

schwers: Right, I'll update

schwers:
:haircut:

ajacksified:
:fish:

PR #423, opened 2015-12-18
Fix dnt.es6.js from breaking JS in Safari

schwers:
Safari doesn’t support assigning to global.navigator.doNotTrack, this
is currently breaking JS in Safari, preventing things like dropdown,
topnav, etc.

(Also fixes linter warnings).

:eyeglasses: @ajacksified @dwick


dwick:
can we just make this window.DO_NOT_TRACK so safari is still supported? :nail_care:

schwers:
That seems fine to me, why was it changed in https://github.com/schwers/reddit-mobile/commit/f76fe02949743e3b253363ced330e2af7f1451f5 ?

dwick:
bad rebase :\

PR #425, opened 2015-12-18
Send boolean instead of T/F in event tracking

ajacksified:
:eyeglasses: @curioussavage or @schwers, and @drunken-economist 

Send the boolean instead of 'T' / 'F'.


schwers:
:fish:

drunken-economist:
:fish: thanks for the quick turnaround

PR #429, opened 2016-01-05
Add styling to indicate a post is locked

nramadas:
Modified the look of locked posts in a few ways:

**Posts**
* [x] Add a lock icon to the left of the flair or subreddit name, if they exist
* [x] Add a lock icon to the left of the post listing footer, if flair and/or subreddit name *doesn't* exist.

**Comments**
* [x] Remove the ability to comment on a post

:eyeglasses: @ajacksified 

The lock sits next to the flair if it exists:
screen shot 2016-01-11 at 12 16 08 pm

If there isn't any flair, it sits at the bottom:
screen shot 2016-01-11 at 12 17 04 pm

When comments are locked:
screen shot 2016-01-11 at 10 51 23 am

When replies are locked:
screen shot 2016-01-11 at 5 45 06 pm


src/views/components/Comment.jsx
```diff
@@ -17,6 +17,21 @@ var replyIcon = (
 );
 
 class Comment extends BaseComponent {
+  static propTypes = {
```
ajacksified: :no_bell: it's not in the styleguide, but we've generally been putting static properties at the bottom of the class definition. Do you have any objections moving these to the bottom?

src/views/components/Comment.jsx
```diff
@@ -243,23 +258,27 @@ class Comment extends BaseComponent {
...
-            <Vote
-              app={ app }
-              setScore={ this.setScore }
-              thing={ comment }
-              token={ token }
-              apiOptions={ apiOptions }
-            />
-          </li>
-          <li>
+          { !locked
```
ajacksified: We've been avoiding ternaries in favor of assigning react elements to variables, but it _is_ nice to have it inline like this. Thoughts, @schwers / @curioussavage

src/views/components/Vote.jsx
```diff
@@ -110,78 +120,96 @@ class Vote extends BaseComponent {
...
+      <li>
+        <form
+          className='vote-form'
+          action={ '/vote/'+ thing.name }
+          method='post'
+        >
+          <input type='hidden' name='direction' value='1'/>
+          <button
+            type='submit'
+            className={ 'vote text-muted ' + (this.getVoteClass() || '') }
```
ajacksified: Prefer template strings (here and in `action` above). Also, since getVoteclass returns empty string, you don't need the `||`

src/views/components/Vote.jsx
```diff
@@ -110,78 +120,96 @@ class Vote extends BaseComponent {
...
-  thing: React.PropTypes.oneOfType([
-    propTypes.comment,
-    propTypes.listing,
-  ]).isRequired,
-};
+  render() {
+    return (
+      <ul className='linkbar linkbar-compact'>
+        { !this.props.locked 
+          ? this.renderUpvote()
```
ajacksified: :no_bell: what do you think about moving the `locked` check to `renderUpvote`, to simplify the `render` function?

src/views/pages/listing.jsx
```diff
@@ -254,15 +263,17 @@ class ListingPage extends BasePage {
...
-            apiOptions={ apiOptions }
-            thingId={ listing.name }
-            user={ user }
-            token={ token }
-            app={ app }
-            ctx={ ctx }
-            onSubmit={ this.onNewComment }
-          />
+          { listing.locked
+            ? <div className="listing-content__locked">Comments are locked</div>
```
ajacksified: single quotes

ajacksified:
:nail_care:

src/views/components/Comment.jsx
curioussavage: I have no strong preference. I used it like this before. Maybe in limited use its fine?

src/views/components/Comment.jsx
nramadas: no strong objection. i would argue there is something nice about reading the prop types first before reading code that uses said props. it gives a clearer picture of what's to come. i don't feel too strongly about it, so if you guys are more comfortable with static properties at the bottom, i'll move them.

src/views/components/Vote.jsx
nramadas: so, this is getting incredibly nitpicky :), but I think it's confusing for a `render*` method to NOT return html. If the `!locked` check were deciding between two templates, I would agree w/ moving into `renderUpvote`. but since there is potential for for no html to be rendered (and getting a `null` instead), i'd rather keep the check out of the `renderUpvote` method.

nramadas:
:haircut:

src/views/components/Comment.jsx
ajacksified: I'd say that for consistency's sake they should be on the bottom, and we can talk about moving everything up as part of a styleguide addition.

src/views/components/Listing.jsx
```diff
@@ -140,6 +140,9 @@ class Listing extends BaseComponent {
     if (subredditNode || flairNode) {
       var row1Node = (
         <div className='Listing-header-row1'>
+          { listing.locked
+            ? <div className="Listing-lock icon-lock"></div>
```
ajacksified: single quotes

src/views/components/Listing.jsx
```diff
@@ -210,9 +213,18 @@ class Listing extends BaseComponent {
...
+    const hasFlair =
+      ListingContent.isNSFW(listing) ||
+      linkFlairText ||
+      (!hideSubredditLabel && subreddit);
+
     return (
       <footer className='Listing-footer'>
         <ul className='linkbar text-muted'>
+          { !hasFlair && listing.locked
+            ? <div className="Listing-lock-large listing-footer-icon icon-lock"></div>
```
ajacksified: single quotes

src/views/components/Vote.jsx
```diff
@@ -6,6 +6,16 @@ import propTypes from '../../propTypes';
 import BaseComponent from '../components/BaseComponent';
 
 class Vote extends BaseComponent {
+  static propTypes = {
+    // apiOptions: React.PropTypes.object,
```
ajacksified: nuke or uncomment

src/views/components/Vote.jsx
```diff
@@ -110,78 +120,96 @@ class Vote extends BaseComponent {
...
+    return voteClass;
+  }
+
+  renderUpvote() {
+    const { thing } = this.props;
+    return (
+      <li>
+        <form
+          className='vote-form'
+          action={ '/vote/'+ thing.name }
```
ajacksified: template string

src/views/components/Vote.jsx
```diff
@@ -110,78 +120,96 @@ class Vote extends BaseComponent {
...
+    );
+  }
 
+  renderDownvote() {
+    const { thing } = this.props;
+    return (
+      <li>
+        <form
+          className='vote-form'
+          action={ '/vote/'+ thing.name }
```
ajacksified: template string

ajacksified:
:nail_care:

src/views/pages/listing.jsx
```diff
@@ -11,6 +11,15 @@ import Loading from '../components/Loading';
 import TopSubnav from '../components/TopSubnav';
 
 class ListingPage extends BasePage {
+  static propTypes = {
```
schwers: :no_bell: I like this being at the top of the class, but everywhere else seems to keep it at the bottom. Maybe we can make that change everywhere and add a linter rule possibly (thoughts @ajacksified @curioussavage )

schwers: Totally missed your convo about this.

src/views/components/Vote.jsx
```diff
@@ -110,78 +120,96 @@ class Vote extends BaseComponent {
     return true;
   }
 
-  render() {
-    var voteClass = '';
+  getVoteClass() {
```
schwers: This could be an actual getter eg `get voteClass()`

tdohz:
FYI I don't think the design is fully locked for this, and some of the requirements listed on the ticket were incorrect. Sorry about that, I wasn't keeping a close eye on it, but in any case there will need to be a few changes - at a minimum, voting should **not** be disabled on locked posts.

src/views/components/Comment.jsx
schwers: I'd prefer avoiding multiline ternaries. I think when one of the cases is null it does make things nice and maybe that's an exception?. Pulling them out into assignments or render functions, to me, makes the overall template more readable and clean. 

One thing that seems like a code smell to me in this case is both of the ternaries checking !locked. I think'd I'd prefer one check for locked and then both li's, but on the other hand you have to then either have to have two assignments or an array (both seem clunky).

Thoughts?

nramadas:
:haircut:

tdohz:
:+1: This looks good to me, especially as a temp solution while we build out the toasts. I don't think Erica has access to this repo so I'll just do a quick check with her in slack.

src/views/pages/listing.jsx
```diff
@@ -254,30 +254,32 @@ class ListingPage extends BasePage {
...
-            user={ user }
-            token={ token }
-            app={ app }
-            ctx={ ctx }
-            onSubmit={ this.onNewComment }
-          />
+          { listing.locked
+            ? <div className='listing-content__locked'>Comments are locked</div>
+            : <CommentBox
+                apiOptions={ apiOptions }
```
schwers: :no_bell: Linter gives a warning on the spaces here because it doesn't know how to deal with the ternary increasing the indentation. Apparently
 ``` 
{ listing.locked
  ? // (can be here or on the previous line)
  <div className='listing-content__locked'>Comments are locked</div>
  :
  <CommentBox
    apiOptions={ apiOptions }
    thingId={ listing.name }
    user={ user }
    token={ token }
    app={ app }
    ctx={ ctx }
    onSubmit={ this.onNewComment }
  /> }
``` is valid and linter warning free

File: src/views/components/Listing.jsx [hunk 14]
```diff
@@ -140,6 +140,9 @@ class Listing extends BaseComponent {
     if (subredditNode || flairNode) {
       var row1Node = (
         <div className='Listing-header-row1'>
+          { listing.locked
+            ? <div className='Listing-lock icon-lock'></div>
```
schwers: small nit: can you make this self-closing?

src/views/components/Listing.jsx
```diff
@@ -210,9 +213,18 @@ class Listing extends BaseComponent {
...
+    const hasFlair =
+      ListingContent.isNSFW(listing) ||
+      linkFlairText ||
+      (!hideSubredditLabel && subreddit);
+
     return (
       <footer className='Listing-footer'>
         <ul className='linkbar text-muted'>
+          { !hasFlair && listing.locked
+            ? <div className='Listing-lock-large listing-footer-icon icon-lock'></div>
```
schwers: same small nit: self-closing preferred

src/views/components/Vote.jsx
```diff
@@ -110,78 +110,100 @@ class Vote extends BaseComponent {
...
+          <input type='hidden' name='direction' value='1'/>
+          <button
+            type='submit'
+            className={ `vote text-muted ${this.voteClass}` }
+            data-vote='up'
+            data-thingid={ thing.name }
+            data-no-route='true'
+            onClick={ this.upvote }
+          >
+            <span className='icon-upvote-circled'></span>
```
schwers: self-closing nti

src/views/components/Vote.jsx
```diff
@@ -110,78 +110,100 @@ class Vote extends BaseComponent {
...
+          <input type='hidden' name='direction' value='-1'/>
+          <button
+            type='submit'
+            className={ `vote text-muted ${this.voteClass}` }
+            data-vote='down'
+            data-thingid={ thing.name }
+            data-no-route='true'
+            onClick={ this.downvote }
+          >
+            <span className='icon-downvote-circled'></span>
```
schwers: self-closing nit

schwers:
:nail_care: only style / linter warning nitpicks. otherwise looks good and works from my testing

nramadas:
@schwers :haircut:

package.json
```diff
@@ -9,7 +9,8 @@
     "watch": "gulp watch",
     "dev": "gulp dev",
     "test": "mocha test/index.js",
-    "lint": "eslint src --ext .js,.jsx"
+    "lint": "eslint src --ext .js,.jsx",
+    "fonts": "gulp icon-fonts"
```
schwers: :no_bell: _maybe_ should be in its own patch but just please update the building page in the wiki when this goes out

schwers:
:fish:

package.json
nramadas: whoops, i meant to take that out actually

Issue #430, opened 2016-01-06
Links from google consistently redirect to home page when on mobile app


Plasma:
When searching for reddit threads via google, you get to see links like this in search results:

https://m.reddit.com/r/brownman/comments/34fhk6/is_anybody_working_on_a_compilation_of_best_of/

(This is just a random thread I found searching for jibberish)

When you click on it via Chrome in iPhone, you just get redirected to https://m.reddit.com which is obviously super frustrating.

If I click on that link now in a web browser on Chrome deskop it works, but if I reload the page it pushes me back to m.reddit.com too!

If I drop the m subdomain then the link works always: https://reddit.com/r/brownman/comments/34fhk6/is_anybody_working_on_a_compilation_of_best_of/

(Referred from https://github.com/reddit/reddit/issues/1494 )

eridal:
This is an internal reddit problem, it's not related to the fact the user came from the an external website.

This is makes browsing reddit really annoying since you cannot navigate to any internal section. All links will trigger the home page. The only one that seems to work is the "next page"

phil303:
This should be fixed now. Feel free to re-open if that's not the case.

PR #431, opened 2016-01-06
Ignore empty token cookies

ajacksified:
Instead of trying to refresh a token when a tokenExpires exists but a
token doesn't, nuke the invalid cookies.

:eyeglasses: @schwers @umbrae

### DIFF
--- a/src/server/index.jsx
+++ b/src/server/index.jsx
@@ -420,16 +420,17 @@ class Server {
         return;
       }
 
-      var now = new Date();
-      var expires = this.cookies.get('tokenExpires');
+      const now = new Date();
 
-      expires = new Date(expires);
+      const cookieToken = this.cookies.get('token');
+      const cookieExpires = this.cookies.get('tokenExpires');
+      const rToken = this.cookies.get('refreshToken');
 
-      if (now > expires) {
-        var rToken = this.cookies.get('refreshToken');
+      const expires = new Date(cookieExpires);
 
+      if (cookieToken && rToken && now > expires) {
         try {
-          var token = yield app.refreshToken(this, rToken);
+          const token = yield app.refreshToken(this, rToken);
 
           this.token = token.token.access_token;
           this.tokenExpires = token.token.expires_at.toString();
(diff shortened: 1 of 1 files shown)


ajacksified:
Fixes #430

src/server/index.jsx
```diff
@@ -443,6 +444,11 @@ class Server {
 
           return;
         }
+      // Sometimes, there's an empty token cookie.
```
umbrae: I wonder if this should be a bit more clear of the particular issue we're seeing, or the edge case we're trying to handle? This comment might make it sound like this is a desired/expected result, but it's not, so we might want to say something explicit about this being error handling for malformed cookies.

umbrae:
:nail_care: code looks good, just one comment thought

src/server/index.jsx
```diff
@@ -420,16 +420,17 @@ class Server {
...
+      const rToken = this.cookies.get('refreshToken');
 
-      if (now > expires) {
-        var rToken = this.cookies.get('refreshToken');
+      const expires = new Date(cookieExpires);
 
+      if (cookieToken && rToken && now > expires) {
         try {
-          var token = yield app.refreshToken(this, rToken);
+          let token = yield app.refreshToken(this, rToken);
```
schwers: This should probably be `const`

schwers:
:nail_care: small nit, but looks good

ajacksified:
:haircut:

spladug:
General question: why two cookies instead of one if their content is conjoined?

umbrae:
:fish: on this fix - we probably could join these cookies, but that would be a larger thing. (I think we got bit by this same thing with `loid` and `loidcreated` :frowning:)

ajacksified:
Leftover implementation detail from before we had refresh tokens
implemented. Another PR should set a matching expiry on the token instead
of keeping a separate tokenExpires cookie, and remove calls to tokenExpires.

ajacksified:
Err, thinking about it again, I think it was also a cheap way of getting
expiry to know when to refresh, instead of reading the expires from the
cookie. No reason it all couldn't be in one base64'd json blob or something.

PR #433, opened 2016-01-08
Only nuke broken cookies

ajacksified:
:eyeglasses: @schwers @umbrae

Don't nuke cookies just because expires exists - nuke them because expires exists while one of the required cookies doesn't.


umbrae:
:fish: d'oh, I misread that if also.

schwers:
:fish:

PR #434, opened 2016-01-08
Don't try to load fresh links when you restore a tab, and alert user about stale next/previous

ajacksified:
Instead, use the old ones, and implement a more user-friendly "no links" error message, helping the user load fresh links.

Obviously needs design help, but is this what you were thinking about, @umbrae?

:eyeglasses: @umbrae and @curioussavage or @schwers

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -116,8 +116,12 @@ function modifyContext (ctx) {
     redirect: redirect.bind(app),
     env: 'CLIENT',
     winWidth: window.innerWidth,
+    // pick up notifications off of the base for the intial load, since the
+    // server sends a delete the first time. delete the basectx.notifications
+    // after the first render.
+    notifications: baseCtx.notifications ||
+                  (decodeURIComponent(cookies.get('notifications') || '')).split(','),
   });
-
 
   if (!ctx.token) {
     ctx.loid = cookies.get('loid');
@@ -309,11 +313,11 @@ function initialize(bindLinks) {
     render(app, app.fullPathName(), false, modifyContext).then(function(props) {
       setTitle(props);
     });
-  }
+  };
 
   app.forceRender = function (view, props) {
     ReactDOM.render(view(props), app.config.mountPoint);
-  }
+  };
 
   var scrollCache = {};
 
@@ -321,18 +325,20 @@ function initialize(bindLinks) {
 
   function postRender(href) {
     return function(props) {
-      if(scrollCache[href]) {
+      if (scrollCache[href]) {
         $body.scrollTop = scrollCache[href];
       } else {
         $body.scrollTop = 0;
       }
 
-      setTitle(props);
+      if (props) {
+        setTitle(props);
 
-      if (!props.data.get('subreddit')) {
-        setMetaColor(constants.DEFAULT_KEY_COLOR);
+        if (!props.data.get('subreddit')) {
+          setMetaColor(constants.DEFAULT_KEY_COLOR);
+        }
       }
-    }
+    };
   }
 
   function logMissingHref($link) {
@@ -432,6 +438,9 @@ function initialize(bindLinks) {
   render(app, app.fullPathName(), true, modifyContext).then(function() {
     app.setState('dataCache');
 
+    // nuke bootstrap notifications
+    app.setState('ctx', { ...bootstrap.ctx, notifications: undefined });
+
     attachEvents();
     referrer = document.location.href;
     sendTimings();
@@ -466,6 +475,15 @@ function initialize(bindLinks) {
 
   app.on(constants.TOGGLE_OVER_18, function(val) {
     cookies.set('over18', val);
+  });
+
+  app.on('notification', function(notification) {
+    this.setNotification(cookies, notification);
+  }.bind(app));
+
+  app.on('pageview', function() {
+    // reset notifications once the page loads
+    cookies.set('notifications');
   });
 
   app.on(constants.HIDE_GLOBAL_MESSAGE, function(message) {
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -429,11 +429,6 @@ function initialize(bindLinks) {
   // config value after render.
   beginRender = Date.now();
 
-  // If we're using an old render cache from a restore, nuke it
-  if ((beginRender - window.bootstrap.render) > 1000 * 60 * 5) {
-    app.setState('dataCache');
-  }
-
   render(app, app.fullPathName(), true, modifyContext).then(function() {
     app.setState('dataCache');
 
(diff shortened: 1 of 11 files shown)


ajacksified:
screen shot 2016-01-08 at 11 07 05 am

umbrae:
UX wise, I think I was expecting that this would be an inline indicator. I don't think there's a strong reason to take over the whole page unless it makes it vastly easier from a dev point of view.

I'm +1 for doing this so we don't nuke the state after 5 minutes, but I also think we can come up with a potentially better solution (like iterating back through the current list and trying next= the previous one or something)

ajacksified:
@umbrae it is vastly easier - this would also catch the case where the API sends nothing back because, say, the listing at `after_id` got deleted or something, which an inline time-based message wouldn't catch. 

Agreed that it's temporary until we find a really good UX.

ajacksified:
Alternate text option: `o no, ur dankness level is too low. click here to load the dankest memes`

schwers:
I think we should seriously consider that text change

umbrae:
@tdohz thoughts on text here? Not the joke one but the real one. :P Ignoring
the design.

tdohz:
cc @librarianavenger

I'm just generally not a fan of whole-page errors when there's only one logical course. Can we just instead take you back to the refreshed previous page (or, I guess, first page, if that's the best we can do) with an error (ideally a nice new [toast](https://reddit.atlassian.net/browse/MOBILEWEB-301)) saying something like "Sorry, we ran into a problem and had to refresh the page"?

Alternately, can we maybe just cache the current pagination cursor so that we can do the refresh-->fetch next cursor loop silently in the background? It still could fail if multiple items were removed, I guess, but with a lower likelihood.

src/views/components/ListingPaginationButtons.jsx
```diff
@@ -79,13 +85,15 @@ class ListingPaginationButtons extends BaseComponent {
...
     }
 
-    if (nextUrl) {
-      nextButton = (
-        <a href={ nextUrl } rel='next' className='btn btn-sm btn-primary IndexPage-button next'>
-          Next Page
-          <span className='glyphicon glyphicon-chevron-right'></span>
-        </a>
-      );
+    if (listings.length === pageSize) {
```
schwers: What is this guard intended to prevent? It breaks two scenarios right now:

1) Subreddit listings seems to be returning 27 listings per page instead of 25, so subreddit pagination is broken. Not sure if this is an api bug or a snoode bug.

2) Search pages often return 3 results less than the page size to 'allow room' for subreddit results not just listing results. I worked around this in `search.jsx` but because the pagination buttons check this the next buttons are missing on search pages too.

ajacksified: This should prevent showing a "next" button if you only got 15 out of 25 listings (incomplete page sizes.) I tested it on listings- where I was seeing the right number. I can change this to a `<` instead, and we can set pageSize on search to 22.

schwers:
:nail_care: I see the loading flash sometimes but I can't readily repro when it happens. Seems to be a lot snappier with this overall tho :D

src/views/components/ListingPaginationButtons.jsx
ajacksified: Update: I suspect you get 27 because of stickies.

ajacksified:
@schwers it'll still happen when the server fails an API request, but should _only_ happen in that case.

ajacksified:
:haircut:

tdohz:
Maybe we can do a quick sync-up to understand what is and isn't possible technically. I think both Erica and I would like to avoid displaying this page if at all possible.

ajacksified:
screen shot 2016-01-20 at 9 35 24 am

Depends on https://github.com/reddit/snoode/pull/72 and https://github.com/reddit/horse-react/pull/11.

:eyeglasses: @librarianavenger and @tdohz. (This uses the current alert style and will be updated to the new toast/alert system soon.)

:haircut:

ajacksified:
also still :eyeglasses: @curioussavage or @schwers or @nramadas

librarianavenger:
I like this solution. :)

assets/js/client.es6.js
```diff
@@ -429,11 +429,6 @@ function initialize(bindLinks) {
   // config value after render.
   beginRender = Date.now();
 
-  // If we're using an old render cache from a restore, nuke it
-  if ((beginRender - window.bootstrap.render) > 1000 * 60 * 5) {
```
schwers: :no_bell: I'm optimistic about this :)

tdohz:
:+1: looks good. To clarify this only happens when you hit previous / next and you run into an error?

ajacksified:
That is correct.

src/views/components/ListingPaginationButtons.jsx
```diff
@@ -85,13 +92,15 @@ class ListingPaginationButtons extends BaseComponent {
...
     }
 
-    if (nextUrl) {
-      nextButton = (
-        <a href={ nextUrl } rel='next' className='btn btn-sm btn-primary IndexPage-button next'>
-          Next Page
-          <span className='glyphicon glyphicon-chevron-right'></span>
-        </a>
-      );
+    if (listings.length >= pageSize) {
```
schwers: Can this be `listings.length >= pageSize && nextUrl`?

ajacksified:
:haircut: :cookie: 4 u @schwers

src/views/pages/index.jsx
```diff
@@ -79,7 +124,7 @@ class IndexPage extends BasePage {
 
     var user = this.state.data.user;
 
-    let pagingPrefix = '';
+    let pagingPrefix = '/';
```
schwers: maybe :no_bell: why is this defined to be `/` when that's the default value for `noListingsPage`

src/routes.jsx
```diff
@@ -73,6 +73,7 @@ function buildAPIOptions(ctx, options={}) {
 
 // Filter the props we want to expose to the react elements.
 function filterContextProps(ctx) {
+  console.log(ctx.notifications);
```
schwers: This should probably be removed before ship

schwers:
Something is causing React's transaction update to throw an error. This is preventing the cookie from being cleared causing the next to pages to see the notification, versus the next page load and clear as intended

To repro and debug this:
* open a new incognito tab
* enable preserve log to to save the output from the `console.log` in `filterContextProps`
* navigate to r/HighQualityGifs via the topnav
* using Chrome inspector, edit the pagination buttons after param to `t3_3tfu67` (https://www.reddit.com/r/HighQualityGifs/comments/3tfu67/meanwhile_inside_reddit_servers/)
* Note the excpetion
* change the sort from `New` to something else
* Note stale indicator still present

For reference, when doing so I get this in the console:



image

schwers:
:nail_care: :(

ajacksified:
:haircut:

curioussavage:
going to :runner: on this one

schwers:
:fish: :tropical_fish:

PR #435, opened 2016-01-08
Move static properties to the top of the class

nramadas:
Since we talked about it, I figured I would make a patch to address it.

I tried to preserve comments as best as I could. Addressing all the todos is out of the scope of this patch.

I also fixed up the import grouping. I noticed that you guys are trying to do "1) Third party / node modules, 2) Constants / propTypes, 3) everything else" so if I caught something breaking that convention, I fixed it.

Checklist:
* [x] Move all the static props in `/components`
* [x] Move all the static props in `/pages`

### DIFF
--- a/src/views/components/CommentPreview.jsx
+++ b/src/views/components/CommentPreview.jsx
@@ -5,7 +5,7 @@ import short from '../../lib/formatDifference';
 
 const subredditRegex = /\/\/www.reddit.com\/r\/([^/]*)/;
 
-function CommentPreview  (props) {
+function CommentPreview(props) {
   const comment = props.comment;
   const submitted = short(comment.created_utc * 1000);
 
--- a/src/views/components/Ad.jsx
+++ b/src/views/components/Ad.jsx
@@ -1,13 +1,18 @@
 import React from 'react';
-
-import constants from '../../constants';
 import { models } from 'snoode';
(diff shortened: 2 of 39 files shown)


ajacksified:
:fish: :tada:

src/views/components/CommentPreview.jsx
```diff
@@ -5,7 +5,7 @@ import short from '../../lib/formatDifference';
 
 const subredditRegex = /\/\/www.reddit.com\/r\/([^/]*)/;
 
-function CommentPreview  (props) {
```
schwers: :no_bell: what do you think of the leaving the space here? @curioussavage and I have been talking about a new linter rule for leaving a space before the paren in function declarations because it makes searching a little easier. In general I don't think we're consistent about it everywhere so we'll probably add a rule / linter for it either way eventually

ajacksified: In my `linting` branch, I have this enabled and set to never: http://eslint.org/docs/rules/space-before-function-paren

I set it to "no space" because the majority of our existing code _doesn't_ have a space, so I went with the easier of the two. (It's also what's already in the examples in our styleguide, although I didn't see a rule specifically calling it out.)

schwers:
:fish: :tropical_fish:

PR #437, opened 2016-01-11
Don't render pagination buttons when there are no listings

schwers:
I've been seeing errors for 'name' of undefined in `ListingPaginationButtons.jsx`. This fix bails out of rendering pagination buttons when there are no listings as the buttons component relies on having them. Doesn't make sense to show them when there are no listings either.

### DIFF
--- a/src/views/components/ListingContainer.jsx
+++ b/src/views/components/ListingContainer.jsx
@@ -9,6 +9,21 @@ import ListingPaginationButtons from './ListingPaginationButtons';
 const Proptypes = React.PropTypes;
 
 class ListingContainer extends BaseComponent {
+  static propTypes = {
+    compact: Proptypes.bool,
+    shouldPage: Proptypes.bool,
+    listingClassName: Proptypes.string,
+    listings: Proptypes.array,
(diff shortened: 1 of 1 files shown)


src/views/components/ListingContainer.jsx
```diff
@@ -47,7 +47,7 @@ class ListingContainer extends BaseComponent {
     let pagination;
 
     // Default paging to `true`
-    if (shouldPage || typeof shouldPage === 'undefined') {
+    if ((shouldPage || typeof shouldPage === 'undefined') && listings.length) {
```
ajacksified: While you're at it, maybe set `shouldPage = true` in the `const { ... } = this.props` above and remove the `undefined` check

ajacksified:
:nail_care: to fix my bad code :(

schwers:
@ajacksified I updated to used React's defaultProps for `shouldPage` so it's more documented. Look okay? :nail_care:

schwers:
:haircut:

ajacksified:
:fish:

PR #438, opened 2016-01-11
Show the no results warning correctly

schwers:
Fixes `search.jsx` to show the no results message when expected.

Removed the var `noResult` as it was impossible for it have any value where it was being used. Cleans up the checks a little.

### DIFF
--- a/src/views/pages/search.jsx
+++ b/src/views/pages/search.jsx
@@ -129,6 +129,21 @@ class SearchPage extends BasePage {
     this.props.app.redirect(url);
   }
 
+  shouldShowNoResultsMessage(data) {
+    if (!data.search) {
+      return true;
+    }
+
+    const props = this.props;
+
+    // If we're searching within a subreddit and have no results don't
+    // render the banner. The generic case will handle linking to the
+    // site-wide search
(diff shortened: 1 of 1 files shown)


src/views/pages/search.jsx
```diff
@@ -113,18 +113,18 @@ class SearchPage extends BasePage {
...
     let controls;
-    let noResult;
 
     if (!state.loaded && props.ctx.query && props.ctx.query.q) {
       controls = (
         <Loading />
       );
-    } else if (!this.state.data.search) {
-      const noResClass = noResult && props.ctx.query.q ? '' : 'hidden';
+    } else if (!data.search || data.search.links.length === 0) {
```
ajacksified: should this also check that `data.search.links` is non-undefined?

ajacksified:
Just a question, :nail_care:

schwers:
I'll add the check back in. If we don't get any response from the api it looks like it can be undefined

schwers:
@ajacksified I made the check a little smarter so you can search sitewide if there's no results in the subreddit. :haircut:

ajacksified:
:fish:

PR #439, opened 2016-01-11
Restyle wiki

curioussavage:
:eyeglasses: @ajacksified @schwers @nramadas 

https://reddit.atlassian.net/browse/MOBILEWEB-365

* [x]  Revisions link should take you to the revisions for that page only

just cleaning up the wikiPage, made revisions look better. also revision link takes you to revisions for the current page only as long as you are not on the pages route.

depends on: https://github.com/reddit/snoode/pull/71

### DIFF
--- a/assets/less/components/TextSubNav.less
+++ b/assets/less/components/TextSubNav.less
@@ -17,17 +17,25 @@
   }
 }
 
-.TextSubNav-title {
-  padding-top: @grid-size;
-  font-weight: bold;
-}
-
 .TextSubNav-ul {
   margin-bottom: 0;
 }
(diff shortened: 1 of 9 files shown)


assets/less/components/TextSubNav.less
```diff
@@ -17,39 +17,41 @@
   }
 }
 
-  .TextSubNav-ul {
-    margin-bottom: 0;
+.TextSubNav-ul {
```
schwers: :no_bell: Thank you for fixing the indentation

ajacksified:
:fish:

PR #440, opened 2016-01-11
Update client env to be client, with fix for undefined ctx

schwers:
:eyeglasses: @ajacksified https://github.com/schwers/reddit-mobile/blob/03fd02fc939c3fae0dca97b273425fc4cbcf49e1/src/app-mixin.jsx#L29 was the line I had to change


ajacksified:
:fish:

PR #442, opened 2016-01-12
Bodylayout: include on everypage

curioussavage:
BodyLayout now takes an optional prop that will hide
the topnav. All pages can now be rendered through body
layout so we can use it for app wide errors and notifications.

@schwers or @nramadas 

For the new login/registration page, which does not have the `topNav` anymore. Instead of rendering it with a custom body function We now pass a flag to indicate the `topNav` should be hidden. 

Also makes it so `bodylayout` can own the toasts since it will always be rendered.

### DIFF
--- a/src/views/layouts/BodyLayout.jsx
+++ b/src/views/layouts/BodyLayout.jsx
@@ -11,6 +11,11 @@ import CommunityOverlayMenu from '../components/CommunityOverlayMenu';
 import { userData } from '../../routes';
 
 class BodyLayout extends BasePage {
+  static propTypes = {
+    hideTopNav: React.PropTypes.bool,
+    showEUCookieMessage: React.PropTypes.bool,
+  };
+
   constructor(props) {
     super(props);
 
@@ -35,9 +40,13 @@ class BodyLayout extends BasePage {
     }
   }
 
-  render () {
-    const { user, userSubscriptions } = this.state.data;
-    const { app, globalMessage, showEUCookieMessage } = this.props;
+  renderTopNavIfVisible(props, state) {
+    const { user, userSubscriptions } = state.data;
+    const { app, globalMessage, showEUCookieMessage, hideTopNav } = props;
+
+    if (hideTopNav) {
+      return null;
+    }
 
     let messages = [];
     if (showEUCookieMessage) {
@@ -49,20 +58,31 @@ class BodyLayout extends BasePage {
     }
 
     return (
-      <div className='container-with-betabanner'>
+      <div>
         <CommunityOverlayMenu
           {...this.props}
           user={ user }
           subscriptions={ userSubscriptions }
+          key='communitymenu'
         />
-        <UserOverlayMenu {...this.props} user={ user } />
-        <TopNav {...this.props} user={ user } />
+        <UserOverlayMenu {...this.props} user={ user } key='usermenu' />
+        <TopNav {...this.props} user={ user } key='topnav' />
         <InfoBar
-          key='onlyRenderThisOnceEver'
           messages={ messages }
           app={ app }
           showEUCookieMessage={ showEUCookieMessage }
+          key='onlyRenderThisOnceEver'
         />
+      </div>
+    );
+  }
+
+  render () {
+    const topNavIfVisible = this.renderTopNavIfVisible(this.props, this.state);
+
+    return (
+      <div className='container-with-betabanner'>
+        { topNavIfVisible }
         <main>
           { this.props.children }
         </main>
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -552,11 +552,8 @@ function routes(app) {
       return this.redirect('/login?originalUrl=' + submitUrl);
     }
 
-    this.body = function(props) {
-      return (
-        <SubmitPage key={ this.key } {...props}/>
-      );
-    };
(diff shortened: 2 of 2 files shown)


src/views/layouts/BodyLayout.jsx
```diff
@@ -25,16 +30,22 @@ class BodyLayout extends BasePage {
...
           app={ app }
           showEUCookieMessage={ showEUCookieMessage }
-        />
+         />),
+      ];
+    }
+
+    return (
+      <div className='container-with-betabanner'>
+        { topNav }
```
schwers: This will have to be rebased now that topnav is out

src/views/layouts/BodyLayout.jsx
```diff
@@ -25,16 +30,22 @@ class BodyLayout extends BasePage {
...
-    return (
-      <div className='container-with-betabanner'>
-        <SideNav {...this.props} user={ user } subscriptions={ userSubscriptions } />
-        <TopNav {...this.props}/>
-        <InfoBar
-          key='onlyRenderThisOnceEver'
+    let topNav;
+    if (!hideTopNav) {
+      topNav = [
+        (<SideNav {...this.props} user={ user } subscriptions={ userSubscriptions } />),
```
schwers: React will want to have key's on these as there in an array (warnings in debug)

src/views/layouts/BodyLayout.jsx
```diff
@@ -8,13 +8,18 @@ import BasePage from '../pages/BasePage';
...
+  };
+
   constructor(props) {
     super(props);
   }
 
   render () {
     const { user, userSubscriptions } = this.state.data;
-    const { app, globalMessage, showEUCookieMessage } = this.props;
+    const { app, globalMessage, showEUCookieMessage, hideTopNav } = this.props;
```
schwers: I would move these variable declarations and the topnav var into a function. Then the template will read really cleany with a function call to something like `renderTopNavIfVisible()`.

schwers:
Let me work on toasts more to see if this is necessary

schwers:
:nail_care:

curioussavage:
:haircut: I'm all for doing this either way to make things consistent in routes.jsx

src/views/layouts/BodyLayout.jsx
```diff
@@ -35,34 +40,47 @@ class BodyLayout extends BasePage {
...
-    const { app, globalMessage, showEUCookieMessage } = this.props;
+  renderTopNavIfVisible(props, state) {
+    const { user, userSubscriptions } = state.data;
+    const { app, globalMessage, showEUCookieMessage, hideTopNav } = props;
 
-    let messages = [];
-    if (showEUCookieMessage) {
-      messages.push({type: constants.messageTypes.EU_COOKIE});
-    }
+    if (!hideTopNav) {
```
schwers: Small style nit: This, to me, reads similarly to `if (!notDoingThing){...}` Which is a little confusing because of the double negative. Can you swap the order of the cases so it's just `if (hideTopNav) { return null }...`

src/views/layouts/BodyLayout.jsx
```diff
@@ -35,34 +40,47 @@ class BodyLayout extends BasePage {
...
+        (<TopNav {...this.props} user={ user } key='topnav' />),
+        (<InfoBar
           messages={ messages }
           app={ app }
           showEUCookieMessage={ showEUCookieMessage }
-        />
+          key='onlyRenderThisOnceEver'
+         />),
+      ];
+    } else {
```
schwers: New Linter :tm: will want you to remove the `else {...}` and just return. See [no-else-return](http://eslint.org/docs/rules/no-else-return.html)

src/views/layouts/BodyLayout.jsx
```diff
@@ -35,34 +40,47 @@ class BodyLayout extends BasePage {
...
+      return null;
+    }
+  }
+
+  render () {
+    const topNav = this.renderTopNavIfVisible(this.props, this.state);
+
+    return (
+      <div className='container-with-betabanner'>
+        { topNav }
```
schwers: `renderTopNavIfVisible` nicely captures the potential null value in `topNav`. I have a preference for preserving that either by changing the `topNav` variable name or just calling `renderTopNavIfVisible` inline

schwers:
Looks good in testing. Just a few nits / new linter comments. :nail_care:

src/views/layouts/BodyLayout.jsx
```diff
@@ -35,34 +40,47 @@ class BodyLayout extends BasePage {
...
-      messages.push(globalMessage);
-    }
+      if (globalMessage) {
+        messages.push(globalMessage);
+      }
 
-    return (
-      <div className='container-with-betabanner'>
-        <CommunityOverlayMenu
+      return [
```
nramadas: out of curiosity, why do you need to return an array of elements here?

nramadas: nvm, i see it

curioussavage:
:haircut: layout looks good. tested for functionality. would appreciate some more sanity checks for component rendering issues.

schwers:
Tested a bunch, component creation versus update characteristics don't appear to have changed. LGTM :fish:

PR #443, opened 2016-01-13
Linting

ajacksified:
mfw I'm trying to clean up the codebase

:eyeglasses: @curioussavage and @schwers and @nramadas 

Good luck.


ajacksified:
:haircut: quick, test it before the code changes again

src/routes.jsx
```diff
@@ -300,9 +302,10 @@ function routes(app) {
   router.use(defaultLayout);
   router.use(globalMessage);
 
-  function * indexPage () {
-    let props = this.props;
-    var sort = this.query.sort || 'hot';
+
+  function *indexPage () {
```
nramadas: did we decide if we're remove spaces between function names and the parens?

ajacksified: Oh, good question. I don't think this has a rule for that yet. We should add a rule after this PR for that. What's your opinion? I think I lead towards no space.

nramadas: Yeah, I like no space as well

curioussavage: no space :+1:

curioussavage:
I say :fish: :shipit:

schwers:
:fish:

PR #447, opened 2016-01-15
Fix null listing url check (log cleanup)

schwers:
:eyeglasses: @ajacksified or @curioussavage or @nramadas 

I've noticed this error popping up in the logs. An alternative would be to provide empty string default via Snoode but this might be the only case where this is an issue


curioussavage:
:fish: lgtm

ajacksified:
I think it makes sense that it returns undefined rather than empty string. :fish:

PR #448, opened 2016-01-15
Fix match of undefined error (log cleanup)

schwers:
Been noticing this error in the log. It looks like some of the requests don’t have a userAgent so that could be the real underlying issue. But in a lot of cases it looks like we check for user-agent not being null so this seems like a valid fix.

:eyeglasses: @ajacksified


curioussavage:
:fish:

ajacksified:
Null UA is accepted by desktop for privacy reasons; this is fine. :fish:

PR #449, opened 2016-01-15
Remove overflow:hidden so subreddit search works

ajacksified:
:eyeglasses: @curioussavage - do you know why this was set to hidden? Search is broken right now (you can't scroll) without this set to visible. Seems fine without overflow:hidden.


curioussavage:
lgtm, having the scrollbar is much better than not being able to scroll.

:fish:

curioussavage:
actually if you increase the height on `.sub-selection-selected` by a pixel or two it no longer scrolls. looks like its content is slightly overflowing

curioussavage:
@ajacksified It has to be at least 50px

ajacksified:
closed for https://github.com/reddit/reddit-mobile/pull/450

PR #450, opened 2016-01-15
remove overflow css property on submit page wrapper.

curioussavage:
This was preventing search from scrolling while posting.
Also increases height of the bottom div to fix scrollbar

:eyeglasses: @ajacksified


ajacksified:
:fish:

PR #454, opened 2016-01-19
Pull isGoogleCrawler from ctx, not props

ajacksified:
And a little bit of cleanup.

:eyeglasses: @schwers or @curioussavage


schwers:
:fish:

PR #456, opened 2016-01-20
Unjankify iOS Chrome scrolling again.

schwers:
It looks like scrolling is awry again. No recent changes in the code
would have caused this. In-fact the issue is present when
time-traveling back to 94cda4847f15ea1e4b8a12b42a3b7b42dce6a319

Google Chrome has no updates listed in the App Store, so I think it’s
an issue with either iOS version 9.2 or 9.3 (testing with 9.3)

:eyeglasses: @ajacksified and @curioussavage (not sure what version you're on Adam, but let me know. Let's do a round of more testing in the morning)

### DIFF
--- a/assets/less/components/Listing.less
+++ b/assets/less/components/Listing.less
@@ -22,14 +22,6 @@
 }
 
 .Listing {
-  position: relative;
-  // Chrome on iOS won't pick this up, but only browsers like Safari
-  // and chrome on android will. Chrome on iOS can't handle the memory
-  // usage or something and the scroll janks. They might not be using
-  // WKWebView which would explain the huge performance difference
-  // than iOS Safari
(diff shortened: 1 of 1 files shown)


schwers:
Update: Tested with @umbrae and @curioussavage and 9.1 and 9.2 are fine. So it might just be 9.3(beta 1). I'm somewhat inclined to test 9.1 and 9.2 against this patch and move forward with it if it's fine on all platforms so we're ready for 9.3. Thoughts?

ajacksified:
Sounds good. Test and :fish: if good!

curioussavage:
looks good on 9.2

umbrae:
Also tested in 9.1, seems good.

PR #457, opened 2016-01-20
Change the look of comments

nramadas:
Large(ish) patch. Aims to completely re-do the comments on mobileweb.

Addresses: https://reddit.atlassian.net/browse/MOBILEWEB-335

**Key changes:**
1) New `<Comment />` component to replace the old.
2) New `<DropdownController />` component to create a simpler dropdown.
3) Brand new design for comments, thanks to our design team.
4) A few new icons.

**Feature checklist (really, regression testing):**
* [ ] Reply to posts
* [ ] Reply to comments
* [ ] Upvote/Downvote comments
* [ ] Edit comments
* [ ] Report comments
* [ ] Permalink, save comments, user profiles, etc

**Engineering checklist:**
* [ ] More stateless subcomponents
* [ ] More centralized logic; the `<Comment />` component does more
* [ ] Two new generic Form Elements
* [ ] CSS (less) is now split into subfolders

**How to review:**
I would start with `/comment/Comment.jsx` as it holds the bulk of the logic. The subcomponents you can review as you get to them. Really, they are fairly simple and single-purposed, so there shouldn't be too much to review there.

**Screenshots:**
screen shot 2016-01-20 at 1 08 04 pm

screen shot 2016-01-20 at 1 08 37 pm

screen shot 2016-01-20 at 1 09 14 pm

screen shot 2016-01-20 at 1 09 40 pm

screen shot 2016-01-20 at 1 12 28 pm

:eyeglasses: @schwers @ajacksified @curioussavage


assets/less/components/comment/CommentDropdownContent.less
```diff
@@ -0,0 +1,78 @@
+.CommentDropdownContent {
+  padding: 10px;
```
schwers: :no_bell: Small-ish question. From skimming this and and CommentHeader.less there seem to be a lot of padding/margins that are awfully close to @grid-size and @half-grid-size. Tavish mentioned wanted to try to keep everything on the grid or half grid. Just wanted to double check if these are intentionally not using those values

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,563 @@
...
+  } else if (distinguished) {
+    return distinguished;
+  } else if (author === op) {
+    return 'op';
+  } else {
+    return '';
+  }
+};
+
+function extractErrorMsg(error) {
```
schwers: This looks the same as https://github.com/reddit/reddit-mobile/pull/457/files#diff-00173e02b61a7e6a7d347c9f1e789c9dR7. Is there are reason to not make them common somewhere?

assets/less/components/comment/CommentDropdownContent.less
nramadas: that's a good point. some of this was written before the gospel of `8px` increments was taught to me. I'll take a pass at standardizing some of it.

schwers: Sounds good. I understand it's intentionally not 'on the grid' but it would be neat if it was :)

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,563 @@
...
+
+const V = React.PropTypes;
+
+function determineVoteDirection(likes) {
+  if (likes === true) return 1;
+  if (likes === false) return -1;
+  return 0;
+};
+
+function determineAuthorType(distinguished, author, op, user) {
```
schwers: :no_bell: More of a discussion. I've been considering proposing something for enums like this. We can use `React.PropTypes.oneOf` to make types. For this it would look like ```const AuthorType = React.PropTypes.oneOf(['self', 'op', ''])```. It would get us warnings in dev when invalid values are passed in.

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,563 @@
...
+      id: this.state.comment.name,
+    });
+  }
+  
+  buildApiOptions() {
+    return this.props.app.api.buildOptions(this.props.apiOptions);
+  }
+  
+  redirectToUserProfile() {
+    window.location = `/u/${this.state.comment.author}`;
```
schwers: Do these, and some of the other actions passed to comment tools, need to be hard redirects? An alternative I can see would be just making those rows/actions normal links. Normal links would prevent a full refresh of the page as they get caught by the router. (For reference that code is in [client.es6.js](https://github.com/reddit/reddit-mobile/blob/master/assets/js/client.es6.js#L361))

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,563 @@
...
+    const currentVoteDirection = this.state.voteDirection;
+    
+    try {
+      await this.ensureUserLoggedIn();
+      
+      const undoVote = (delta - currentVoteDirection) === 0;
+      const newVoteDirection = undoVote ? 0 : delta;
+      const vote = this.buildVote(newVoteDirection);
+      
+      // In determining the new score:
```
schwers: :no_bell: Thank you for adding the explanation. This was tricky to understand before.

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,563 @@
...
+      // do nothing;
+    }
+  }
+  
+  async ensureUserLoggedIn() {
+    // Ensure the user is logged in by validating that the token exists
+    if (this.props.token) {
+      return true;
+    } else {
+      window.location = this.props.app.config.loginPath;
```
schwers: Can you make sure the `originalURL` is set so we're sure to come back to the same place? 

@ajacksified Is there any consensus on if we should land users on register instead of login (given that most are probably anon and don't have accounts)?

src/views/components/comment/CommentDropdownContent.jsx
```diff
@@ -0,0 +1,186 @@
+import React        from 'react';
+import SquareButton from '../formElements/SquareButton';
+
+const V = React.PropTypes;
```
schwers: I'm all for using an alias for PropTypes as it cleans things up. Just want to make a note that we should standardize. `V` makes me think of Validators, which they kind of are. I could see an argument for 'T' for types as well.

src/views/components/comment/Comment.jsx
ajacksified: (If not, use `app.redirect(URL)`)

ajacksified: Disregard, use normal links. App.redirect is for non-link redirects, like
after a form post.

assets/less/components/comment/CommentDropdownContent.less
```diff
@@ -0,0 +1,78 @@
...
+      color: @neutral;
+      font-size: 22px;
+      line-height: 30px;
+      
+      &:before {
+        color: @neutral; // for the gold icon
+      }
+      
+      &.icon-save.m-selected {
+        &:before { // why???
```
schwers: Alternatively you can just the `.lime` modifier. As to why... class selectors :( https://github.com/reddit/reddit-mobile/blob/master/assets/less/components/icon-font.less#L15

src/views/components/comment/CommentDropdownContent.jsx
nramadas: `Validators` is why i've been using V. I'm happy with either `V` or `T`. Both beat writing out `React.PropTypes` each time.

assets/less/components/comment/CommentDropdownContent.less
```diff
@@ -0,0 +1,78 @@
...
+    color: @dark-grey;
+    font-size: 18px;
+    line-height: 30px;
+  }
+  
+  &__reportForm {
+    position: relative;
+    padding-right: 96px;
+    height: 100%;
+    width: calc(~'100% - 40px');
```
schwers: :no_bell: Just a note that we talked about calc and browsers in person for anyone reading the PR comments

src/views/components/LinkTools.jsx
```diff
@@ -0,0 +1,125 @@
+import React            from 'react';
```
ajacksified: don't line up imports (goes against linter rules)

src/views/components/comment/CommentDropdownContent.jsx
ajacksified: I prefer T, fwiw.

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,551 @@
...
+        }),
+      };
+      
+      this.props.app.emit('report', this.state.comment.id);
+    } catch(e) {
+      // do nothing;
+    }
+  }
+  
+  async ensureUserLoggedIn() {
```
ajacksified: Why is this async?

src/views/components/LinkTools.jsx
```diff
@@ -0,0 +1,125 @@
...
+      
+      this.props.onNewComment(newComment);
+    } catch(e) {
+      this.setState({
+        error: extractErrorMsg(e),
+      });
+    }
+  }
+  
+  async ensureUserLoggedIn() {
```
ajacksified: why is this async?

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,551 @@
...
+      </div>
+    );
+  }
+  
+  renderBody() {
+    const { collapsed, comment } = this.state;
+    const bodyText = mobilify(comment.body_html);
+    
+    let cls = 'Comment__body';
+    if (collapsed) cls += ' m-hidden';
```
ajacksified: Use braces, per styleguide.

src/views/components/comment/CommentDropdownContent.jsx
schwers: Agreed

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,551 @@
...
+    } catch(e) {
+      // do nothing;
+    }
+  }
+  
+  async ensureUserLoggedIn() {
+    // Ensure the user is logged in by validating that the token exists
+    if (this.props.token) {
+      return true;
+    } else {
```
ajacksified: Since the `if` returns, don't use an `else`; just throw.

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,551 @@
...
+        />
+      </div>
+    );
+  }
+  
+  renderLockedReplyForm() {
+    const { collapsed, showReplyBox } = this.state;
+    
+    let cls = 'Comment__replyForm';
+    if (collapsed || !showReplyBox) cls += ' m-hidden';
```
ajacksified: braces

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,551 @@
...
+            token={ this.props.token }
+            op={ this.props.op }
+            user={ this.props.user }
+            nestingLevel={ this.props.nestingLevel + 1}
+            permalinkBase={ this.props.permalinkBase }
+            repliesLocked={ this.props.repliesLocked }
+          />
+        </div>
+      );
+    } else {
```
ajacksified: no `else` since the above condition returns (here and elsewhere)

src/views/components/comment/CommentHeader.jsx
```diff
@@ -0,0 +1,113 @@
...
+};
+
+function renderDots(count) {
+  const content = fill(Array(count), '•').join(' ');
+  
+  return <div className='CommentHeader__dots'>{ content }</div>
+};
+
+function renderCaron(collapsed, dots) {
+  const style = dots ? {width: 26 + (10 * dots)} : null;
```
ajacksified: spaces in braces

ajacksified:
Some new-linter fails that should be fixed everwhere (new lint file hasn't been merged, so not your fault, but they should be fixed.)

Separately:

* dropdowns should be a separate commit, like we talked about
* this would be a fantastic place to add some tests for new / refactored jsx

:nail_care:

src/views/components/comment/CommentDropdownContent.jsx
curioussavage: :+1:

assets/less/components/comment/CommentDropdownContent.less
curioussavage: *whispers flexbox is the way to go...*

src/views/components/comment/Comment.jsx
curioussavage: This is a good question.

I am in favor of login, because, reasons..

assets/less/components/comment/CommentDropdownContent.less
schwers: _display:table-cell is calling..._

src/views/components/comment/Comment.jsx
ajacksified: I prefer register. It's easy enough to get to login

src/views/components/formElements/SquareButton.jsx
```diff
@@ -0,0 +1,45 @@
...
+function SquareButton(props) {
+  let cls = 'SquareButton';
+  if (!props.enabled) cls += ' m-disabled';
+  
+  function fireClick(e) {
+    if (props.enabled) props.onClick(e);
+  }
+  
+  return (
+    <div
```
curioussavage: Not using a button element hurts accessibility. Not something we have spent much time on but IMO much better to avoid something that directly impacts it.

ajacksified: Good catch - intractable elements should use their most semantically correct html counterpart (button for doing things, link for going places, etc.)

src/views/components/comment/Comment.jsx
schwers: Agreed, I think register is better

nramadas: I'm going to punt on this for now. We don't have a `registerPath` in our config, and I'm hesitant to add one at this point, since this would be the only place where it would be used. I suggest a separate patch to change all our `loginPath` usage to `registerPath`.

nramadas:
:haircut:

assets/less/components/comment/CommentHeader.less
```diff
@@ -0,0 +1,133 @@
...
+  &__flair,
+  &__timestamp,
+  &__separator {
+    margin-right: @half-grid-size;
+  }
+  
+  &__username,
+  &__flair,
+  &__timestamp {
+    color: @neutral;
```
schwers: Can you follow up with design on this color?I have a feeling this should be `@grey-text` just on principle. I think the username / flair / arrows and etc are really hard to read with how light they are compared to the background. Even with `@grey-text` its too light to pass accessibility checks from the likes of http://contrastchecker.com/ which has some supposedly standard metrics.

nramadas: Sure thing

assets/less/components/comment/CommentHeader.less
```diff
@@ -0,0 +1,133 @@
...
+  
+  &__caron {
+    margin-right: @grid-size;
+    
+    &[class*=" icon-"] {
+      color: @blue;
+    }
+    
+    &.m-collapsed {
+      transform: rotate(-90deg);
```
schwers: :no_bell: Did you experiment with add `.tween` to `&__caron`? I believe that would animate the rotation for free

nramadas: cool. i'll add that.

src/lib/extractErrorMsg.es6.js
```diff
@@ -0,0 +1,9 @@
+export default function extractErrorMsg(error) {
+  if (error.errors && error.errors.length) {
```
schwers: Library functions are probably a good example of where we should add tests

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,542 @@
...
+  constructor(props) {
+    super(props);
+    
+    const { comment } = props;
+    
+    this.state = {
+      comment: props.comment, // set as state since editing can change this
+      commentScore: comment.score,
+      commentReplies: comment.replies || [],
+      commentDeleted: comment.author === '[deleted]', // there has GOT to be a better way
```
schwers: It's maybe worth moving this into [Snoode](https://github.com/reddit/snoode) as part of the parsing step for comments and assign it as a bool

src/views/components/comment/CommentDropdownContent.jsx
```diff
@@ -0,0 +1,211 @@
+import React from 'react';
+
+import SquareButton from '../formElements/SquareButton';
+
+const T = React.PropTypes;
+
+export default class CommentDropdownContent extends React.Component {
```
schwers: It looks like this and a few other of these components opt out of using `BaseComponent`, is that intentional?

nramadas: Yes, I wanted to keep these as pure prop-driven react components.

src/views/components/comment/CommentDropdownContent.jsx
```diff
@@ -0,0 +1,211 @@
...
+  
+  renderReportText() {
+    return (
+      <div className='CommentDropdownContent__text' >
+        Report
+      </div>
+    );
+  }
+  
+  renderReportForm() {
```
schwers: Did you have any luck with preventing page scroll in Safari? I've noticed when opening the report form on comments near the top the page is scrolled to the bottom in the background, so the comment you see underneath is misleading.

src/views/pages/listing.jsx
```diff
@@ -1,14 +1,14 @@
-import React from 'react';
-import remove from 'lodash/array/remove';
-import { models } from 'snoode';
+import React                  from 'react';
```
schwers: Looks like the top needs un-aligning here

src/views/components/comment/CommentDropdownContent.jsx
nramadas: I have not. The ideal solution here might be a product one; adding a separate view to handle reporting a comment. Otherwise, we might be forced to do something particularly hacky (and likely won't work well), like scroll the page ourselves, etc.

src/views/pages/listing.jsx
nramadas: ugh, missed one

src/views/components/comment/CommentDropdownContent.jsx
schwers: Ya I agree, This feels like it should be a form with the full comment maybe.

ajacksified:
Design nit: nested "More Comments" buttons don't have indentation bullets
next to them. I noticed it on
https://www.cloudflarestagingformobilereddit.com/r/politics/comments/41vobb/bernie_sanders_stunning_popularity_is_almost/
On Jan 20, 2016 11:55 AM, "Niranjan Ramadas" <notifications@github.com>
wrote:

Large(ish) patch. Aims to completely re-do the comments on mobileweb.

Addresses: https://reddit.atlassian.net/browse/MOBILEWEB-335

*Key changes:*
1) New <Comment /> component to replace the old.
2) New <DropdownController /> component to create a simpler dropdown.
3) Brand new design for comments, thanks to our design team.
4) A few new icons.

*Feature checklist (really, regression testing):*

   - Reply to posts
   - Reply to comments
   - Upvote/Downvote comments
   - Edit comments
   - Report comments
   - Permalink, save comments, user profiles, etc

*Engineering checklist:*

   - More stateless subcomponents
   - More centralized logic; the <Comment /> component does more
   - Two new generic Form Elements
   - CSS (less) is now split into subfolders

*How to review:*
I would start with /comment/Comment.jsx as it holds the bulk of the logic.
The subcomponents you can review as you get to them. Really, they are
fairly simple and single-purposed, so there shouldn't be too much to review
with there.

*Todo Items (will get resolved before shipping):*
1) Remove old <Comment /> component code
2) Remove unneeded styles

[image: :eyeglasses:] @schwers <https://github.com/schwers> @ajacksified
<https://github.com/ajacksified> @curioussavage
<https://github.com/curioussavage>
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/reddit/reddit-mobile/pull/457
Commit Summary

   - Change the look of comments

File Changes

   - *R* assets/fonts/icons-0.15726539911702275.eot
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-0> (0)
   - *R* assets/fonts/icons-0.15726539911702275.ttf
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-1> (0)
   - *A* assets/fonts/icons-0.15726539911702275.woff
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-2> (0)
   - *D* assets/fonts/icons-0.6747430041432381.woff
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-3> (0)
   - *A* assets/icons/mod.svg
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-4> (9)
   - *A* assets/icons/op.svg
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-5> (11)
   - *A* assets/icons/reply2.svg
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-6> (8)
   - *A* assets/icons/snoosilhouette.svg
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-7> (13)
   - *M* assets/js/client.es6.js
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-8> (8)
   - *M* assets/less/_icons.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-9> (114)
   - *M* assets/less/base.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-10> (20)
   - *A* assets/less/components/LinkTools.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-11> (11)
   - *M* assets/less/components/ListingPage.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-12> (10)
   - *M* assets/less/components/comment.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-13> (175)
   - *A* assets/less/components/comment/Comment.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-14> (56)
   - *A* assets/less/components/comment/CommentBody.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-15> (7)
   - *A* assets/less/components/comment/CommentDropdownContent.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-16> (78)
   - *A* assets/less/components/comment/CommentHeader.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-17> (133)
   - *A* assets/less/components/comment/CommentReplyForm.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-18> (30)
   - *A* assets/less/components/comment/CommentSeeMore.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-19> (20)
   - *A* assets/less/components/comment/CommentTools.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-20> (55)
   - *A* assets/less/components/comment/importAll.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-21> (7)
   - *A* assets/less/components/dropdown/DropdownController.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-22> (24)
   - *A* assets/less/components/dropdown/importAll.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-23> (1)
   - *A* assets/less/components/formElements/SquareButton.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-24> (20)
   - *A* assets/less/components/formElements/TextField.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-25> (12)
   - *A* assets/less/components/formElements/importAll.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-26> (2)
   - *M* assets/less/components/layout.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-27> (2)
   - *M* assets/less/mixins.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-28> (8)
   - *M* assets/less/variables.less
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-29> (2)
   - *A* src/views/components/LinkTools.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-30> (134)
   - *A* src/views/components/comment/Comment.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-31> (563)
   - *A* src/views/components/comment/CommentBody.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-32> (20)
   - *A* src/views/components/comment/CommentDropdownContent.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-33> (186)
   - *A* src/views/components/comment/CommentEditForm.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-34> (59)
   - *A* src/views/components/comment/CommentHeader.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-35> (113)
   - *A* src/views/components/comment/CommentReplyForm.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-36> (101)
   - *A* src/views/components/comment/CommentSeeMore.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-37> (33)
   - *A* src/views/components/comment/CommentTools.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-38> (203)
   - *A* src/views/components/dropdown/DropdownController.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-39> (126)
   - *A* src/views/components/formElements/SquareButton.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-40> (45)
   - *A* src/views/components/formElements/TextField.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-41> (27)
   - *M* src/views/pages/listing.jsx
   <https://github.com/reddit/reddit-mobile/pull/457/files#diff-42> (92)

Patch Links:

   - https://github.com/reddit/reddit-mobile/pull/457.patch
   - https://github.com/reddit/reddit-mobile/pull/457.diff

—
Reply to this email directly or view it on GitHub
<https://github.com/reddit/reddit-mobile/pull/457>.

schwers:
Can you change how links (<a />) for `Permalink` and `<User> profile` work in the Dropdown? Right now the links are just the text for that so tapping the icon or the row (to the right of the text) just closes the Dropdown (shouldn't clicks inside the dropdown not close it?) without taking you to the permalink or user profile.

Also permalink'd comments aren't visually distinguished. EG [Staging](https://www.cloudflarestagingformobilereddit.com/r/IAmA/comments/41w2hg/hey_reddit_its_kris_rob_and_dave_from_cyanide/cz5ljk3) vs [Prod](https://m.reddit.com/r/IAmA/comments/41w2hg/hey_reddit_its_kris_rob_and_dave_from_cyanide/cz5ljk3)

schwers:
:nail_care:

nramadas:
:haircut:

ajacksified:
:fish:

ajacksified:
Let's take another pass after rebase; my rebased branch can't reply or vote. :haircut:

src/views/components/comment/CommentDropdownContent.jsx
```diff
@@ -0,0 +1,215 @@
...
+const T = React.PropTypes;
+
+export default class CommentDropdownContent extends React.Component {
+  static propTypes = {
+    username: T.string.isRequired,
+    userOwned: T.bool,
+    saved: T.bool,
+    permalinkUrl: T.string,
+    userLoggedIn: T.bool,
+    onEditClicked: T.func,
```
curioussavage: Why not make these required and drop the fields in default props?

nramadas: It's my understanding that a required prop should be something that is absolutely required for the component to render. Callbacks are basically never a requirement, so I don't specify them as such.

However, the reason I default them is because I want to avoid ever doing an existential check.

So, this format both makes the callback optional in the component's "API", but simplifies the component code by avoiding existentials.

src/views/components/comment/CommentHeader.jsx
```diff
@@ -0,0 +1,133 @@
...
+      </div>
+      { separator }
+      { flair ? renderFlair(flair, highlight) : null }
+      { flair ? separator : null }
+      { renderTimestamp(created, highlight) }
+    </td>
+  );
+}
+
+function renderFlair(flair, highlight) {
```
curioussavage: There was an issue with long flair with the old design. Was the max length considered here? looks like it just makes the table cell really tall.

nramadas: The flair should wrap

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,545 @@
...
+      });
+    }
+  }
+  
+  async submitVote(delta) {
+    const currentScore = this.state.commentScore;
+    const currentVoteDirection = this.state.voteDirection;
+    
+    try {
+      if (this.props.app.needsToLogInUser()) { return; }
```
schwers: Does this need to be in the try{...}? It looks like this could just be the first line of the function.

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,545 @@
...
+    } catch (e) {
+      this.setState({
+        loadingMoreComments: false,
+      });
+    }
+  }
+  
+  async submitCommentEdit(newCommentText) {
+    try {
+      if (this.props.app.needsToLogInUser()) { return; }
```
schwers: Same question about this being in the try. I think it might be a little clearer that it's just a guard and won't throw an exception if it it's outside of the try{}

src/views/components/LinkTools.jsx
```diff
@@ -0,0 +1,119 @@
...
+  
+  handleFormChange(value) {
+    this.setState({
+      commentValue: value,
+    });
+  }
+  
+  async submitNewComment() {
+    try {
+      if (!this.ensureUserLoggedIn()) { return; }
```
schwers: I think this is a rebase artifact and should be the new function off app. This looks like it will throw a calling undefined error

src/views/components/comment/Comment.jsx
schwers: :no_bell: What was the conclusion of the deleted flag? Is someone adding that at the api layer someday?

src/views/components/comment/CommentDropdownContent.jsx
schwers: Maybe worth noting: React is perfectly happy to be passed an undefined function for `onClick` and the like. Most of these don't _technically_ need to have defaults. One option would be to only define the defaults for callbacks the component is going to call manually. That might serve as a way of documenting what's intended to be a react callback versus a component callback. But this is mostly just a style / philosophical thing.

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,545 @@
...
+
+const T = React.PropTypes;
+
+function determineVoteDirection(likes) {
+  if (likes === true) { return 1; }
+  if (likes === false) { return -1; }
+  return 0;
+}
+
+function determineAuthorType(distinguished, author, op, user) {
```
curioussavage: I think this is also a good candidate for something to move into snoode

src/views/components/comment/CommentDropdownContent.jsx
curioussavage: I don't think there is any benefit from defining a default prop for these because these components aren't useable without providing the methods. 

After testing it looks like if there is as default prop you don't get the warning that the prop was not included which negates half the benefit of defining a proptype.

The prop might not be required for the code to not throw an error, but it is required for the function to do what it is intended to do.

curioussavage:
Just a broad comment/question about the PR. 

My first impression when reading through it is that it looks like a different codebase. There are plenty of different styles and conventions for writing React/CSS used here that don't match the existing code. After reading through again I think there are less than my initial impression, some we already talked about. I think to keep the codebase cohesive It would be better to stick to the existing styles in other files and the Javascript style guide.

One example would be the usage of (BEM)[https://en.bem.info/method/] methodology for writing LESS as well as the nesting. I am actually in favor of BEM, but although I believe we talked about it, I don't think we came to a conclusion about using it. This is more than just a little style preference, it is an entirely different methodology for writing css. kind of a big change.

tldr; There are a lot of significant unique style differences between this PR and the rest of the code base. Can we either follow existing styles more closely or make some calls about patterns and styles used.

curioussavage:
Can the file/directory names all be lower case per the style guide?

Unless we want to change it to some other convention we should stick to it. I have already been bitten by issues with import path casing in ubuntu. We can even enforce the convention with ESlint.

src/views/components/comment/CommentDropdownContent.jsx
nramadas: Let's talk about this in person.

src/views/components/comment/Comment.jsx
nramadas: I included it in the try/catch since the catch block handlers errors. Right now, we don't have any sort of messaging available, but I put these here with the thought that once we have the error toaster set up, we can use that to message when something goes wrong.

src/views/components/LinkTools.jsx
nramadas: Gah, i thought I got all of them.

schwers:
@curioussavage Out of curiosity what happens with importing camel case files? Is it just if you have two different files that when lowercased are the same name? It looks like we use camelcasing everywhere in components atm. I agree we should add something in ESlint for it though

schwers:
RE: BEM methodology for LESS. I'm planning on using it for listings. We've talked about improving css/less architecture and removing bootstrap. Moving to BEM and removing bootstrap is probably something best done incrementally. If we all agree I think we should just say new, large, chunks of less should be in BEM, like this patch. Then we can clean up as we edit and restyle portions of the project; eventually a semi-large patch (versus a large all at once) patch could finish cleaning up styles.

src/views/components/comment/Comment.jsx
schwers: I think it would be clearer to just move it to the top of the function. `needsToLogInUser` doesn't throw an error that can be consumed by the catch statement. It's intention is just be a guard or assertion check before you call stuff that _will_ error you're not logged in. In the meantime `needsToLoginUser` handles redirecting to `/register` for indicating you need to log in. It should probably be where any toasts or modals are draw in the future for this specific case.

ajacksified: :+1: on moving towards BEM.

nramadas: gotcha. i'll move them out.

curioussavage:
I don't mind camel case at all but we did add a rule to the style guide already saying to use lowercase. I'm all for doing camel case but the advantage of all lowercase is we can make linter rule to enforce it. 

How would we enforce camel case?  another option Niranjan offered was underscores

ajacksified:
:+1: to BEM.

About filenames: either all lower or camelcase would be preferable to introducing underscores, imo. All lowercase keeps the rule simple and lessens the chance of case-sensitive filesystems breaking. On the bright side, we have travis testing now, so we'll know if it fails before we try to deploy, so maybe that lets us go with the nicer option of camelcase?

src/views/components/comment/Comment.jsx
ajacksified: I disagree - you don't necessarily know if `self` (because you have no concept of the user during `comments.get`), and this is mostly just to check the existing flags to build UI.

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,545 @@
+import React from 'react';
+import { models } from 'snoode';
+import without from 'lodash/array/without';
+
+import mobilify from '../../../lib/mobilify';
+import extractErrorMsg from '../../../lib/extractErrorMsg';
+import BaseComponent   from '../BaseComponent';
```
ajacksified: extra space before the `from`

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,545 @@
...
+import CommentHeader from './CommentHeader';
+import CommentBody from './CommentBody';
+import CommentTools from './CommentTools';
+import CommentReplyForm from './CommentReplyForm';
+import CommentSeeMore  from './CommentSeeMore';
```
ajacksified: extra space before the `from`

src/views/components/comment/Comment.jsx
```diff
@@ -0,0 +1,545 @@
...
+    );
+  }
+  
+  renderReply(reply) {
+    const { loadingMoreComments } = this.state;
+    const { nestingLevel, highlightedComment } = this.props;
+    const replyCallback = () => this.loadMoreComments(reply);
+    
+    let cls = 'Comment__reply';
+    if (reply.id === highlightedComment) cls += ' m-highlight';
```
ajacksified: needs `{ }`

test/lib/extractErrorMsg.es6.js
```diff
@@ -0,0 +1,36 @@
+import chai from 'chai';
+import sinon from 'sinon';
```
ajacksified: no need to import `sinon`

ajacksified:
Good news! With those ^ 4 fixes, this rebases very happily on top of the megalinting branch. :haircut:, and then I suspect a :f ish: soon to follow.

nramadas:
:haircut:

ajacksified:
:fish:

src/views/components/comment/Comment.jsx
curioussavage: very good point

curioussavage:
:fish:

PR #458, opened 2016-01-21
pre-commit: lint only changed js files

curioussavage:
:eyeglasses: @schwers or @ajacksified or @nramadas 

this will only lint staged js files on pre commit

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -70,7 +70,7 @@ function loadShim() {
   shimScript.type = 'text\/javascript';
   shimScript.onload = function() {
     initialize(false);
-  }
+  };
 
   $head.appendChild(shimScript, document.currentScript);
 
@@ -159,17 +159,17 @@ function refreshToken (app) {
 
         Object.assign(app.getState('ctx'), {
           token: token.token,
-          tokenExpires: token.tokenExpires
+          tokenExpires: token.tokenExpires,
         });
 
         app.setState('refreshingToken', false);
         app.emit('token:refresh', token);
 
         window.setTimeout(function() {
-          refreshToken(app).then(function(){
+          refreshToken(app).then(function() {
             Object.assign(app.getState('ctx'), {
               token: token.token,
-              tokenExpires: token.tokenExpires
+              tokenExpires: token.tokenExpires,
             });
 
             app.setState('refreshingToken', false);
@@ -177,7 +177,7 @@ function refreshToken (app) {
           });
         }, (expires - now) * .9);
       });
-  })
+  });
 }
 
 function findLinkParent(el) {
@@ -216,7 +216,7 @@ function sendTimings() {
         .send({
           rum: timings,
         })
-        .end(function(){});
+        .end(function() {});
     }
   }
 }
@@ -238,8 +238,6 @@ function render (app, ...args) {
 
 function initialize(bindLinks) {
   const dataCache = window.bootstrap.dataCache;
-  var plugin;
-  var p;
 
   referrer = document.referrer;
 
@@ -280,7 +278,7 @@ function initialize(bindLinks) {
     refreshMS = Math.max(refreshMS - (1000 * 60), 0);
 
     window.setTimeout(function() {
-      refreshToken(app).then(function(){});
+      refreshToken(app).then(function() {});
     }, refreshMS);
   } else if (!cookies.get('loid')) {
     setLoggedOutCookies(cookies, app);
@@ -292,8 +290,7 @@ function initialize(bindLinks) {
 
   // env comes from bootstrap from the server, update now that the client is loading
   app.state.ctx.env = 'CLIENT';
-  modifyContext = modifyContext.bind(app);
-  app.modifyContext = modifyContext;
+  app.modifyContext = modifyContext.bind(app);
 
   var history = window.history || window.location.history;
   app.pushState = (data, title, url) => {
@@ -317,7 +314,7 @@ function initialize(bindLinks) {
 
     // Set to the browser's interpretation of the current name (to make
     // relative paths easier), and send in the old url.
-    render(app, app.fullPathName(), false, modifyContext).then(function(props) {
+    render(app, app.fullPathName(), false, app.modifyContext).then(function(props) {
       setTitle(props);
       postRender(path);
     });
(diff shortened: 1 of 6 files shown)


curioussavage:
I could also do this for pushing but as far as I know I would need to use `git diff --name-only <upstream>/master` so we would need a convention for the main repo name.

curioussavage:
I realized client.es6.js was not being linted. last two commits fixes that

curioussavage:
bleh, I didn't realize that what branch I was on when I added the last 3 commits, still a relatively small PR. Unless anyone thinks its an issue I will leave them on this pr

curioussavage:
:eyeglasses: @dwick  for the ads files

assets/js/frames.es6.js
```diff
@@ -164,10 +163,10 @@ const frames = {
     let namespaceProxies = proxies[namespace];
 
     if (namespaceProxies) {
-      namespaceProxies.targets = [].concat(namespaceProxies.targets, target);
+      namespaceProxies.targets = [].concat(namespaceProxies.targets, targets);
```
curioussavage: this looked like a typo..

assets/js/frames.es6.js
```diff
@@ -183,19 +182,18 @@ const frames = {
    * param {Object} [context=this] The context the callback is invoked with.
    * returns {Object} The listener.
    */
-  receiveMessageOnce: function (source, type, callback, context) {
-    return frames.receiveMessage(source, type, function() {
+  receiveMessageOnce(source, type, callback, context) {
+    const listener = frames.receiveMessage(source, type, function() {
```
curioussavage: @dave I saw that listener was not defined but does `receiveMessageOnce` need to return anything?

curioussavage: also, random thought. If this is shared between codebases maybe it could be an npm module?

schwers: :+1: to npm modules idea if it's shared

assets/js/client.es6.js
```diff
@@ -279,7 +276,7 @@ function initialize(bindLinks) {
     refreshMS = Math.max(refreshMS - (1000 * 60), 0);
 
     window.setTimeout(function() {
-      refreshToken(app).then(function(){});
```
schwers: :no_bell: We shouldn't need then(...) but this refresh code should just be cleaned up in separate pr

assets/js/frames.es6.js
dwick: :+1:

assets/js/frames.es6.js
dwick: yeah, i've got a repo for it but havent gotten around to modularizing it.

assets/js/frames.es6.js
```diff
@@ -183,19 +182,18 @@ const frames = {
...
       callback && callback.apply(this, arguments);
-
-      listener.off();
     }, context);
+    listener.off();
```
dwick: this should stay inside the function or else the callback will never be caled

assets/js/gtm.es6.js
```diff
@@ -1,25 +1,23 @@
...
-let jail = document.getElementById('gtm-jail');
-
-let gtm = {
-
-  trigger: function(eventName, payload) {
-    if (payload) {
-      this.set(payload);
+export default {
+  trigger(eventName, payload) {
+    const jailEl = document.getElementById('gtm-jail');
```
dwick: :no_bell: I think this can be outside of the function.  it's either defined immediately or ever.

dwick:
:nail_care:

dwick:
thanks for fixing this

assets/js/client.es6.js
```diff
@@ -12,16 +12,16 @@ function onError(message, url, line, column) {
...
-    hivemind: window.bootstrap && window.bootstrap.config ? window.bootstrap.config.statsURL : undefined,
+    hivemind: window.bootstrap && window.bootstrap.config ?
+      window.bootstrap.config.statsURL : undefined,
   });
 }
 
 // Register as early as possible
 window.onerror = onError;
 
-import React from 'react';
```
schwers: :no_bell: wat :water_buffalo:

assets/js/gtm.es6.js
schwers: Shouldn't we be waiting until onready or etc has happened before looking it up though? `getElementyById` should be fast. Alternatively we could cache this lookup the first time the function is called.

assets/js/frames.es6.js
schwers: @dwick been talking to @curioussavage about this. It looks like the current code shouldn't be working since listener isn't in scope when the callback is defined. Should it be this?

```
receiveMessageOnce(source, type, callback, context) {
  const listener = frames.receiveMessage(source, type, function() {
    callback && callback.apply(this, arguments);
    listener.off();
  }, context);

  return listener;
},
```

assets/js/gtm.es6.js
curioussavage: My worry was that since it was being run when the bundle is loaded it could run before that default element renders. Even if that isn't the case now future changes to how the app loads/renders could affect that.

I was using a function that would cache it but ryan pointed out that getElementById should be really fast anyways.

assets/js/client.es6.js
curioussavage: yep. React isn't used here.

assets/js/frames.es6.js
```diff
@@ -53,11 +52,11 @@ function _addEventListener(type, handler, useCapture) {
   }
 }
 
-function _removeEventListener(type, handler, useCapture) {
+function _removeEventListener(type, handler) {
   if (global.removeEventListener) {
     global.removeEventListener(type, handler);
   } else if (global.detachEvent) {
-    global.attachEvent(`on${type}`, handler);
+    global.detachEvent(`on${type}`, handler);
```
curioussavage: pretty sure this was another typo. but for sanity @dwick ^

curioussavage:
:haircut:  comentz if any please

assets/js/frames.es6.js
schwers: @curioussavage The docstring makes me thing `listener` needs to be returned

schwers: yey

ajacksified:
:fish:

PR #459, opened 2016-01-21
Set background to key_color on community header icons

schwers:
Some of the new subreddit icons rely on the transparency of icon images to render properly. Eg r/apple

In this patch:
image

Versus prod:
image

:eyeglasses: @ajacksified or @curioussavage or @nramadas


ajacksified:
:fish:

curioussavage:
:shipit: :fish:

PR #460, opened 2016-01-21
Server should always send google metadata

ajacksified:
Cloudflare returns cached data with no Google metadata often. So always
render it.

:eyeglasses: @xilvar and @umbrae and maybe @schwers


schwers:
LGTM :fish:

PR #461, opened 2016-01-21
Fix googlebot's mobile js crawler classification

schwers:
:eyeglasses: @ajacksified

### DIFF
--- a/src/lib/errorLog.es6.js
+++ b/src/lib/errorLog.es6.js
@@ -3,6 +3,10 @@ import superagent from 'superagent';
 function simpleUA(agent) {
   if (/server/i.test(agent)) { return 'server'; }
 
+  // Googlebot does silly things like tell us it's iPhone, check first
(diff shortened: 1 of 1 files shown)


src/lib/errorLog.es6.js
```diff
@@ -3,6 +3,10 @@ import superagent from 'superagent';
 function simpleUA(agent) {
   if (/server/i.test(agent)) { return 'server'; }
 
+  // Googlebot does silly things like tell us it's iPhone, check first
+  // see https://googlewebmastercentral.blogspot.com/2014/01/a-new-googlebot-user-agent-for-crawling.html
+  if (/Googlebot/i.test(agent)) { return 'googlebot-js-client'; }
```
ajacksified: you upper cased it, but you're using case-insensitivity :water_buffalo:

ajacksified:
:fish:

src/lib/errorLog.es6.js
schwers: ya I wanted to match the user-agent from the blogpost -- but case-insensitive just to be safe :)

PR #462, opened 2016-01-22
Make login/register use real links and redirect back properly.

schwers:
This changes login/register to be regular links. They no longer require
setting `data-no-route={ true }` or setting the originalUrl. Cleans up
`app.redirect` in `client.es6.js` and removes duplicated logic in the
click handlers to make this possible.

Adds a `REGISTER_PATH` that automatically is `/register` for apps that
use `/login`.

The app links to `LOGIN_PATH` when there’s a good  chance the user is
intending to log in to an existing account. eg if they click ‘log in’
on the top of the page or user menu. Otherwise when an action that
requires having an account is performed `app.needsToLogInUser` will
properly redirect to `REGISTER_PATH`

For testing: Should be tested in staging to make sure the non-development `originalUrl` validation is working as intended

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -85,17 +85,6 @@ function onLoad(fn) {
   }
 }
 
-function redirect(status, path) {
-  if ((typeof status === 'string') && !path) {
-    path = status;
-  }
-
-  if (path.indexOf('/login') > -1 || path.indexOf('/register') > -1 ) {
-    window.location = path;
-  } else {
-    this.redirect(path);
-  }
-}
 
 // A few es5 sanity checks
 if (!Object.create || !Array.prototype.map || !Object.freeze) {
@@ -121,7 +110,7 @@ function modifyContext (ctx) {
     showOver18Interstitial: (cookies.get('over18') || 'false').toString() === 'false',
     showEUCookieMessage: (EUCookie < constants.EU_COOKIE_HIDE_AFTER_VIEWS) && isEUCountry,
     showGlobalMessage: cookies.get((app.config.globalMessage || {}).key) === undefined,
-    redirect: redirect.bind(app),
+    redirect: app.redirect,
     env: 'CLIENT',
     winWidth: window.innerWidth,
     // pick up notifications off of the base for the intial load, since the
@@ -313,15 +302,39 @@ function initialize(bindLinks) {
     }
   };
 
-  app.redirect = function(url) {
-    app.pushState(null, null, url);
+  app.redirect = function redirect(status, path) {
+    if ((typeof status === 'string') && !path) {
+      path = status;
+    }
+
+    if (path.indexOf(config.loginPath) === 0 ||
+        path.indexOf(config.registerPath) === 0) {
+      window.location = path;
+      return;
+    }
+
+    app.pushState(null, null, path);
 
     // Set to the browser's interpretation of the current name (to make
     // relative paths easier), and send in the old url.
     render(app, app.fullPathName(), false, modifyContext).then(function(props) {
       setTitle(props);
+      postRender(path);
     });
-  };
+  }.bind(app);
+
+  // Redirects to the proper register path if the user isn't logged in.
+  //
+  // Return truthy because that's easier for consumers to check.
+  // Doesn't throw an exception for control flow out of the caller.
+  // This would make the consuming code a oneliner but it's not really
+  // an error. Plus it will simplfy things for the client-side errors.
+  app.needsToLogInUser = function() {
+    if (!this.getState('token')) {
+      this.redirect(this.config.registerPath);
+      return true;
+    }
+  }.bind(app);
 
   app.forceRender = function (view, props) {
     ReactDOM.render(view(props), app.config.mountPoint);
@@ -420,11 +433,9 @@ function initialize(bindLinks) {
         a.href = currentUrl;
         referrer = a.href;
 
-        app.pushState(null, null, href);
-
-        // Set to the browser's interpretation of the current name (to make
-        // relative paths easier), and send in the old url.
-        render(app, app.fullPathName(), false, modifyContext).then(postRender(href));
+        // Let app.redirect do the heavy lifting. It has the the fancy
+        // check for login / register
+        app.redirect(href);
       });
 
       window.addEventListener('popstate', function(e) {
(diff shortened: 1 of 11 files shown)


schwers:
:eyeglasses: @ajacksified @curioussavage @nramadas

assets/js/client.es6.js
```diff
@@ -289,6 +278,7 @@ function initialize(bindLinks) {
     window.location = '/oauth2/login';
   });
 
+
```
ajacksified: dual newlines :scream:

assets/js/client.es6.js
```diff
@@ -301,15 +291,38 @@ function initialize(bindLinks) {
     }
   };
 
-  app.redirect = function(url) {
-    app.pushState(null, null, url);
+  app.redirect = function(status, path) {
```
ajacksified: :no_bell: `... = function redirect` if you want nice traces

assets/js/client.es6.js
```diff
@@ -301,15 +291,38 @@ function initialize(bindLinks) {
...
+    if ((typeof status === 'string') && !path) {
+      path = status;
+    }
+
+    if (path.indexOf(config.loginPath) > -1 || path.indexOf(config.registerPath) > -1) {
```
ajacksified: An equality check would be safer; this would match `/r/login`

src/config.es6.js
```diff
@@ -20,6 +20,9 @@ if (globalMessage) {
 }
 
 function config() {
+  const loginPath = process.env.LOGIN_PATH || '/oauth2/login';
+  const registerPath = loginPath === '/login' ? '/register' : loginPath;
```
ajacksified: shouldn't this read from the `REGISTER_PATH` env var?

src/routes.jsx
```diff
@@ -598,14 +599,40 @@ function routes(app) {
...
 
+  function loginRegisterOriginalURL(ctx) {
+    let { originalUrl } = ctx.query;
+
+    if (!originalUrl && ctx.headers && ctx.headers.referer) {
+      const parsedReferrer = url.parse(ctx.headers.referer);
+
+      // Be extra careful about using referrer's only from the same origin
+      // when this is in prod
+      if (config.env === 'development') {
```
ajacksified: why not always?

src/routes.jsx
```diff
@@ -598,14 +599,40 @@ function routes(app) {
...
+          originalUrl = parsedReferrer.path;
+        }
+      }
+    }
+
+    // Make sure originalUrl only points back to mobile web
+    if (originalUrl && originalUrl.indexOf('//') === -1 && originalUrl[0] === '/') {
+      return originalUrl;
+    }
+  }
```
ajacksified: should this return a sane default (`/` maybe) in case these checks fail?

src/routes.jsx
```diff
@@ -598,14 +599,40 @@ function routes(app) {
     this.body = makeBody(FAQPage);
   });
 
+  function loginRegisterOriginalURL(ctx) {
```
ajacksified: :no_bell: instead of passing in ctx, you could pass in query and headers to make it clearer what you're using.

ajacksified:
:nail_care:

assets/js/client.es6.js
schwers: woops

src/config.es6.js
schwers: Originally I thought it would just be config'd entirely. Then I realized adding the logic here would make it not need any setup to deploy (? I assume so) or any config changes locally. I could add reading from env var too , thoughts?

src/routes.jsx
schwers: TL;DR seemed less invasive to how we develop  and config locally. 

I usually develop against `<machine-name>.local:4444` as that url works both on my iPhone and Mac without having to find my IP. Plus android phones in local dev are requesting `<instance-ip-address>:4444`. One option is finding the IP in `config.es6.js` and making that the new default for `origin` when it's not passed in via the env variable. But then to log in locally we'd lose `localhost:4444` and `<machine-name>.local:4444`, unless I'm missing something. Thoughts?

src/routes.jsx
schwers: :+1:

src/routes.jsx
schwers: `getLoginRedirect` in `oauth.es6.js` already handles going to `/` when originalUrl is undefined. Plus `login.jsx` and `register.jsx` would add the query param and hidden inputs for `/` that aren't really needed. On the other hand, I'm happy to make it explicit if you think that's clearer, better, etc

src/config.es6.js
schwers: Updated to remove the entry in config.md. This way its more clear that `registerPath` is derived from config, but not configurable via an env variable.

schwers:
:haircut: Now includes tests

curioussavage:
:+1:  for tests!

src/routes.jsx
```diff
@@ -239,6 +240,24 @@ function * globalMessage(next) {
   yield next;
 }
 
+
```
ajacksified: nuke newline :scream:

test/index.js
```diff
@@ -25,3 +25,6 @@ require('./views/layouts/BodyLayout');
 // pages
 require('./views/pages/Index');
 require('./views/pages/userSaved');
+
+// functions from routes
+require('./routes/loginRegisterOriginalUrl.jsx');
```
ajacksified: no need for `.jsx`

test/routes/loginRegisterOriginalUrl.jsx
```diff
@@ -0,0 +1,66 @@
+import chai from 'chai';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+
+var expect = chai.expect;
```
ajacksified: `const`

schwers:
:haircut:

assets/js/client.es6.js
```diff
@@ -305,15 +294,39 @@ function initialize(bindLinks) {
     }
   };
 
-  app.redirect = function(url) {
-    app.pushState(null, null, url);
+  app.redirect = function redirect(status, path) {
+    if ((typeof status === 'string') && !path) {
```
ajacksified: `'string'` should be a const like `const STRING = 'string'`

ajacksified:
:fish:

PR #464, opened 2016-01-22
GTM add null check in trigger and set methods

curioussavage:
This was causing an error locally when the config variables
aren't set and there is no 'jail' element. The two methods
try to access a property on the element.

also fully es6ified it. and moved `getElementbyId` into the functions so It doesn't run when the file is loaded. 

@dwick @schwers

### DIFF
--- a/assets/js/gtm.es6.js
+++ b/assets/js/gtm.es6.js
@@ -1,25 +1,23 @@
-'use strict';
-
 import frames from './frames';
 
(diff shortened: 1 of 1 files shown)


curioussavage:
:eyeglasses: @dwick and @schwers or @ajacksified or @nramadas

curioussavage:
moved to https://github.com/reddit/reddit-mobile/pull/458 since that pr changed modified the same file.

PR #466, opened 2016-01-25
Super simple client error logging

ajacksified:
:eyeglasses: @schwers 

This adds the ability for the client to send error logs to the server, which will be `console.log`. You can then grep the server logs for output to see what's going on on the client. Includes stack traces if available.

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -1,27 +1,35 @@
 import 'babel/polyfill';
 
-import '../../src/lib/dnt';
+import superagent from 'superagent';
 
 import errorLog from '../../src/lib/errorLog';
 
+const POST_ERROR_URL = '/error';
 
 function onError(message, url, line, column) {
-  errorLog({
+  const details = {
     userAgent: window.navigator.userAgent,
     message,
     url,
     line,
     column,
-    requestUrl: window.location.toString()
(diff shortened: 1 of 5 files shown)


src/lib/errorLog.es6.js
```diff
@@ -49,7 +49,6 @@ function formatLog(details) {
 
 function errorLog(details, errorEndpoints, config={}) {
   let formattedLog = formatLog(details);
-  console.log(formattedLog);
```
schwers: We should probably keep this log

assets/js/client.es6.js
```diff
@@ -56,6 +62,8 @@ var $body = document.body || document.getElementsByTagName('body')[0];
 var $head = document.head || document.getElementsByTagName('head')[0];
 
 var config = defaultConfig();
+// the client should post errors to the /error endpoint.
+config.postErrorURL = POST_ERROR_URL;
```
schwers: Should this just be in `config`? Then `routes.es6.js` could use it too

schwers:
:nail_care:

schwers:
:fish:

ajacksified:
:fish:

ajacksified:
:eyeglasses: @ajacksified

PR #470, opened 2016-01-27
Login/Register: Fix bug with timeout

curioussavage:
When logging in times out there was no response object. This
Fixes a TypeError that was being thrown because of attempts to
look up properties on it.

### DIFF
--- a/src/server/oauth.es6.js
+++ b/src/server/oauth.es6.js
@@ -403,7 +403,7 @@ var oauthRoutes = function(app) {
         .type('form')
         .send(data)
         .timeout(constants.DEFAULT_API_TIMEOUT)
-        .end((err, res) => {
+        .end((err, res = {}) => {
           if (err || !res.ok) {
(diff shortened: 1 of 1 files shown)


curioussavage:
:eyeglasses: @ajacksified

curioussavage:
update: changed to default prop to be more concise.

src/server/oauth.es6.js
```diff
@@ -403,7 +403,7 @@ var oauthRoutes = function(app) {
         .type('form')
         .send(data)
         .timeout(constants.DEFAULT_API_TIMEOUT)
-        .end((err, res) => {
+        .end((err, res = {}) => {
```
curioussavage: maybe I should test some more because this seemed to work, but I thought arrow functions can't have default args???

ajacksified: Arrow functions can use defaults :+1:

ajacksified:
:fish:

PR #471, opened 2016-01-27
Bug Fix: prevent render on first popstate in safari

curioussavage:
Safari triggers a popstate event on initial render.
On the login route this was causing it to render the
page again with the wrong referer.


curioussavage:
:eyeglasses: @schwers or @ajacksified

schwers:
:fish: if the travis works out

ajacksified:
:fish:

schwers:
:tropical_fish:

PR #472, opened 2016-01-27
Only run `finish` setstate on the client

ajacksified:
Avoid extra synchronous setstates on the server by only setting state on
the client once all of the data loads.

:eyeglasses: @schwers


schwers:
:fish:

PR #473, opened 2016-01-28
Docker

ajacksified:
This adds a dockerfile (and, as an aside, improves index.js startup if you're not using console.)

Tested with dokku on digitalocean; docs to follow in the wiki.

:eyeglasses: @curioussavage

### DIFF
--- a/index.js
+++ b/index.js
@@ -90,7 +90,7 @@ require('./version');
 
 // Require in the express server.
 const Server = require('./src/server');
-const Console = require('./src/server/console');
+let Console;
 
 const cluster = require('cluster');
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -120,6 +120,7 @@ if (cluster.isMaster) {
 
   // If we used `node index.js --console`, instantiate a dashboard.
   if (process.argv[2] && process.argv[2] === '--console') {
+    Console = require('./src/server/console');
```
curioussavage: :+1: always a fan of startup time going down

curioussavage:
lgtm :fish:

PR #474, opened 2016-01-28
Fix permalink's not working in comments

nramadas:
Fixes the issue with permalink's being undefined in the dropdowns


nramadas:
:eyeglasses: @schwers @ajacksified @curioussavage

schwers:
:fish:

PR #476, opened 2016-02-01
Make a few changes to comments

nramadas:
This addressed a few things, but most importantly, it should speed up the rendering of comments a little bit.

1) Remove the `<CommentBody />` component and move its rendering into its parent
2) Render the `<CommentReplyForm />` component conditionally
3) Add a border around flair
4) Make all the callbacks required


nramadas:

src/views/components/comment/Comment.jsx
```diff
@@ -414,9 +412,10 @@ export default class Comment extends BaseComponent {
     if (collapsed) { cls += ' m-hidden'; }
     
     return (
```
curioussavage: Would it help perf a little on comments page to just return null? Seems to me like it would save a bunch of calls to `mobilify` and `dangerouslySetInnerHTML`

nramadas: I don't know if it'll help with the mount / first-render since there are no collapsed comments. Maybe in the future when we cache collapsed comment state, not rendering the body could help for the first render.

For subsequent re-renders, I guess it's a question of un-mounting/re-mounting versus updating and which one is cheaper.

ajacksified: We'll also need to check the strategy on `renderReplies`, once that's the case.

ajacksified:
:fish:

PR #477, opened 2016-02-02
Restyle register

curioussavage:
work in progress...


curioussavage:
:eyeglasses: @ajacksified @nramadas

curioussavage:
Would definitely recommend going through by commit. When I was working on the token refresh client side I felt like the giant client file made it hard to follow along with so I broke that out too. Mostly shuffling code around, with a few changes.

Another change is the old modal component is stateless. 

Its on staging and needs a little polishing still. Just wanted to get reviews started now since the diff is huge.

ajacksified:
* when I first log in, when I click to upvote something, it takes me to register
* when I hit 'enter' / submit from the password field, it runs the "view password" action instead of submitting the form

:nail_care:

assets/less/components/LoginPage.less
```diff
@@ -1,4 +1,66 @@
...
+    margin-right: -5px;
+  }
+}
+
+.login-register-link {
+  text-align: center;
+  padding-top: 5px;
+}
+
+.login-form {
```
ajacksified: kill this class if it's empty

ajacksified:
Erg, sorry, added comments on commits rather than on the PR. Be nice if github automatically resolved them.

curioussavage:
img_0797
img_0796
img_0795

curioussavage:
darn, I rebased the big commit with the client stuff into the login stuff. going to try to fix that

curioussavage:
issue with voting and other actions after login should be resolved.

curioussavage:
:haircut:  I am still converting the other less files to the new style but the major issues were resolved

curioussavage:
CSS is looking much better now.

curioussavage:
paging @librarianavenger

curioussavage:
:eyeglasses: @schwers could you look at that last commit

assets/js/lib.es6.js
```diff
@@ -0,0 +1,229 @@
...
+  const totalScroll = top + el.offsetHeight;
+  if (totalScroll === el.scrollHeight) {
+    el.scrollTop = top - 1;
+    return false;
+  }
+
+  return true;
+}
+
+export const findLinkParent = function findLinkParent(el) {
```
nramadas: out of curiosity, why declare these as named function expressions? wouldn't `export function findLinkParent(el) {...}` work just as well?

src/views/components/forgotpassword.jsx
```diff
@@ -0,0 +1,152 @@
...
+      this.setState({
+        error: ERRORS[e.name] ||
+               ERRORS[e.status] ||
+               ERRORS.DEFAULT,
+      });
+    }
+  }
+
+  async makeRequest(uri, postData) {
+    return await new Promise(async function(resolve, reject) {
```
nramadas: async functions already return promises, so I don't think you need to wrap it in a promise. you could turn it into an IIFE instead:

```
return await (async function() {...})();

File: assets/js/lib.es6.js [hunk 4]
```diff
@@ -0,0 +1,229 @@
+import superagent from 'superagent';
```
schwers: :no_bell: Would you be opposed to renaming this to something like `clientUtils` or `clientLib` to disambiguate it from `lib/` ?

src/views/components/forgotpassword.jsx
```diff
@@ -0,0 +1,152 @@
...
+      } catch (e) {
+        if (e.timeout) {
+          e.name = 504;
+        }
+        return reject(e);
+      }
+    });
+  }
+
+  updateName(e, val) {
```
nramadas: it looks like `updateName` is either being fired as a callback or by `clear`. Since `val` is always either undefined or `''`, I think we can drop it and just say:
```
const name = e ? e.target.value : '';
```

src/views/components/forgotpassword.jsx
```diff
@@ -0,0 +1,152 @@
...
+      </button>
+    );
+  }
+
+  clear() {
+    this.updateName(null, '');
+  }
+
+  render() {
+    const name = this.state.name || '';
```
nramadas: why not default this in the constructor?

src/views/components/snooiconheader.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+function SnooIconHeader(props) {
```
nramadas: can we add some type checks? I know there are only two props, but for consistency, we should always add these checks.

src/views/pages/login.jsx
```diff
@@ -1,15 +1,69 @@
...
+};
+
+const placeholderText = {
+  register: {
+    username: 'Choose a unique username',
+    password: 'Choose a unique password',
+    email: 'Your email',
+  },
+  login: {
+    username: 'username',
```
nramadas: nitpick, but should these be capitalized to match the copy in the register form?

src/views/pages/login.jsx
```diff
@@ -1,15 +1,69 @@
...
+    { 'Privacy Policy ' }
+  </a>
+);
+const content = (
+  <a
+    href='/help/contentpolicy/'
+    className='text-link'
+    target='_blank'
+  >
+    { 'Content Policy' }
```
nramadas: Couldn't you drop the curly brackets?

src/views/pages/login.jsx
```diff
@@ -1,15 +1,69 @@
...
-  500: 'There was a problem with the server',
-  400: 'Wrong password',
-  default: 'An error occured.',
+  504: 'Sorry, it took too long for the server to respond',
+  500: 'Sorry, something has gone wrong with the server',
+  WRONG_PASSWORD: 'Sorry, that’s not the right password',
+  default: 'Sorry, Something has gone wrong and we\'re not sure what',
+};
+
+const PasswordFieldTypes = {
```
nramadas: :no_bell: just a personal preference of mine, I like naming true constants with all uppercase, like you did for `ERROR_MESSAGES`. So,
```
const PASSWORD_FIELD_TYPES = {
    PASSWORD: 'password',
    TEXT: 'text',
};
```

src/views/components/forgotpassword.jsx
schwers: @nramadas Does superagent or babel do the promise wrapping now automatically? There's a plugin available to add real promises with .then and in most of the code we wrap it in Promises ourselves. If we can just await it across the board -- would be a nice clean up patch

nramadas: @schwers I believe babel (through the regenerator runtime) adds the promises in for us: https://github.com/facebook/regenerator/blob/master/runtime.js#L136

The annoying part is that this isn't revealed in the transpiled code. I believe the promise is generated at runtime.

ajacksified: Or, we switch to fetch \o/

src/views/pages/login.jsx
ajacksified: Yeah, I thought it was in the style guide (but I guess not), but file-global constants should be uppercased.

curioussavage: yep, good catch. something I forgot to clean up

src/views/pages/login.jsx
curioussavage: not without losing the spaces. I would prefer this to adding &nbsp;

src/views/components/forgotpassword.jsx
curioussavage: good catch, another thing that wasn't cleaned up.

assets/js/lib.es6.js
curioussavage: Thanks for pointing this one out. For some reason I had in my head the idea that you had to assign something to a variable to use named exports

src/views/components/forgotpassword.jsx
curioussavage: fetch would be better. If we rely on babel doing that at runtime then its more work if we wan't to start running without babel on the server.

curioussavage:
:haircut: all

src/views/pages/login.jsx
```diff
@@ -17,82 +72,379 @@ class LoginPage extends BasePage {
...
+    //   // set global error
+    // }
+
+    const message = e.message || ERROR_MESSAGES[e.error];
+    this.setState({ error: newError, errorMessage: message });
+  }
+
+  makeRequest(endpoint, data) {
+    return new Promise((resolve, reject) => {
+      superagent
```
nramadas: Does superagent return promises? If so, we can avoid wrapping the api call here in another promise.

ajacksified: It does not.

curioussavage: although we could start just wrapping this stuff in async calls since apparently regenerator will wrap it with a generator.

schwers: We should be careful with testing that. I'm curious if it's smart enough to work with superagent specifically. I'm guessing they assume that there's a `.then()` method of what you're awaiting and superagent has a [faux-promise-then method](https://github.com/visionmedia/superagent/blob/master/lib/request-base.js#L56). I'm of the opinion that it's okay to move that into a separate 'cleanup-superagent-calls' patch OR wait till the new `fetch` work happens

curioussavage:
:eyeglasses: @spez because this is a visual change.

this is live on https://www.cloudflarestagingformobilereddit.com/ right now

src/views/pages/login.jsx
ajacksified: :+1: to waiting

spez:
:fish:

curioussavage:
:haircut:

assets/less/components/LoginPage.less
```diff
@@ -1,4 +1,71 @@
...
+  &__wrapper {
+    background-color: @off-white;
+
+    .row {
+      margin-right: -5px;
+    }
+  }
+
+  &__register-link {
+    color: @alien-blue-text;
```
ajacksified: Are all links supposed to be this color by default now? Can we just apply this to the base `a` color?

assets/less/components/LoginPage.less
```diff
@@ -1,4 +1,71 @@
...
+    text-align: center;
+    padding-top: 5px;
+  }
+
+  &__forgot-link {
+    color: @alien-blue-text;
+    margin-top: 5px;
+  }
+
+  &__submit-btn {
```
ajacksified: The whole squarebutton class seems needlessly overcomplex, but I guess now isn't the time to fix it. But... wut. :no_bell:

assets/less/components/LoginPage.less
```diff
@@ -1,4 +1,71 @@
...
+
+    [class^="icon-"] {
+      font-size: 24px;
+    }
+
+    .icon-x {
+      color: @red;
+    }
+
+    &.blue {
```
ajacksified: `.blue [class^=icon]` is sufficient - no need to nest with `&` here. Move this into `text.less`.

assets/less/components/Modal.less
```diff
@@ -1,8 +1,19 @@
...
 	margin: 130px auto 30px;
-}
\ No newline at end of file
+}
+
+.modal-content {
+  border-radius: 0;
+}
+
+.Modal__close {
```
ajacksified: the mixing of cases is literally making me cry :no_bell:

assets/less/components/minimalinput.less
```diff
@@ -0,0 +1,42 @@
+.inputMinimal {
```
ajacksified: the file is `minimalInput`, the class is `inputMinimal`. I'd suggest changing the filename

assets/less/components/snooiconheader.less
```diff
@@ -0,0 +1,24 @@
+.SnooIconHeader-wrapper {
```
ajacksified: If we're moving towards camelcase for new css, can we camelcase all of this without breaking anything existing?

src/server/index.jsx
```diff
@@ -28,7 +28,7 @@ import setLoggedOutCookies from '../lib/loid';
 
 import Config from '../config';
 
-const ignoreCSRF = ['/timings', '/error'];
+const ignoreCSRF = ['/timings', '/error', '/login', '/register'];
```
ajacksified: shouldn't have to ignore it now that we're not cdn-cached.

kjoconnor: I wouldn't assume we'll always be running uncached.

ajacksified: If we're implementing logged-out oauth, we're going to have to.

ajacksified:
Three minor things; :nail_care:

assets/less/components/LoginPage.less
curioussavage: erica and I talked about this, Ill check again and if that was the case will definitely do that.

assets/less/components/LoginPage.less
schwers: The color modifiers in `icon-font.less` should take care of this for you. You should be able to remove this rule altogether and add the class `blue` to the icon itself and get the same effect.

assets/less/components/snooiconheader.less
curioussavage: did you mean camelcase for the filename or were you meaning to say use __ for bem syntax?

ajacksified: Yes, and also snooIconHeader

ajacksified: err, maybe we're not. I'm not sure. I thought we were moving to camelcase, but @nramadas just added a bunch of stuff with capitals in https://github.com/reddit/reddit-mobile/pull/506/files#diff-7ad847eaa65fdf0831fdc66da8abb916R1

assets/less/components/minimalinput.less
ajacksified: and we should figure out capitalization

curioussavage:
:haircut: addressed issues, also tweaked some css a little to increase hit area on all links/ close buttons

ajacksified:
:fish:

PR #478, opened 2016-02-02
Unify several dropdowns

nramadas:
This patch fixes up the differing dropdown styles in several areas.

NOTE: this does NOT touch `<ListingDropdown />`s. Schwers is going to tackle that in his listings patch.

Some screenshots!

**Listings**
screen shot 2016-02-02 at 2 57 11 pm

screen shot 2016-02-02 at 2 57 38 pm

**Comments**
screen shot 2016-02-02 at 2 58 22 pm

screen shot 2016-02-02 at 2 58 46 pm

**User activity**
screen shot 2016-02-02 at 2 59 16 pm

screen shot 2016-02-02 at 2 59 41 pm

screen shot 2016-02-02 at 2 59 57 pm

**Search results**
screen shot 2016-02-02 at 3 00 56 pm

screen shot 2016-02-02 at 3 01 15 pm

screen shot 2016-02-02 at 3 01 30 pm


nramadas:

ajacksified:
Should 'sorted by' be gone in search results, and the time dropdown be added to home/subreddit view? (Additionally, I _think_ time should only show up for `top`.)

assets/less/components/SortSelector.less
```diff
@@ -0,0 +1,104 @@
...
+  }
+  
+  &__currentIcon,
+  &__dropdownIcon {
+    font-size: 24px;
+    color: @grey-text;
+    text-align: center;
+    margin-left: -@half-grid-size;
+    
+    // I wish the icons were actually normalized...
```
ajacksified: :(

assets/less/components/formElements/TextField.less
```diff
@@ -9,4 +9,6 @@
   box-shadow: none;
   -webkit-appearance: none;
   border-radius: 0;
+  font-size: 13px;
```
ajacksified: Variablize this into `@textFieldFontSize` or something?

src/views/components/LinkTools.jsx
```diff
@@ -117,4 +124,26 @@ export default class LinkTools extends React.Component {
...
+          sortValue={ sort }
+          sortOptions={ [
+            SORTS.CONFIDENCE,
+            SORTS.HOT,
+            SORTS.NEW,
+            SORTS.TOP,
+            SORTS.QA,
+          ] }
+          onSortChange={ this.props.onSortChange }
+          title="Sort comments by:"
```
ajacksified: single quotes. we really should get that into the linter

src/views/components/SortSelector.jsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+  
+  renderCurrentSort() {
+    const { sortValue } = this.props;
+    const { text, icon } = SORT_VALUES_MAP[sortValue];
+    
+    return (
+      <div className='SortSelector__currentSort'>
+        <div className='SortSelector__currentSortContent' onClick={ this.toggleDropdown }>
+          { icon ? this.renderCurrentIcon() : null }
```
ajacksified: since `renderCurrentIcon` gets the value of icon from props anyway, it should return undefined if there's no icon - and then you can get rid of the ternary.

src/views/components/SortSelector.jsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+    
+    if (!icon) {
+      iconCls += ' m-no-icon';
+    }
+    
+    // define callbacks
+    const handleClick = (function() {
+      this.handleSortClick(sortName);
+    }).bind(this);
```
ajacksified: `handleClick = this.handleSortClick.bind(this, sortName)`

nramadas:
I can remove the sorted by. I don't think we need that text anymore.

As for time just for top, our current product has both visible at all times. For this patch, I didn't want to change the product around too much (the patch is already close to 1000 additions), so I just substituted the old dropdowns with new ones.

We definitely have mocks that would unify both these dropdowns, and I'd be happy to tackle that next.

src/views/components/TopSubnav.jsx
```diff
@@ -47,7 +59,7 @@ function TopSubnav (props) {
 
   return (
     <div className='TopSubnav'>
-      { sort }
+      { showSort ? renderSortSelector(props.sort, props.app, props.ctx.url) : null }
```
ajacksified: `renderSortSelector` should return early, and kill the ternary.

assets/less/components/formElements/TextField.less
nramadas: good idea.

src/views/components/LinkTools.jsx
nramadas: huh, the linter usually catches that.

src/views/components/UserActivitySubnav.jsx
```diff
@@ -1,171 +1,175 @@
...
+        </DropdownContent>
+      </DropdownController>
+    );
+  }
+  
+  renderActivityRow(sortName) {
+    // define callback
+    const handleSortClick = (function() {
+      this.handleActivitySortClick(sortName);
+    }).bind(this);
```
ajacksified: (same binding comment as above)

src/views/components/comment/CommentDropdownContent.jsx
```diff
@@ -84,48 +86,49 @@ export default class CommentDropdownContent extends React.Component {
...
       <div className='CommentDropdownContent'>
-        { userOwned ? this.renderEdit() : null }
-        { userOwned ? this.renderDelete() : null }
-        { !userOwned && false /* disabled for now */ ? this.renderGold() : null }
-        { this.renderShare() }
-        { userLoggedIn ? this.renderSave() : null }
-        { this.renderProfile() }
-        { userLoggedIn && !userOwned ? this.renderReport() : null }
+        <DropdownContent>
+          { userOwned ? this.renderEdit() : null }
```
ajacksified: (same kill-the-ternary comments as above; we should just rely on `renderX` functions to return early.)

ajacksified:
@nramadas sounds good on the not-changing-how-time-works.

ajacksified:
A few nits (looks like bad linter rules?), but overall looks great!

ajacksified:
oh, forgot to add :nail_care:

nramadas:
:haircut:

src/views/components/SortSelector.jsx
ajacksified: Discussed in-person; decided that since `renderCurrentIcon` does exactly what it says on the tin, this is okay. It's semantically equivalent to setting a variable to `renderCurrentIcon` in an if statement anyways and using the variable.

I was worried that `renderCurrentIcon` might throw unexpected errors if `icon` is undefined, but that should be covered by proptypes in all practical cases.

src/views/components/UserActivitySubnav.jsx
```diff
@@ -1,171 +1,173 @@
...
+        <DropdownContent>
+          { Object.keys(activityMap).map(this.renderActivityRow) }
+        </DropdownContent>
+      </DropdownController>
+    );
+  }
+  
+  renderActivityRow(sortName) {
+    // define callback
+    const handleSortClick = this.handleActivitySortClick.bind(this. sortName);
```
ajacksified: typo: `this. ` instead of `this,`.

ajacksified:
:nail_care:

src/views/components/UserActivitySubnav.jsx
nramadas: asdf;lkj.

src/views/components/UserActivitySubnav.jsx
```diff
@@ -1,171 +1,173 @@
...
+        offset={ 8 }
+        onClose={ this.toggleActivitySort }
+      >
+        <DropdownContent>
+          { Object.keys(activityMap).map(this.renderActivityRow) }
+        </DropdownContent>
+      </DropdownController>
+    );
+  }
+
```
ajacksified: also, a lot of extra whitespace keeps being added (another thing we should add to the linter)

nramadas:
:haircut:

umbrae:
:eyeglasses: @tdohz real fast as to check which sorts for which views - for
ex we don't need hot sort on comments pages anymore.

tdohz:
Comments should be: Best, Top, New, Controversial, Q&A <-- this removes Old but I can't think of a good product reason to keep it. Hot we are no longer calculating for comments anyway.

Posts should be: Hot, Top, New, Controversial (I don't think we need rising, but am open to arguments that we do)

Search should be: Relevance, Hot, Top, New, Comments (also open to removing Hot, but since it's only post sort, feels more consistent to include here). 

Activity for combined posts + comments should be the same as for posts; for comments only it should be the same as comments.

ajacksified:
One thing I noticed: usernames show the `:visited` style in the top nav; not sure if this is desirable (I assume not, since it's a nav toolbar element rather than a link you might want tracked, like comments/titles.) I haven't found anything else to `:nail_care:`, so I think with that fixed we're good to ship!

:nail_care:

nramadas:
:haircut:

ajacksified:
:fish:

PR #480, opened 2016-02-04
Feet

ajacksified:
:eyeglasses: @schwers @umbrae 

This adds in feature flagging. This first pass hardcodes configuration into featureflags.es6.js, but will eventually be moved to an external, non-source-controlled configuration file in a followup PR.

It also sets up base rules for a list of configured users, whether a user is logged-in, an employee, admin, or in the beta program, and also querystring-based flags.

### DIFF
--- a/src/featureflags.es6.js
+++ b/src/featureflags.es6.js
@@ -1,0 +1,45 @@
+import Feet from 'feet';
+
+const config = {
+  beta: true,
+};
+
+const feet = new Feet(config);
+
+function extractUser(ctx) {
+  const state = ctx.state;
+  if (!state || !state.data || !state.data.user) {
+    return {};
+  }
+
+  return state.data.user;
+}
+
+feet.addRule('loggedin', function(val) {
+  return !!this.props.ctx.token === val;
+});
+
+feet.addRule('users', function(users) {
+  const user = extractUser(this);
+  return users.includes(user.name);
+});
+
+feet.addRule('employee', function(val) {
+  return extractUser(this).is_employee === val;
+});
+
+feet.addRule('admin', function(val) {
+  return extractUser(this).is_admin === val;
+});
+
+feet.addRule('beta', function(val) {
+  return extractUser(this).is_beta === val;
+});
+
+feet.addRule('url', function(query) {
+  // turns { feature_thing: true, wat: 7 } into { thing: true }
+  const parsedQuery = Feet.parseConfig(this.props.ctx.query);
+  return Object.keys(parsedQuery).includes(query);
+});
+
+export default feet;
--- a/src/views/layouts/BodyLayout.jsx
+++ b/src/views/layouts/BodyLayout.jsx
@@ -66,7 +66,7 @@ class BodyLayout extends BasePage {
           key='communitymenu'
         />
         <UserOverlayMenu {...this.props} user={ user } key='usermenu' />
-        <TopNav {...this.props} user={ user } key='topnav' />
+        <TopNav {...this.props} feature={ this.state.feature } user={ user } key='topnav' />
         <InfoBar
           messages={ messages }
           app={ app }
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "dom-serializer": "git://github.com/ajacksified/dom-serializer.git#c999b029caa6d4a9f1fbeb512b48ec7c75c1dddb",
     "es5-shim": "4.1.11",
     "event-tracker": "git://github.com/reddit/event-tracker.git#28bc2dd9ce3b80540fdf189e1b003f8264cd0d08",
-    "feet": "^1.1.0",
+    "feet": "1.2.0",
     "horse": "1.3.2",
     "horse-react": "1.1.4",
     "koa": "1.0.0",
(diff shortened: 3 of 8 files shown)


src/featureflags.es6.js
```diff
@@ -0,0 +1,42 @@
...
+  return !!this.props.ctx.token === val;
+});
+
+feet.addRule('users', function(users) {
+  const user = extractUser(this);
+  return users.includes(user.name);
+});
+
+feet.addRule('employee', function(val) {
+  return extractUser(this).is_admin === val;
```
curioussavage: should be employee

ajacksified:
:haircut:

src/featureflags.es6.js
```diff
@@ -0,0 +1,43 @@
...
+
+feet.addRule('admin', function(val) {
+  return extractUser(this).is_admin === val;
+});
+
+feet.addRule('beta', function(val) {
+  return extractUser(this).is_beta === val;
+});
+
+feet.addRule('querystring', function(query) {
```
umbrae: can we call this "url" to match how feature flagging on desktop works?

```
# On by URL, like ?feature=public_flag_name
feature_some_flag = {"url": "public_flag_name"}
```

It'd probably be confusing if those looked different.

umbrae: (we'll probably want the `feature=` bit too)

src/views/layouts/BodyLayout.jsx
```diff
@@ -66,7 +66,7 @@ class BodyLayout extends BasePage {
           key='communitymenu'
         />
         <UserOverlayMenu {...this.props} user={ user } key='usermenu' />
-        <TopNav {...this.props} user={ user } key='topnav' />
+        <TopNav {...this.props} feature={ this.state.feature } user={ user } key='topnav' />
```
umbrae: Hmm so does this mean we'll need to pass `feature` down to every component that might use it? As a relative react noob, that feels pretty nonideal to me. Is there a way to have this just generally accessible?

umbrae:
:nail_care: neat

umbrae:
(Is there a pull request or a way we're reviewing feet itself?)

src/views/layouts/BodyLayout.jsx
ajacksified: There is, once we refactor to start using react contexts.

ajacksified:
@umbrae I passed it around in slack, but otherwise, not sure what the best method of formally reviewing a library is.

ajacksified:
:haircut:

src/featureflags.es6.js
```diff
@@ -0,0 +1,45 @@
...
+
+feet.addRule('employee', function(val) {
+  return extractUser(this).is_employee === val;
+});
+
+feet.addRule('admin', function(val) {
+  return extractUser(this).is_admin === val;
+});
+
+feet.addRule('beta', function(val) {
```
schwers: In Feet.enabled, this will never get used because `beta` is a bool in the config. Is that intentional for now?

src/featureflags.es6.js
```diff
@@ -0,0 +1,45 @@
...
+  return extractUser(this).is_beta === val;
+});
+
+feet.addRule('url', function(query) {
+  // turns { feature_thing: true, wat: 7 } into { thing: true }
```
schwers: Is this idea here to allow queryParam overriding for testing and/or links that toggle features? If I understand this correctly it looks like every flag would have to manually opt-in to using the 'url' rule. Might be nice for testing flags if this was baked a little more into Feet. Something like when the Feet instance is first created with query params it overrides config so the override is persisted across state changes.

schwers:
:nail_care: Overall looks good. Just have some questions about the rules and how query params are supposed to work

src/featureflags.es6.js
ajacksified: The config sets up a _feature_ named `beta`, but does not use the rule yet - the rule will be used once we start enabling "beta" user features. Correct, it is not used right now, but is for later.

src/featureflags.es6.js
ajacksified: Yes, it is for allowing querystrings to flip on features. Flags will have to manually opt-in, which is correct. I don't think it's desirable to make every test querystring-flippable, but if that proves to be wrong, we can move that in.

ajacksified:
:haircut:

src/featureflags.es6.js
schwers: Okay. I disagree in that I think it's nice to be able to querystring a feature for debugging purposes, especially in prod. But I agree that this is fine for now

schwers:
:fish:

umbrae:
@ajacksified I think we need actual review of feet. Maybe you could just do a dummy PR that is not for merging that adds it in a nonsensical place, just for review purposes?

schwers:
@umbrae FWIW I tried to include any questions I had about Feet itself as part of my fish. I think the code looks fine for a first version, its relatively small and has tests. I'd be happy to fish it if there was another PR

umbrae:
@schwers cool I can consider that a fish if you've done a review and are fishing it here, sgtm! /cc @ajacksified

PR #481, opened 2016-02-04
Fix comments page not loading on sponsored links

nramadas:
Addresses: https://reddit.atlassian.net/browse/MOBILEWEB-378

The issue was that when there were 0 comments, `comments` was an object with other properties in it, not an empty array. So, the fix here is to check if `comments` is an array before mapping over it.

### DIFF
--- a/src/views/pages/listing.jsx
+++ b/src/views/pages/listing.jsx
@@ -21,7 +21,7 @@ class ListingPage extends BasePage {
 
   constructor(props) {
     super(props);
-    
+
     this.state = {
       ...this.state,
       editing: false,
@@ -109,7 +109,7 @@ class ListingPage extends BasePage {
       editing: !this.state.editing,
     });
   }
-  
+
   async loadMore(e) {
     e.preventDefault();
     const { app, apiOptions, sort } = this.props;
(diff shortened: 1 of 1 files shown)


src/views/pages/listing.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import remove from 'lodash/array/remove';
+import isArray from 'lodash/lang/isArray';
```
ajacksified: ES6 gives us `Array.isArray`, which we should use rather than copying this in from lodash.

src/views/pages/listing.jsx
```diff
@@ -185,7 +186,23 @@ class ListingPage extends BasePage {
     let commentsList;
     let googleCarousel;
 
-    if (comments) {
+    /*
```
ajacksified: :+1: nice comments

nramadas:
:eyeglasses: @ajacksified

src/views/pages/listing.jsx
nramadas: ah, nice. i'll switch it over

nramadas:
:haircut:

ajacksified:
:fish:

PR #483, opened 2016-02-08
Fix 'load more comments" link

nramadas:
Addresses: https://reddit.atlassian.net/browse/MOBILEWEB-393

The issue was pretty simple. When we received new data, we chose not to use it (derp). This patch fixes that by actually replacing `data` with `newData`.


ajacksified:
whoops. :fish:

PR #484, opened 2016-02-09
Upgrade from Babel 5 to Babel 6

nramadas:
Upgrades our babel from 5 to 6.

This is a WIP. Cannot ship until:
* [x] fix two lines in `snoode` w/r/t static properties (https://github.com/reddit/snoode/pull/77)
* [x] figure out how to include the optimizations we had before `['optimisation.react.inlineElements', 'optimisation.react.constantElements']`

A few things to keep in mind:

1. Babel 6 requires that we include specific rule sets for what we want transpiled, hence all the extra node modules. We define which ones are applicable in a `.babelrc` file.

2. Babel 5 had a 'compatibility-mode' of sorts with `require` statements where if a file would `export default`, a normal `require` would behave as expected. Babel 6 removes this functionality. So, now anything that is exported with `export default` has to be required with `require(<thing>).default`.

3. Babel 5 would auto-include the polyfill (I think). Babel 6 does not, so we need to explicitly include it in various places.


.babelrc
```diff
@@ -0,0 +1,12 @@
+{
+  "presets": [
+    "es2015",
+    "react"
+  ],
+  "plugins": [
+    "transform-object-rest-spread",
+    "transform-async-to-generator",
+    "transform-class-properties",
+    "syntax-trailing-function-commas"
```
ajacksified: where do we do this?

package.json
```diff
@@ -18,7 +18,15 @@
     "node": ">= 4.1.0"
   },
   "dependencies": {
-    "babel": "5.8.34",
+    "babel": "^6.5.1",
```
ajacksified: kill all the `^`s, this has burned us several times :(

test/index.js
```diff
@@ -1,9 +1,10 @@
-require('babel/register')({
+require('babel-polyfill');
```
ajacksified: in theory, with babel 6, you can pass it in as a mocha transform and nuke test/index. see https://github.com/reddit/feet/blob/master/package.json#L7 as an example

.babelrc
nramadas: In `Comments` (ln 80) for one example. It's also `stage-2`, so I thought it'd be safe to include. If you're uncomfortable w/ it, I can get rid of it.

ajacksified: I mean, I dislike it, but there's an argument to be made for it (since we use trailing commas on objects and arrays). We should probably get it into the style guide one way or another. For now, makes sense to keep it so you have to make minimal changes.

test/index.js
ajacksified: talked in person - going to leave this for now. there doesn't appear to be a way to pass babel-register options (such as extensions to register) from the .babelrc. :(

ajacksified:
:fish:

ajacksified:
Next step: once we can get on a node release with V8 4.9, we should see how many babel rules we can nuke from the server's babel register (and how much it improves performance.)

PR #486, opened 2016-02-10
Play gifs after one-tap in most cases. Support gfycat.

schwers:
I've noticed playing gif's doesn't work as expected. When you tap a gif we change the preview image to an embedly embed, but that doesn't really autoplay the gif so you have to then tap a second time. This is even worse when it's a gfycat gif. This path uses the current html5 video rendering code, and uses it when you tap a gif, with some modifications so autoplay works on iOS. 

As part of this I looked into why preview images are blurry and why iOS gifs don't play inline. Preview images are still blurry in some cases (the ones we make for gfycat tend to be lower res) and could be fixed by always using the html5 video element which fetches the first frame for display.. In practice I'm not sure that's the right approach; some _very_ brief research makes it sound like setting `preload='auto'` or `preload='none'` can still result in lots of data requests on a page full of `<video />` elements. 

iOS gif's don't play inline because Safari/Chrome haven't set a flag that would enable inline video playing. We could support it whenever they turn that on. Alternatively we could just the real gif but that uses more data and the playback controls wouldn't be nearly as good.

### DIFF
--- a/src/views/components/ListingContent.jsx
+++ b/src/views/components/ListingContent.jsx
@@ -7,13 +7,30 @@ import propTypes from '../../propTypes';
 import BaseComponent from './BaseComponent';
 
 const PropTypes = React.PropTypes;
-const _gfyRegex = /https?:\/\/(?:.+)\.gfycat.com\/(.+)\.gif/;
+const _gfyCatRegex = /^https?:\/\/(.*\.?)gfycat.com/;
+const _gfyCatURLBase = 'https://thumbs.gfycat.com';
 const _DEFAULT_ASPECT_RATIO = 16 / 9;
 
+function gfycatMP4Url(gfyCatUrl) {
+  // gif doesn't seem to be there, so the .gif replace is a safety check
+  return `${gfyCatUrl.replace(_gfyCatRegex, _gfyCatURLBase).replace(/\.gif$/, '')}-mobile.mp4`;
+}
+
 function _gifToHTML5(url) {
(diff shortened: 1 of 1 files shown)


ajacksified:
> `gif's`

schwers:
:(

schwers:

schwers:
I was testing this on Android a while ago, but don't have access to any Android devices at home. If you guys could try it on one as part of review that would be awesome.

src/views/components/ListingContent.jsx
```diff
@@ -7,13 +7,30 @@ import propTypes from '../../propTypes';
...
 
 const PropTypes = React.PropTypes;
-const _gfyRegex = /https?:\/\/(?:.+)\.gfycat.com\/(.+)\.gif/;
+const _gfyCatRegex = /^https?:\/\/(.*\.?)gfycat.com/;
+const _gfyCatURLBase = 'https://thumbs.gfycat.com';
 const _DEFAULT_ASPECT_RATIO = 16 / 9;
 
+function gfycatMP4Url(gfyCatUrl) {
+  // gif doesn't seem to be there, so the .gif replace is a safety check
+  return `${gfyCatUrl.replace(_gfyCatRegex, _gfyCatURLBase).replace(/\.gif$/, '')}-mobile.mp4`;
```
ajacksified: amazing. you have found a way around the extension issue. A++

ajacksified:
:fish: pls ship immediately

PR #487, opened 2016-02-10
Server fixes

ajacksified:
:eyeglasses: @schwers or @curioussavage

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -52,11 +52,13 @@ function buildAPIOptions(ctx, options={}) {
 
   const apiOptions = merge({
     origin: app.getConfig(apiOrigin),
-    headers: {
-      'user-agent': ctx.headers['user-agent'],
-    },
+    headers: { },
     env: ctx.env || 'SERVER',
   }, options);
+
+  if (ctx.headers['user-agent']) {
+    apiOptions.headers['user-agent'] = ctx.headers['user-agent'];
+  }
 
   if (ctx.token) {
     apiOptions.headers.Authorization = `bearer ${ctx.token}`;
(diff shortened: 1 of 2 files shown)


src/routes.jsx
```diff
@@ -52,12 +52,16 @@ function buildAPIOptions(ctx, options={}) {
...
-    },
+    headers: { },
     env: ctx.env || 'SERVER',
   }, options);
 
+  if (ctx.headers['user-agent']) {
+    apiOptions.headers['user-agent'] = ctx.headers['user-agent'];
+  }
+
+  console.log(apiOptions.headers['user-agent']);
```
schwers: is this needed for debug?

src/views/pages/index.jsx
```diff
@@ -35,7 +35,7 @@ class IndexPage extends BasePage {
     subredditName: React.PropTypes.string,
   };
 
-  static isStalePage (query, listings) {
+  static isStalePage (query, listings={}) {
```
schwers: Do you have a link that repros the undefined error? `listings` was `undefined` before so shouldn't `!headers.before` through a `can't read property 'before' of undefined` error?

ajacksified: I don't have a link - I tested it by setting `this.props.datacache.listings = undefined` in routes in the line before isStalePage is called. Oddly.. it didn't throw, but it looks like it should have. I'l just return false early if `listings` is undefined.

schwers: :+1:

schwers: That is odd though, (shrug)

schwers:
:fish:

PR #488, opened 2016-02-10
Support Giphy mp4s. Play non-html5 gifs inline

schwers:


src/views/components/ListingContent.jsx
```diff
@@ -27,6 +28,12 @@ function _gifToHTML5(url) {
     };
   }
 
+  if (url.indexOf('giphy.com') > -1) {
```
ajacksified: We should start replacing `indexOf(domain)` with checks using `url.parse` for safety.

ajacksified:
One note (from slack conversation) - :nail_care:

schwers:
Fixes gifToHTML5 to be more secure and adds tests.
Adds support for giphy gifs. 
Plays all other gifs we don't know about inline. 
Higher res gfycat previews

image

is now

image

ajacksified:
:fish: assuming green

PR #489, opened 2016-02-11
listingContent: add back check for video playing status

curioussavage:
with this removed the videos only show previews


ajacksified:
:fish: once green

PR #491, opened 2016-02-11
Google carousel: prevent undefined error

curioussavage:


ajacksified:
:fish: :roller_coaster:

PR #496, opened 2016-02-15
Logs: changes separator to |

curioussavage:
change the separator so we can parse logs easier.

The tabs get stripped out of our log files.


curioussavage:
:eyeglasses: @ajacksified

ajacksified:
:fish:

PR #497, opened 2016-02-16
Support manual multis, random, and randnsfw

schwers:
Right now manual multis (like `/r/destinythegame+halo+division` fetch
links but are never rendered by the ListingPage because there were two
different definitions of fake subs in the code. Along the way adds
support for random and randnsfw

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -32,6 +32,7 @@ import WikiPage from './views/pages/wikipage';
 import constants from './constants';
 import defaultConfig from './config';
 import { SORTS } from './sortValues';
+import isFakeSubreddit, { randomSubs } from './lib/isFakeSubreddit';
 
 const config = defaultConfig();
 
@@ -142,9 +143,7 @@ function getSubreddit(ctx) {
       metaDescription: `r/${subreddit} at reddit.com`,
     });
 
-    const fakeSubs = ['all', 'mod', 'friends', 'random', 'randnsfw', 'myrandom'];
-    if (subreddit.indexOf('+') === -1 && !fakeSubs.includes(subreddit)) {
-
+    if (!isFakeSubreddit(subreddit)) {
       const subredditOpts = buildAPIOptions(ctx, {
         id: subreddit.toLowerCase(),
       });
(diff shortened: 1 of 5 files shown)


src/routes.jsx
```diff
@@ -360,6 +359,40 @@ function routes(app) {
...
+    });
+
+    try {
+      const result = yield randomRedirectResult;
+      if (result && result.header && result.header.location) {
+        const parsedLocation = url.parse(result.header.location);
+        if (parsedLocation && parsedLocation.path) {
+          return this.redirect(parsedLocation.path);
+        }
+      }
```
ajacksified: in what cases does this break? should you redirect back to `/` with an error message or something?

src/routes.jsx
```diff
@@ -360,6 +359,40 @@ function routes(app) {
...
+
+    } catch (e) {
+      console.log(e, e.stack);
```
ajacksified: should this be `app.error` instead?

test/lib/isFakeSubreddit.es6.js
```diff
@@ -0,0 +1,32 @@
...
+describe('lib: isFakeSubreddit', () => {
+  it('is a function', () => {
+    expect(isFakeSubreddit).to.be.a('function');
+  });
+
+  it('returns false for normal subreddits', () => {
+    expect(isFakeSubreddit('askreddit')).to.be.false;
+    expect(isFakeSubreddit('reactjs')).to.be.false;
+    expect(isFakeSubreddit('highqualitygifs')).to.be.false;
+    expect(isFakeSubreddit('programming')).to.be.false;
```
ajacksified: why include four subs instead of just one? what do the latter three tests test for that the first doesn't?

test/lib/isFakeSubreddit.es6.js
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  it('returns true for manual multi reddits', () => {
+    expect(isFakeSubreddit('destinythegame+division+halo')).to.be.true;
+    expect(isFakeSubreddit('javascript+reactjs')).to.be.true;
+    expect(isFakeSubreddit('onion+nottheonion')).to.be.true;
```
ajacksified: what does this last test add?

ajacksified:
:nail_care:

src/routes.jsx
schwers: It will break if the api errors or times out. I thought an error page was better than redirecting back to '/'; if you're hitting `/random` you probably want to see something new and different than your frontpage so giving them the error with a reload button seemed better

src/routes.jsx
schwers: This would be an api error, I thought we console logged those (from reading `tryLoad` in here and in `errorLog.es6.js`)

test/lib/isFakeSubreddit.es6.js
schwers: I tend to lean towards redudancy in tests and thought it made it clearer to someone reading it that most subs are not fake subs. I can take them out if its just confusing

test/lib/isFakeSubreddit.es6.js
schwers: Same as above, I'll get rid of it

schwers:
:haircut:

src/routes.jsx
```diff
@@ -360,6 +359,42 @@ function routes(app) {
   // The homepage route.
   router.get('index', '/', indexPage);
 
+  router.get('random', '/r/:randomSubreddit(random|randnsfw)', function *() {
```
ajacksified: add `myrandom` here

schwers:
:haircut:

ajacksified:
:fish:

PR #501, opened 2016-02-19
Super-er Agent

nramadas:
Basically, create a helper that wraps `superagent` and auto-promisifies it. I also replaced superagent calls in client side code to use the new helper.

I left the server side code alone (`oauth.es6.js`, `console.es6.js`, and `routes.es6.js`). Mostly because (with the exception of `routes.es6.js`) they seemed to heavily depend on superagent's api to function. I could have converted `routes.es6.js` to use the new helper, but it doesn't really use promises at all, so i figured there was little advantage.

TODO:
 * [x] add tests for `makeRequest`

### DIFF
--- a/src/lib/makeRequest.es6.js
+++ b/src/lib/makeRequest.es6.js
@@ -1,0 +1,67 @@
+import superagent from 'superagent';
+
+// Subclass the Request method, and overwrite the `end` and `then` methods.
+class Request extends superagent.Request {
+  end() {
+    throw new Error('Please use \'then\' instead');
+  }
+
+  then(onSuccess, onError) {
+    // return a Promise
+    const p = new Promise((resolve, reject) => {
+      superagent.Request.prototype.end.call(this, function(err, res) {
+        if (err) {
+          reject({ ...res, ...err });
+        } else {
+          resolve(res);
+        }
+      });
+    });
+
+    // chain the promise if neccessary
+    if (onSuccess || onError) {
+      return p.then(onSuccess, onError);
+    }
+
+    return p;
+  }
+}
+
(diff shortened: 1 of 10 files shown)


dwick:
I just came here to say that I love the commit message

nramadas:
:eyeglasses: @ajacksified

src/lib/makeRequest.es6.js
```diff
@@ -0,0 +1,67 @@
+import superagent from 'superagent';
+import util from 'util';
+
+// Subclass the Request method, and overwrite the `end` and `then` methods.
+function Request(...args) {
```
ajacksified: Could you `class Request extends superagent.request` instead, or does superagent do weird things?

assets/js/client.es6.js
```diff
@@ -208,13 +201,13 @@ function sendTimings() {
...
 
-      superagent
+      makeRequest
         .post('/timings')
         .timeout(constants.DEFAULT_API_TIMEOUT)
         .send({
           rum: timings,
         })
-        .end(function() {});
+        .end();
```
ajacksified: rather than `end`, we should probably encourage calling `then` everywhere. (and maybe even get rid of `end` entirely.)

ajacksified:
I'm torn on the `end` vs `then`; I think if it's going to feel "promisey", we should hide `end` and say "call `then` to kick it off". I don't feel super strongly, though.

:nail_care: on the other question

nramadas:
:haircut:

ajacksified:
:fish:

dwick:
:running:

PR #502, opened 2016-02-24
Begin tracking API 504s and 429s

ajacksified:
:eyeglasses: @schwers

This was previously causing too much noise, but now that API errors are separated out, it would be a good thing to track. We should keep an eye on whether this slows down response times while it waits on logging to stdout if things get busy.

We may also want to track incoming API responses the same way we track outgoing HTTP codes, eventually.


schwers:
:fish:

PR #506, opened 2016-03-01
Redesign the search page

nramadas:


nramadas:

src/views/components/communities/CommunityRow.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+
+const T = React.PropTypes;
+
+function prettyPrintNum(number) {
```
ajacksified: this should probably be in a higher level file, and include tests

src/views/components/search/SearchBar.jsx
```diff
@@ -0,0 +1,87 @@
...
+    const inputValue = e.target.value;
+    this.setState({inputValue});
+    this.props.onInputChange(inputValue);
+  }
+
+  handleKeyPress(e) {
+    const { onSubmit } = this.props;
+    const { inputValue } = this.state;
+
+    if (e.key === 'Enter' && inputValue) {
```
ajacksified: wrap this in a form and use the form submit event, instead of checking e.key. also, `e.key` has poor browser support: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

src/views/components/search/SearchBarController.jsx
```diff
@@ -0,0 +1,108 @@
...
+    this.state = {
+      showSearchBar: false,
+    };
+
+    this.toggleSearchBar = this.toggleSearchBar.bind(this);
+    this.handleSearch = this.handleSearch.bind(this);
+  }
+
+  toggleSearchBar(e) {
+    e.preventDefault();
```
ajacksified: might as well call `this.preventClose(e)`. (maybe `preventClose` also belongs in baseComponent?)

src/views/components/search/SearchBarController.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+  renderSearchBar() {
+    const { ctx } = this.props;
+
+    return (
+      <div className='SearchBarController__overlay' onClick={ this.toggleSearchBar }>
+        <div className='SearchBarController__searchArea'>
+          <div
+            className='SearchBarController__close icon-nav-arrowback'
+            onClick={ this.toggleSearchBar }
```
ajacksified: is this onclick necessary? does the parent (`SearchBarController__overlay`) not capture the event? :no_bell:

src/views/pages/search.jsx
```diff
@@ -1,89 +1,152 @@
...
+    token: T.string,
+    after: T.string,
+    before: T.string,
+    data: T.object,
+    subredditName: T.string,
+    subreddits: T.object,
   };
-  
+
+  get track() {
```
ajacksified: static property? `track = 'search';`

src/views/pages/search.jsx
```diff
@@ -1,89 +1,152 @@
...
+    } else { // blank search page w/ no query
+      this.setState({ isLoading: false });
+    }
+  }
 
-    this.onSearch = this.onSearch.bind(this);
-    this._composeSortingUrl = this._composeSortingUrl.bind(this);
-    this.handleShowMoreClick = this.handleShowMoreClick.bind(this);
+  handleSortChange(newSort) {
+    this.redirectAfterNewSort({ sort: newSort });
```
ajacksified: if you want to save yourself a tiny bit of code:

```
 handleSortChange(sort) {
    this.redirectAfterNewSort({ sort });
```

src/views/pages/search.jsx
```diff
@@ -96,201 +159,212 @@ class SearchPage extends BasePage {
...
-        <div className="container search-bar-container">
-          <SearchBar action='/search'
-            onSearch={ this.onSearch }
-            defaultValue={ this.props.ctx.query.q }
+      <div className='SearchPage__links'>
+        <div className='SearchPage__linksHeader clearfix'>
+          <div className='SearchPage__linksHeaderTitle'>Posts</div>
+          <div className='SearchPage__linksHeaderTools'>
+            <div className='SearchPage__linksHeaderSort'>
+              <SortSelector
```
ajacksified: :no_bell: not from this PR, but sortselector should be a link or a button; something with a hover state. something to fix later.

ajacksified:
A couple things - functionality looks great. :nail_care:

src/views/components/search/SearchBar.jsx
nramadas: the reason I went with an `onChange` event is because it is easier to erase input fields when the input's value is managed by the state. and the `e.key` is something that should be normalized with React's synthetic event. i'm open to other suggestions for handling the reset though.

are no-script users something we target? i wasn't aware no-script was a high priority for us.

ajacksified: Yes, more in the context of _broken javascript_ than _js turned off_. I didn't realize React normalizes it, which is nice, but a form submit is simpler and gets you the no-js version for free (and, besides, runs less javascript while a user is typing.)

Resetting the value can done through the ref.

src/lib/prettyPrintNum.es6.js
```diff
@@ -0,0 +1,3 @@
+export default function prettyPrintNum(number) {
+  return number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
```
schwers: You can use `lib/formatNumber` for this. These tests are probably better though.

nramadas: derp. i should have searched harder for this

nramadas:
:haircut:

ajacksified:
:fish:

src/views/components/ListingPaginationButtons.jsx
```diff
@@ -5,12 +5,19 @@ import BaseComponent from './BaseComponent';
 
 class ListingPaginationButtons extends BaseComponent {
   static propTypes = {
+    listings: React.PropTypes.array.isRequired,
```
schwers: :no_bell: I'd approve switching `React.PropTypes` to `T` while you're in here, if you wanted to

src/views/components/ListingPaginationButtons.jsx
```diff
@@ -70,15 +77,16 @@ class ListingPaginationButtons extends BaseComponent {
     } = this.props;
 
     const {
+      preventUrlCreation,
       listings,
       pageSize = 25,
```
schwers: Should the url creation code be using `this.props.pageSize`?

src/views/components/communities/CommunityRow.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+const T = React.PropTypes;
+
+function renderIcon(iconUrl, url, color) {
+  const style = color ? { backgroundColor: color } : null;
+
+  return (
+    <a className='CommunityRow__icon' href={ url }>
+      { iconUrl
+        ? <img className='CommunityRow__iconImg' src={ iconUrl }/>
```
schwers: You need to apply style when there's an icon as well. There are multiple subreddits that rely on this behavior and have icons with transparent backgrounds. (Which is neat because they can easily change their look by tweaking the background color.) r/apple or r/smoo are good ones to test with

assets/less/components/communities/CommunityRow.less
```diff
@@ -0,0 +1,71 @@
...
+  &__add {
+    width: 40px;
+    padding: 0 @grid-size;
+    text-align: center;
+  }
+
+  &__addIcon {
+    &[class*=" icon-"] {
+      font-size: 3 * @grid-size;
+      color: @blue;
```
schwers: Why not use the `blue` and `lime` modifiers from `icon-font.less`?

assets/less/components/search/SearchPage.less
```diff
@@ -0,0 +1,104 @@
...
+    line-height: 2;
+  }
+
+  &__communitiesHeaderTitle {
+    float: left;
+  }
+
+  &__communitiesHeaderMore {
+    float: right;
+    font-size: 13px;
```
schwers: There's a few `font-size: 13px` in here, it looks thats to get back to normal font size the wrappers set for alignment? Could these just be `font-size: @font-size-base;`?

schwers:
There should probably be an empty state for when there are no search results reddit-wide. Right now it shows up blank (eg search string = 'thorn tlw rap battle coffee yoda'). Bonus points for the awesome Moose&Snoo graphic Erica has

schwers:
Looks really good! Just a few small edgecases and nits. :nail_care:

src/views/components/ListingPaginationButtons.jsx
nramadas: I think it uses it to determine if it needs to render a 'next page' button. if the number of listings equals the page size, it'll add a button to paginate.

nramadas:
:haircut:

assets/js/trackingEvents.es6.js
```diff
@@ -364,6 +374,10 @@ function trackingEvents(app) {
       gaSend('send', 'event', 'messages', 'reply', message.text.match(/\S+/g).length);
     }
   });
+
+  app.on('searchBar', function(payload) {
+    eventSend('search_events', payload.type, omit(payload, 'type'));
```
schwers: :no_bell: do we need the event out before releasing to track changes?

src/views/components/TopNav.jsx
```diff
@@ -118,8 +119,12 @@ class TopNav extends BaseComponent {
           <a
             className='MobileButton TopNav-floaty'
             href={ `${currentSubredditPath}/search` }
+            data-no-route={ true }
```
schwers: :no_bell: Out of curiosity why data-no-route; does `SearchBarController` squashing the click not work? On pc there's a little bit of hit area where you can click the icon but not be clicking the SearchBarController but _shrug_ can't repro on a phone, so doesn't seem like an issue.

schwers:
:fish: ship it

assets/js/trackingEvents.es6.js
nramadas: there isn't really an equivalent in our current product. these events fire from the `<SearchBarController />`, which doesn't exist on master.

The big event is the screen view event when search results actually load. That's already in place, so we should be good to go there if we want to track changes.

schwers: okay, sounds good

ajacksified:
:fish::fish::fish:

PR #507, opened 2016-03-01
Add secret oauth to auth header for registration

ajacksified:
:eyeglasses: @KeyserSosa and @schwers 

This will allow r2 to perform different checks for internal, trusted clients, such as allowing reddit-mobile to implement its own captcha.


schwers:
:fish:

KeyserSosa:
:fish: looks plausible :)

PR #508, opened 2016-03-02
Enable Listings Redesign for employees

schwers:
This enables listings for employees. 

There's still some:
* refactoring to be done for PostContent's media state, but I think that's better for a subsequent PR. 
* a few props like winWidth aren't getting the correct defaults on some auxiliary pages with Listings, I'll fix those and pull the commits into the PR. But for now I wanted to start getting feedback asap. 

I did some refactoring of dropdown rows, defining feature flags (just the names), and voting behavior. Those could be pulled out into a separate PR but it might not be worth it without these changes, up for discussion.

For tests, it would be my preference to delay until we try adding https://github.com/airbnb/enzyme as it would allow us to more easily and thoroughly test components.

**Screenshots**:

image

image

image

**Compact**:

image

image

image

**Details**:

image
image

image


assets/less/components/listings/PostContent.less
```diff
@@ -0,0 +1,205 @@
+.PostContent {
+
+  &.size-compact {
+    width: @new-compact-thumbnail-width;
+    height: @new-compact-thumbnail-width;
```
ajacksified: :no_bell: make sure to nuke the `new-` from these styles once the old stuff is kill

assets/less/components/listings/PostContent.less
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    &:hover {
+      color: @alien-blue-text;
+    }
+  }
+
+  &__link-bar-text {
+    display: inline-block;
+    width: calc(~'100% - 3em');
```
ajacksified: :no_bell: check `calc` in old browsers - specifically, make sure that this _not_ computing properly still works in old android browser (non-chrome).

assets/less/components/listings/PostContent.less
```diff
@@ -0,0 +1,205 @@
...
+      background-image: url(../img/placeholder.svg);
+    }
+  }
+
+  &__video-wrapper {
+    background-color: @light-gray;
+  }
+
+  &__iframe-wrapper {
+    /* fix ios8, ios5 iframe scrolling */
```
ajacksified: :100:

assets/less/components/listings/PostContent.less
```diff
@@ -0,0 +1,205 @@
...
+  &__nsfw-warning-button {
+    line-height: 0;
+    width: 100%;
+    border-radius: 3px;
+    border: 1px solid @white;
+    padding: 20px;
+  }
+
+  @playback-icon-size: 48px;
+  &__playback-action-circle {
```
ajacksified: :no_bell: what do you think about a slight opacity on the play button?

assets/less/components/listings/PostHeader.less
```diff
@@ -0,0 +1,184 @@
+.PostHeader {
+
+  // We can't easily tell from react if calling a componet is null or not.
```
ajacksified: `component`

assets/less/components/listings/PostHeader.less
```diff
@@ -0,0 +1,184 @@
...
+    border: 1px solid @grey-text;
+    background-color: @off-white;
+    color: @dark-grey;
+  }
+
+  &__sponsored-flair {
+    color: @blue;
+  }
+
+  &__lime-text {
```
ajacksified: maybe this should be `mod-text`?

src/views/components/behaviorcomponents/VoteableBehaviorComponent.es6.js
```diff
@@ -0,0 +1,102 @@
...
+import { models } from 'snoode';
+
+import propTypes from '../../../propTypes';
+import validatePropTypes from '../../../lib/validatePropTypes';
+const T = React.PropTypes;
+
+export default class VoteableBehaviorComponent {
+  // we're a plain class, but we're going to use react proptypes for debug warnings
+  // in this way a component can have vote arrows, but render them however
+  // it chooses -- no overriding or css magic necessary
```
ajacksified: this is an interesting idea. :+1:

src/views/components/dropdown/DropdownReportRow.jsx
```diff
@@ -0,0 +1,102 @@
...
+    };
+
+    this.toggleReportFormVisible = this.toggleReportFormVisible.bind(this);
+    this.onReportSubmitClicked = this.onReportSubmitClicked.bind(this);
+    this.onReportFormClicked = this.onReportFormClicked.bind(this);
+    this.onReportReasonInputChange = this.onReportReasonInputChange.bind(this);
+  }
+
+  toggleReportFormVisible() {
+    this.setState({ showwingReportForm: !this.state.showwingReportForm });
```
ajacksified: `showwingReportForm` -> `showingReportForm`

src/views/components/listings/NewListing.jsx
```diff
@@ -0,0 +1,143 @@
...
+          user={ user }
+          showAds={ showAds }
+          loid={ loid }
+          showHidden={ showHidden }
+          hideSubredditLabel={ hideSubredditLabel }
+          subredditTitle={ subredditTitle }
+          subredditIsNSFW={ subredditIsNSFW }
+          showOver18Interstitial={ showOver18Interstitial }
+        />
+        { shouldRenderPaginationButtons ? this.renderPagiationButtons(compact) : null }
```
ajacksified: `renderPagiationButtons` -> `renderPaginationButtons`

src/views/components/listings/PostAndCommentList.jsx
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    this.hasListeners = false;
+
+
+    this.lazyLoad = this.lazyLoad.bind(this);
+    this.afterAdDidLoad = this.afterAdDidLoad.bind(this);
+  }
+
+  get srnames() {
```
ajacksified: we should standardize on when we use getters - I'm not sure what the difference between this and, say, `hasAd` is. (both return a value based on props.)

src/views/components/listings/PostAndCommentList.jsx
```diff
@@ -0,0 +1,262 @@
...
+
+  lazyLoad() {
+    const { postsAndComments } = this.props;
+    const loadedDistance = this.getLoadedDistance();
+
+    // We load one item at a time during scrolling to not be
+    // super load heavy. If we hit the bottom of the function
+    // everything is loaded and we can remove the loader listneners
+
+    for (let i = 0; i < postsAndComments.length; i++) {
```
ajacksified: `postsAndComments.forEach((p, i) => ` is marginally preferable.

src/views/components/listings/PostContent.jsx
```diff
@@ -0,0 +1,674 @@
...
+  }
+
+  // need this to make iOS really autoplay the gif
+  gif.play();
+}
+
+//there are css values in aspect-ratio.less that must correlate with _WIDEST and _TALLEST
+const _WIDEST = 3;
+const _TALLEST = 1 / 3;
+function _limitAspectRatio(aspectRatio) {
```
ajacksified: you should probably export all of these functions and write tests. (I know they weren't there before, but since you're touching them anyways...)

src/views/components/listings/PostContent.jsx
```diff
@@ -0,0 +1,674 @@
...
+  return PLAYABLE_TYPE.NOT_PLAYABLE;
+}
+
+function forceProtocol(url, https) {
+  const urlObj = URL.parse(url);
+  urlObj.protocol = https ? 'https:' : urlObj.protocol;
+  return URL.format(urlObj);
+}
+
+class LinkDescriptor {
```
ajacksified: I wonder if ES8 will have structs.

src/views/components/listings/PostContent.jsx
```diff
@@ -0,0 +1,674 @@
...
+    const selftextNode = this.buildSelfTextContent(post, isCompact, single);
+
+    if (!mediaContentNode && !selftextNode) {
+      if (isDomainExternal && !isCompact) {
+        // When in compact mode, the PostHeader Component is responsible for rendering
+        // outbound links. But when in list mode, we are, so make sure we render
+        // the linkbar to outbound links if needed.
+        return this.renderMediaContent(null, false, true, post.domain, outboundLink);
+      }
+      return null;
```
ajacksified: is `null` specifically necessary, or can you just return?

src/views/components/listings/PostContent.jsx
```diff
@@ -0,0 +1,674 @@
...
+
+    let nsfwNode;
+    if (needsNSFWBlur) {
+      nsfwNode = this.renderNSFWWarning(isCompact);
+    }
+
+    const { single } = this.props;
+    const aspectRatio = this.getAspectRatio(single, previewImage.width, previewImage.height);
+
+
```
ajacksified: extra newline :scream:

src/views/components/listings/PostDropdownController.jsx
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.hidePost = this.hidePost.bind(this);
+    this.savePost = this.savePost.bind(this);
+    this.editPost = this.editPost.bind(this);
+    this.deletePost = this.deletePost.bind(this);
+    this.reportPost = this.reportPost.bind(this);
```
ajacksified: :no_bell: perhaps baseController needs a `bindAll`

src/views/components/listings/PostHeader.jsx
```diff
@@ -0,0 +1,263 @@
...
+
+
+import {
+  isPostNSWF,
+} from './listingUtils';
+
+const T = React.PropTypes;
+
+const SEPERATOR = (
+  <span className='PostHeader__seperator PostHeader__flush-w-icon' />
```
ajacksified: :no_bell: could this be a CSS :after rule instead?

ajacksified:
Functionality notes:

* Subreddit "do not display thumbnails" prefs aren't respected; if there is a gif, it is always embedded regardless of setting. (/r/me_irl, for example)
* Compact thumbnails are sometimes gifs instead of a static image when there is no thumbnail; maybe there should be nothing, or a placeholder? (/r/me_irl/comments/48hd8o/me_irl/)

Awesome work. Big, chewy PR. Feature flag works splendidly.

:nail_care:

assets/less/components/listings/PostContent.less
schwers: I'll test it out. In general how far back should we support? http://caniuse.com/#search=calc claims Android browser 4.3 and back don't support calc. We depend on it for layout in a few places now that _probably_ break when calc isn't supported. I can make a pr to move away from calc if that's the case and we want 4.3 & below

ajacksified: We do want 4.3; last I  checked, it was 5%ish of our traffic. It doesn't
have to be pristine in 4.3, but it shouldn't be obviously broken.

assets/less/components/listings/PostContent.less
schwers: I think so :)

assets/less/components/listings/PostContent.less
schwers: I'm leaning towards completely opaque, it grew on me. I'll try it out though

src/views/components/listings/PostAndCommentList.jsx
schwers: Agreed, my thoughts were that srnames will not change ever based on props. But hasAd technically changes if the ad is mounted or not. I can turn this into a regular function if its more clear

src/views/components/listings/PostContent.jsx
schwers: Sounds good, I'll move this and the best fit finding code into another file and add tests

src/views/components/listings/PostContent.jsx
schwers: that would be great

schwers:
@ajacksified re: functionality notes

* Respecting subreddit settings: I think this may be working correctly. I checked r/me_irl logged out forcing the new listings and don't see the gifs. I then noticed I had turned on the 'show thumbnails next to gifs' turned on for my account, so when I'm logged in I see them. However if I change my thumbnail preference to the 'respect the subreddit settings' option, the gifs go away (the api no longer sends back media / embed / thumbnail). Is it possible your account has the 'force thumbnails' option turned on?

* Gif thumbnails: This post in particular seems weird, we aren't getting back anything in the preview object or thumbnail but the oembed's thumbnail url is a gif. I could add some logic to try detecting a gif and running the html5 converter and see if we can generate the static image url ourselves. I think having the gif is better than nothing or the placeholder though, especially as this seems like a rare edge case

assets/less/components/listings/PostContent.less
schwers: It looks like the link itself is okay most of the time, but the post in compact seems broken



Patching that

src/views/components/listings/PostAndCommentList.jsx
schwers: I'm switching it `Array.some` instead as that will let me break out of the loop early. This makes the `removeListeners` logic clearer as well :D

schwers:
:haircut: @ajacksified 

:eyeglasses: @nramadas @curioussavage

schwers:
:eyeglasses: spez for notifying of visual changes, bunch of screenshots and it's on staging

ajacksified:
:fish::fish::fish:

assets/less/components/listings/PostContent.less
```diff
@@ -0,0 +1,207 @@
...
+    top: 0;
+  }
+
+  &__image-link {
+    display: block;
+    background-size: cover;
+    background-repeat: no-repeat;
+    background-position: center;
+    background-color: @light-gray;
+    width: 100%;
```
nramadas: doesn't `display: block` imply this?

assets/less/components/listings/PostFooter.less
```diff
@@ -0,0 +1,94 @@
+.PostFooter {
+  .clearfix;
```
nramadas: what does this do?

assets/less/components/listings/PostContent.less
schwers: ya, fixing now

assets/less/components/listings/PostFooter.less
schwers: fixes height technically, http://stackoverflow.com/questions/8554043/what-is-clearfix
In theory its good to do when a div has floated content. Defined in `snooboots/less/utilities.less`

nramadas: just noting that we talked about this. I promise I know what clearfix is :), I was more confused about less's extension syntax. I didn't know `.foo;` extended the current class's rules with foo's.

src/propTypes.es6.js
```diff
@@ -60,90 +60,100 @@ const _subreddit = shape({
...
+  subreddit: string,
+  thumbnail: string,
+  title: string.isRequired,
+  token: string,
+  url: string.isRequired,
+});
+
+const _postOrComment = oneOfType([_comment, _listing]);
+
+export default {
```
nramadas: I feel like export default-ing an object is a bad practice. According to spec,
```es6
export default { foo, foo2 };
```
is NOT equivalent to
```es6
export foo;
export foo2;
```

If you want to export multiple items in one statement, I prefer something like:
```
export { foo, foo2 }
```
So we're spec compliant.

src/views/components/listings/ListingRedesignProxy.jsx
```diff
@@ -0,0 +1,94 @@
...
+      prevUrl,
+      pageSize,
+      hideUser,
+      showHidden,
+      listingClassName,
+    } = this.props;
+
+    return (
+      <NewListing
+        postsAndComments={ listings }
```
nramadas: :no_bell: most of these props share the same name. maybe:
```
<NewListing
    { ...this.props }
    postsAndComments={ this.props.listings }
/>

schwers: I might do once its shipped everywhere. This greatly helped track all the properties that are coming in from splats, as there's a bunch. React context will help _sortof_ help with this in the future.

PR #509, opened 2016-03-02
Events: add track getter to the pages that did not have it

curioussavage:
In basePage the  method is not called unless
there is a  property. This is a hack to make sure
finish is called once since bodyLayout extends basePage.
Since chariot will fix the issue with basePage code running
twice this is an easy way to make sure pageview events fire on
every page.

### DIFF
--- a/src/views/layouts/BodyLayout.jsx
+++ b/src/views/layouts/BodyLayout.jsx
@@ -16,6 +16,10 @@ class BodyLayout extends BasePage {
     showEUCookieMessage: React.PropTypes.bool,
   };
 
+  get track() {
+    return false;
+  }
+
   constructor(props) {
     super(props);
 
--- a/src/views/pages/BasePage.jsx
+++ b/src/views/pages/BasePage.jsx
@@ -14,6 +14,10 @@ class BasePage extends BaseComponent {
     app: React.PropTypes.object.isRequired,
   };
 
+  get track() {
+    return true;
+  }
+
   constructor (props) {
     super(props);
 


src/views/pages/error.jsx
```diff
@@ -13,7 +13,11 @@ class ErrorPage extends BasePage {
     title: React.PropTypes.string.isRequired,
     referrer: React.PropTypes.string,
   };
-  
+
+  get track() {
```
ajacksified: if `basePage` set `get track` to `return true`, you could override it with `return false` in the layout instead of duplicating it across all of the pages.

ajacksified:
:nail_care: for simplification

src/views/pages/error.jsx
curioussavage: I had that idea too. I think @schwers was not a fan. Either way not a big deal, but lets do it then.

src/views/pages/error.jsx
```diff
@@ -13,7 +13,7 @@ class ErrorPage extends BasePage {
     title: React.PropTypes.string.isRequired,
     referrer: React.PropTypes.string,
   };
-  
+
```
curioussavage: oops, will rebase this out

curioussavage:
:haircut: @ajacksified @schwers

src/views/pages/error.jsx
schwers: @curioussavage Sorry there must have been some communication. I thought defaulting to `true` and overriding in `BodyLayout` was what we decided to do yesterday.

schwers:
Tested, works fine. Just remove the whitespace and :fish:

ajacksified:
e7d3d00895d32eab88fc11282f564c6765bc51c3

PR #510, opened 2016-03-02
Remove mobile-web-app-capable meta tag to avoid full-screen on iOS

umbrae:
The mobile-web-app-capable tag makes the app open in full-screen
mode when saved to your home screen, but that's not super ideal as
many URLs link out to other sites on reddit, which brings up the
browser chrome anyway, causing an app switch. I think we should
remove this and let it open up in the browser chrome as normal.

Thoughts?

:eyeglasses: @schwers or @ajacksified

### DIFF
--- a/src/views/layouts/DefaultLayout.jsx
+++ b/src/views/layouts/DefaultLayout.jsx
@@ -110,7 +110,6 @@ function DefaultLayout (props) {
           content='width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no'
         />
         <meta name='theme-color' content={ keyColor } />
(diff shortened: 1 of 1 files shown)


ajacksified:
A++. Thought we nuked this a while ago.

:fish:

schwers:
This is much better :fish:

curioussavage:
I tested this the other day with `content=no` which does show the url bar. I wonder if there is any difference though

umbrae:
https://github.com/reddit/reddit-mobile/commit/f3146ac2dc3116a7dd4bcb9ce070589f345b51f8

Issue #513, opened 2016-03-05
Ctrl+click does not open link in new tab on desktop browsers


polybuildr:
While I guess this isn't a conventional way of using the mobile reddit site, ctrl+click ing on a title link in a subreddit, say https://m.reddit.com/r/jokes does not open in a new tab. Instead it changes views (single page app stuff) and loads content in the same tab.

PR #514, opened 2016-03-07
Don't capture ctrl/cmd key clicks

ajacksified:
For issue #513


schwers:
Tested this and it doesn't work for me on latest Chrome OSX locally. The problem appears to be caused by our version of FastClick; `e.metaKey` and `e.ctrlKey` are always false until I remove fastClick and then this works as expected. Click delay seems no-existent in iOS and Chrome has fixed the click delay for a while, so it could be fine to remove fastClick entirely

schwers:
:nail_care:

schwers:
Build this morning was broken for some reason, never mind :fish:

PR #517, opened 2016-03-09
Fix ads response check

dwick:
:eyeglasses: @schwers 

bug introduced here: https://github.com/reddit/reddit-mobile/commit/a8e00f2b66f4e25082f0c57626b1f7970eb198b6#diff-2df9f64054b7cb0e816eade2a7a72605R90


schwers:
makes sense, :fish:

PR #520, opened 2016-03-10
Add r/ distinction to subreddit names

schwers:
This adds r/ to subredit display names in the community menu and in search results. We're doing this when rendering posts and post dropdrowns, so this makes things consistent. Especially because the r/ makes things feel uniquely Reddit. Both of these cases are lists so the distinction is smaller and lighter than normal font to make scanning the list easier.

image

image

### DIFF
--- a/assets/less/components/communities/CommunityRow.less
+++ b/assets/less/components/communities/CommunityRow.less
@@ -44,6 +44,12 @@
     color: @off-black;
   }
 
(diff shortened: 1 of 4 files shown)


schwers:
:eyeglasses: @ajacksified or @nramadas

ajacksified:
:fish:

PR #521, opened 2016-03-10
Add time filter for top|controversial listings. Hide username because it eats too much space

schwers:


schwers:
:eyeglasses: @ajacksified

ajacksified:
:fish:

PR #523, opened 2016-03-14
Don't pass urls to router when already on that page

schwers:


schwers:

curioussavage:
:fish:

PR #527, opened 2016-03-15
Fix linter error, remove unneeded reference to token in new listings

schwers:


dwick:
unsolicited :fish:

schwers:
:eyeglasses: quick sanity check  @curioussavage

curioussavage:
:fish: :tropical_fish:

PR #533, opened 2016-03-16
Make it so that the SortSelector does not crash when given arbitrary sorts

nramadas:


nramadas:
:eyeglasses: @schwers

ajacksified:
:fish:

schwers:
:fish: :tropical_fish: :party:

PR #537, opened 2016-03-16
oauth: Fix TypeError

curioussavage:
The refresh request is not working and throwing errors
because we were not checking if the token existed before
accessing properties

:eyeglasses: @schwers or @ajacksified or @nramadas

### DIFF
--- a/assets/js/clientLib.es6.js
+++ b/assets/js/clientLib.es6.js
@@ -120,28 +120,32 @@ export function refreshToken (app) {
     .then(res => {
       const token = res.body;
 
-      const now = new Date();
-      const expires = new Date(token.tokenExpires);
+      if (token && token.token) {
+        const now = new Date();
+        const expires = new Date(token.tokenExpires);
 
-      Object.assign(app.getState('ctx'), {
-        token: token.token,
-        tokenExpires: token.tokenExpires,
-      });
-
-      app.setState('refreshingToken', false);
-      app.emit('token:refresh', token);
-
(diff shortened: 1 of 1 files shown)


assets/js/clientLib.es6.js
```diff
@@ -120,28 +120,32 @@ export function refreshToken (app) {
...
+      if (token && token.token) {
+        const now = new Date();
+        const expires = new Date(token.tokenExpires);
 
-      Object.assign(app.getState('ctx'), {
-        token: token.token,
-        tokenExpires: token.tokenExpires,
-      });
+        Object.assign(app.getState('ctx'), {
+          token: token.token,
```
nramadas: https://en.wikipedia.org/wiki/Semantic_satiation

curioussavage: yap...

curioussavage: ill defer improvements to that to a future pr though

curioussavage:
looks like this might require some more work. I was assuming a retry would occur but we might need to manually retry and or refresh the page if the token already expired

kjoconnor:
If you have any OAuth questions please let me know! Especially whatever pain points you are running into. I'm working with the mobile team on OAuth stuff so happy to just be a reference if needed.

ajacksified:
:fish: - better than current, but we need a good solution to handle failed api requests that don't involve routing. probably snoode v2.

PR #538, opened 2016-03-16
Fix replace of undefined error

schwers:
:eyeglasses: @nramadas or @curioussavage


nramadas:
:fish:

curioussavage:
:tropical_fish:

PR #542, opened 2016-03-18
Update comment vote arrows. Add vote bounce animation.

schwers:
This updates comment's voteing arrows to:
* have the same hit areas as the listings arrows. Part of this change is making the score an upvote target.

All vote arrows now use the bounce animation, similar to the app betas.
- [ ] Check that the animation is okay on old android phones
- [ ] Check with design that the easing curve / durations are okay

### DIFF
--- a/assets/less/animations/importAll.less
+++ b/assets/less/animations/importAll.less
@@ -1,0 +1,1 @@
+@import "voteBounce.less";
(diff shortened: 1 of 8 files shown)


ajacksified:
> voteing

ajacksified:
:fish:

PR #543, opened 2016-03-24
add visual studio code files to gitignore

curioussavage:
:eyeglasses: @schwers 

cc @ajacksified and @nramadas 

any other stuff that should be in here?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,4 @@ start.sh
 workspace.json
 .DS_Store
 .idea
-
+.vscode/*


ajacksified:
:fish:

spladug:
unsolicited but fwiw, I don't think IDE-specific stuff should be in a project's gitignore. You should configure your personal environment (via global gitignore) for any IDEs you use and projects should only have the stuff that's relevant to that project. Otherwise you'll end up with everyone's laundry list of editors in your gitignore which isn't really that great in the long run.

curioussavage:
^ this is a great point. I think this should be closed in favor of spladug's suggestion.

schwers:
@curioussavage I agree. I didn't realize there was a global config; that seems like the perfect place for this kind of stuff

curioussavage:
ok, closing.

ajacksified:
Fair  enough. My idea was it lowers the barrier slightly for open source
prs, but yeah, could get messy fast.

PR #545, opened 2016-03-24
Add CSP headers

ajacksified:
This restricts the domains that can host content that is run on mobile
web. The headers added are:

* Self to scripts, styles, and fonts, which lets us load locally
* Localhost for scripts, for livereload
* Google analytics for scripts, for obvious reasons
* ASSET_URL from config, for cdn'd stuff

:eyeglasses: @nramadas (for smart banner) and @spladug

### DIFF
--- a/src/server/index.jsx
+++ b/src/server/index.jsx
@@ -28,7 +28,7 @@ import setLoggedOutCookies from '../lib/loid';
 
 import Config from '../config';
 
-const ignoreCSRF = ['/timings', '/error'];
+const ignoreCSRF = ['/timings', '/error', '/csp-report'];
 
 function getBucket(loid) {
   return parseInt(loid.substring(loid.length - 4), 36) % 100;
@@ -57,7 +57,9 @@ function skipAuth(app, url) {
     url.indexOf('/login') === 0 ||
     url.indexOf('/register') === 0 ||
     url.indexOf('/oauth2') === 0 ||
-    url.indexOf('/timings') === 0
+    url.indexOf('/timings') === 0 ||
+    url.indexOf('/error') === 0 ||
+    url.indexOf('/csp-report') === 0
   );
 }
 
@@ -101,11 +103,14 @@ function setCompact(ctx, app) {
 const defaultConfig = Config;
 function formatBootstrap(props) {
   let p;
+
   for (p in props.config) {
     if (!defaultConfig.hasOwnProperty(p)) {
       delete props.config[p];
     }
   }
+
+  props.nonce = props.ctx.csrf;
 
   return props;
 }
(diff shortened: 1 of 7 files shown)


ajacksified:
Some info: http://www.html5rocks.com/en/tutorials/security/content-security-policy/

spladug:
We probably want to start out in report-only mode and collect reports somewhere to be safe?

https://developer.mozilla.org/en-US/docs/Web/Security/CSP/Using_CSP_violation_reports

ajacksified:
:haircut: 

related: https://github.com/reddit/horse-react/pull/14

spladug:
some interesting links:

* https://blogs.dropbox.com/tech/2015/09/csp-the-unexpected-eval/
* https://blogs.dropbox.com/tech/2015/09/on-csp-reporting-and-filtering/

src/server/routes.es6.js
```diff
@@ -93,6 +93,26 @@ const serverRoutes = function(app) {
...
+
+      const report = this.body['csp-report'];
+
+      const log = [
+        'CSP REPORT',
+        report['document-uri'],
+        report['blocked-uri'],
+      ];
+
+      console.log(log.join('|'));
```
spladug: you can send server-side events to the data pipeline, right? it might be useful to pipe these over there so we can actually look into the details of reports in a more structured manner?

spladug:
This looks really good to me, I'm just gonna do some reading on the specific values being set on CSP to make sure I'm not missing anything then come back with appropriate emoji. Thanks for working on this!

src/server/routes.es6.js
ajacksified: Yeah, my thought was to get this out as
-is now, and get it into the data pipeline after we come up with a spec
with the data team.

src/server/index.jsx
```diff
@@ -323,6 +328,24 @@ class Server {
       this.set('X-Content-Type-Options', 'nosniff');
       this.set('X-XSS-Protection', '1; mode=block');
 
+      const GAURL = 'http://www.google-analytics.com';
+      const SELF = `'self'`;
+      const INLINE = `'unsafe-inline'`;
+      const LOCALHOST = 'localhost:*';
```
spladug: Why is `localhost:*` needed when you have `self`? Is the port different for statics in dev or something?

src/server/index.jsx
```diff
@@ -323,6 +328,24 @@ class Server {
       this.set('X-Content-Type-Options', 'nosniff');
       this.set('X-XSS-Protection', '1; mode=block');
 
+      const GAURL = 'http://www.google-analytics.com';
```
spladug: We're loading GA over unsecured HTTP?

src/server/index.jsx
```diff
@@ -323,6 +328,24 @@ class Server {
...
+      const SELF = `'self'`;
+      const INLINE = `'unsafe-inline'`;
+      const LOCALHOST = 'localhost:*';
+      const ASSETS = app.config.assetPath;
```
spladug: This the static domain?

src/server/index.jsx
spladug: (and is `self` necessary when not in production?)

src/server/index.jsx
```diff
@@ -323,6 +328,24 @@ class Server {
...
+      const NONCE = `'nonce-${this.csrf}'`;
```
spladug: Here are some security notes on nonces: https://w3c.github.io/webappsec-csp/#security-nonces

I _think_ this will still be OK, but just for my edification how frequently do you get new modhashes?

src/server/index.jsx
```diff
@@ -323,6 +328,24 @@ class Server {
...
+
+      const cspRules = [
```
spladug: There are a lot of other `-src` directives, do we want to do a `default-src` to catch them all? (e.g. I assume we can block `object-src`)

spladug:
:nail_care: just some questions about directives.

src/server/index.jsx
ajacksified: While in dev, livereload loads a script over locahost:36789 or something like that. I'll restrict it to non-production, though.

src/server/index.jsx
ajacksified: Whoops, only from localhost.

src/server/index.jsx
ajacksified: Yup - https://www.w3.org/TR/CSP2/#source-list-path-patching

src/server/index.jsx
ajacksified: Once every server request; _probably_ once a session.

ajacksified:
also :eyeglasses: @schwers because of new listings - take a look at the image, frame, and media hosts, where we're directly embedding if a user hits `play`.

ajacksified:
:haircut: with the concession that we need to limit the allowed hosts asap

src/server/index.jsx
```diff
@@ -323,6 +328,49 @@ class Server {
...
+      const ORIGIN = app.config.origin;
+      const API = `${app.config.nonAuthAPIOrigin} ${app.config.authAPIOrigin}`;
+      const IMAGE_HOST = '*.redditmedia.com thumbs.gfycat.com *.imgur.com *.giphy.com *.ytimg.com';
+      const FRAME_HOST = '*.embedly.com *.imgur.com *.gfycat.com  *.youtube.com';
+      const MEDIA_HOST = '*.embedly.com *.imgur.com *.gfycat.com *.giphy.com';
+
+      const cspRules = [
+        `report-uri /csp-report`,
+        `default-src ${self} ${localhost} ${ASSETS} ${API} ${ORIGIN}`,
+        `frame-src ${FRAME_HOST}`,
```
spladug: `frame-src` is obsolete/deprecated. do you want to use `child-src` here? (or both for older browsers?)

src/server/index.jsx
```diff
@@ -323,6 +328,49 @@ class Server {
...
+      if (app.config.googleAnalyticsId) {
+        gaurl = `http${secure}://www.google-analytics.com`;
+      }
+
+      const INLINE = `'unsafe-inline'`;
+      const ASSETS = app.config.assetPath;
+      const NONCE = `'nonce-${this.csrf}'`;
+      const ORIGIN = app.config.origin;
+      const API = `${app.config.nonAuthAPIOrigin} ${app.config.authAPIOrigin}`;
+      const IMAGE_HOST = '*.redditmedia.com thumbs.gfycat.com *.imgur.com *.giphy.com *.ytimg.com';
```
spladug: where do these lists come from? is it gonna be hard to keep them in sync with whatever causes them to be needed? is there a way to make it more explicit on both ends that a list needs to be kept in sync?

spladug:
> with the concession that we need to limit the allowed hosts asap

what's this mean?

spladug:
thanks a bunch for doing this. i think it looks good as is (:fish:) but have a couple of questions.

src/server/index.jsx
ajacksified: It is going to be hard - I'd like to nuke them entirely and do everything through imgix, but there are some edge cases where some of these slip through somehow. We need to figure out why we're not getting preview images here, but for now it seems like this covers everything that I've come across through extensive browsing. It's bound to have a few rules missing.

src/server/index.jsx
ajacksified: I'll use both.

ajacksified:
> what's this mean?

It means I want to nuke imgur, gfycat, youtube, ytimg, and giphy from image/frame/media hosts, and instead use embedly or our own image scraper.

PR #548, opened 2016-03-24
Fix quoting of Hide in PostDropdown

schwers:
:eyeglasses: @curioussavage


ajacksified:
:fish:

schwers:
:fishing_pole_and_fish:

PR #550, opened 2016-03-28
Create a custom banner

nramadas:
:eyeglasses: @schwers

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -6,6 +6,7 @@ import { sendTimings, onError, loadShimThenInitialize, onLoad, render } from './
 window.onerror = onError;
 
 import '../../src/lib/dnt';
+import initBanner from '../../src/lib/initBanner';
 
 import forOwn from 'lodash/object/forOwn';
 
@@ -24,6 +25,15 @@ import trackingEvents from './trackingEvents';
 
 import setAppMethods from './appMethods';
 import setEvents from './clientEvents';
+import constants from '../../src/constants';
+
+const APP_DOWNLOAD_URLS = {
+  'index': constants.BANNER_URLS.FRONTPAGE,
+  'index.subreddit': constants.BANNER_URLS.LISTING,
+  'comments.index': constants.BANNER_URLS.COMMENTS,
+};
+
+const SMARTBANNER_BUCKET = 5;
 
 // A few es5 sanity checks
 if (!Object.create || !Array.prototype.map || !Object.freeze) {
@@ -86,6 +96,21 @@ function initialize(bindLinks) {
 
   // env comes from bootstrap from the server, update now that the client is loading
   app.state.ctx.env = 'CLIENT';
+
+  app.emitter.once('pageview', data => {
+    // add a smart banner/toaster if the user lands on certain pages.
+    const userId = (data.data.loid || data.data.user.id);
+    const userIdSum = userId.split('').reduce((sum, chr) => sum + chr.charCodeAt(0), 0);
+    const bucket = userIdSum % 100;
+    const featureEnabled = data.feature.enabled(constants.flags.SMARTBANNER);
+    const downloadUrl = APP_DOWNLOAD_URLS[data.actionName];
+    const inBucket = data.actionName === 'comments.index' ? bucket < SMARTBANNER_BUCKET : true;
+
+    if (featureEnabled && downloadUrl && inBucket) {
+      const { IMPRESSION, CLICK } = downloadUrl;
+      initBanner(IMPRESSION, CLICK);
+    }
+  });
 
   if (window.bootstrap.config.googleAnalyticsId) {
     trackingEvents(app);
--- a/assets/less/components/DefaultLayout.less
+++ b/assets/less/components/DefaultLayout.less
@@ -1,0 +1,20 @@
+.DefaultLayout {
+  &.banner-is-open {
+    margin-bottom: @banner-height + @grid-size;
+  }
+}
+
+.SmartBannerContainer {
+  .shadow;
+  position: fixed;
+  bottom: -@banner-height;
+  left: 0;
+  right: 0;
+  height: @banner-height;
+  z-index: 11; // this ensures it sits above the TopNav's boxShadow
+
+  .themeify({
+    border-top: @theme-border;
+    background-color: @theme-nav-bar-color;
+  });
+}
--- a/assets/less/components/SmartBanner.less
+++ b/assets/less/components/SmartBanner.less
@@ -1,0 +1,107 @@
+.SmartBanner {
+  @icon-size: 58px;
+  @button-container-size: 116px;
+  position: relative;
+  height: 100%;
+  width: 100%;
+  max-width: 460px;
+  padding: @grid-size;
+  padding-left: @half-grid-size;
+  margin: 0 auto;
+
+  .themeify({
+    background-color: @theme-nav-bar-color;
+  });
+
+  &__left,
+  &__right {
+    position: absolute;
+    top: 50%;
+    height: @icon-size;
+    margin-top: -(@icon-size / 2);
+  }
+
+  &__left {
+    left: 0;
+    right: @button-container-size;
+  }
+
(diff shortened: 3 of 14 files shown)


assets/js/client.es6.js
```diff
@@ -87,6 +89,17 @@ function initialize(bindLinks) {
   // env comes from bootstrap from the server, update now that the client is loading
   app.state.ctx.env = 'CLIENT';
 
+  app.emitter.once('pageview', data => {
```
schwers: :no_bell: quick rules confirmation, this only happens on a page-refresh of a comments page or directly landing on a comments page?

assets/js/client.es6.js
```diff
@@ -95,7 +108,7 @@ function initialize(bindLinks) {
   // config value after render.
   const beginRender = Date.now();
 
-  render(app, app.fullPathName(), true, app.modifyContext).then(function() {
+  render(app, app.fullPathName(), true, app.modifyContext).then(function(props) {
```
schwers: `props` looks like it isn't needed. just code cleanup?

assets/less/components/DefaultLayout.less
```diff
@@ -0,0 +1,27 @@
+.navbar-offcanvas-target {
```
schwers: I would just rename this to `.DefaultLayout` or `.DefaultLayout__body` etc. It doesn't look like we depend on this classname anywhere and would be a nice tiny step towards removing bootstrap stuff

assets/less/components/DefaultLayout.less
```diff
@@ -0,0 +1,27 @@
...
+  &.banner-is-open {
+    margin-top: 122px;
```
schwers: 122px should be in a variable, it's used in a lot of less in this patch, or is a part of a computation

assets/less/components/DefaultLayout.less
```diff
@@ -0,0 +1,27 @@
...
+
+    .TopNav {
+      top: 122px;
+    }
+
+    .OverlayMenu {
+      top: 170px;
```
schwers: @smartbanner-height + @top-nav-height

assets/less/components/SmartBanner.less
```diff
@@ -0,0 +1,59 @@
+.SmartBanner {
+  height: 100%;
```
schwers: Although I'm generally a fan of making components that can be have sizing from their parent, 122px seems like a necessity here given the current layout; eg too short leaves the download button overflowing and too tall leaves a bunch of whitespace

assets/less/components/SmartBanner.less
```diff
@@ -0,0 +1,59 @@
...
+  padding: @grid-size 2 * @grid-size;
+  background-color: @white;
+
+  &__top {
+    margin-bottom: @grid-size;
+  }
+
+  &__icon {
+    .SnooIcon {
+      height: 58px;
```
schwers: :no_bell: Something a little weird happens here on page load. The icon colors are inverted and then blink to the correct colors. might be worth looking into

src/lib/initBanner.es6.js
```diff
@@ -0,0 +1,51 @@
...
+import ReactDOM from 'react-dom';
+import moment from 'moment';
+import SmartBanner from '../views/components/SmartBanner';
+
+const TIME_LIMIT = 2;
+const TIME_INTERVAL = 'weeks';
+const ANIMATION_DELAY = 1000;
+
+export default function(url) {
+  if (!localStorage || !localStorage.getItem || !localStorage.setItem) {
```
schwers: You should try useing `lib:localStorageAvailable`. It captures the edgecase where `setItem` exists but either doesn't work or throws an exception when you call it (private mode safari in come cases)

schwers: If you don't want to call it directly, its set on config as well, could pass that in if you wanted to use the cached value

src/lib/initBanner.es6.js
```diff
@@ -0,0 +1,51 @@
...
+      } }
+    />,
+    bannerContainer
+  );
+
+  // animate the banner. add a small delay so that the content 'settles in'
+  // before we animate anything. it feels less jarring this way.
+  setTimeout(() => {
+    bannerContainer.style.top = '0px';
+    document.body.className = `${document.body.className} banner-is-open`;
```
schwers: I would use the `classList` api for this and the `onClose` callback

schwers:
:no_bell: because biz-logic, and this is mostly product related, not directed at you. Is showing the download banner on all subreddit pages / second in the list set in stone? My gut reaction from just a few minutes of browsing is that I really dislike it. I think it mostly stems from not being able to close it, even if I already have the app. (I don't _think_ the apps will be capturing urls so you even with the app you can end up on mweb with the banner still there. If the real urls were smart enough to try redirecting to the app if you have it that's one thing but...) As-is it feels very intrusive to the browsing experience. It especially stands out on subreddits with no preview images like `r/me_irl`

schwers:
Codewise, looks good! Mostly small cleanup nits, expected for a wip. Lots of biz-logic questions / concerns, sorry. :nail_care:

assets/less/components/SmartBanner.less
nramadas: it is mostly intentional until I get final designs. for example, the banner might be slightly larger in card view mode vs compact, and etc.

you're right in that the banner didn't make use of additional space, so i've updated it to do that.

assets/js/client.es6.js
nramadas: only for direct lands. navigating to the comments page in-app won't cause the banner to appear.

nramadas:
the positioning of the banner in the list isn't set in stone; i went with position 2 for easy testing (i didn't want to scroll to far to see it). beyond that, for now it does seem like they want the banner in the list, but during testing, product might decide that it's too invasive. the un-dismissable nature of the list banner is actually considered one of its 'pros', so we'll have to wait and see once people get to test it whether or not folks like it.

nramadas:
:haircut:

assets/less/components/SmartBanner.less
schwers: Okay cool, wasn't sure how locked in the designs were

nramadas:
:haircut:

assets/less/components/SmartBanner.less
```diff
@@ -0,0 +1,89 @@
+.SmartBanner {
+  @icon-size: 58px;
+  height: 100%;
+  width: 100%;
+  max-width: 460px;
+  padding: @grid-size;
+  padding-left: @half-grid-size;
+  margin: 0 auto;
+  display: flex;
```
schwers: Layout looks a little broken in Android 4.3, mostly because you can't see the CTA. I think that's the lowest we target right now so it's probably worth looking at?

image

schwers: We have some mixins for flexbox, not sure if those are worth using and would help the layout

src/lib/initBanner.es6.js
```diff
@@ -0,0 +1,49 @@
...
+import ReactDOM from 'react-dom';
+import SmartBanner from '../views/components/SmartBanner';
+import TrackingPixel from './TrackingPixel';
+import { shouldShowBanner, markBannerClosed } from './smartBannerState';
+
+const ANIMATION_DELAY = 1000;
+const ANIMATION_LENGTH = 200;
+const BANNER_HEIGHT = 94;
+
+export default function(impressionUrl, clickUrl) {
```
schwers: :no_bell: IIRC if you name the function here in the export it will show up in stacktraces. It sounds like we'll have sourcemaps working in prod so it might be worth doing

schwers:
Looks great, only concern is layout not showing CTA in old browsers with poor flexbox support (see screenshot).

schwers:
:nail_care:

nramadas:
💇

schwers:
Consider: https://github.com/nramadas/reddit-mobile/pull/1/files
Otherwise: :fish:

PR #551, opened 2016-03-28
listing: fix undefined error

curioussavage:
A reference to user.id was throwing an error since user was not defined

:eyeglasses: @schwers


schwers:
:fish: pls ship

PR #552, opened 2016-03-29
Improve table display in links/comments

ajacksified:
This adds borders and padding so that tables are easier to read in comments and links.

screen shot 2016-03-29 at 2 20 40 pm
screen shot 2016-03-29 at 2 21 22 pm

:eyeglasses: @nramadas

### DIFF
--- a/assets/less/components/tables.less
+++ b/assets/less/components/tables.less
@@ -1,4 +1,4 @@
-.PostContent, .Comment__body {
+.table () {
   table {
     margin: 1em 0;
     width: 100%;
@@ -9,3 +9,5 @@
     border-bottom: solid 1px @neutral;
   }
 }
+
+
--- a/assets/less/components/tables.less
+++ b/assets/less/components/tables.less
@@ -1,0 +1,11 @@
+.PostContent, .Comment__body {
+  table {
+    margin: 1em 0;
+    width: 100%;
+  }
+
+  td, th {
+    padding: .25em;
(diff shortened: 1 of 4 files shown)


assets/less/components/tables.less
```diff
@@ -0,0 +1,11 @@
+.PostContent, .Comment__body {
```
nramadas: I'm thinking it would be nice if instead, this was a mixin-able rule.

```css
.tableStyles { ... }
```

Then, we'd mix it into `.PostContent` and `.Comment__body`. Otherwise, we're starting to break some of our BEM rules, plus it's nice to know that all the styling for a particular component lives in the same place.

nramadas:
:nail_care:

assets/less/components/tables.less
ajacksified: Sure, I'll turn it into a mixin. That makes sense.

ajacksified:
:haircut:

curioussavage:
:+1:

nramadas:
:fish:

assets/less/components/tables.less
```diff
@@ -0,0 +1,13 @@
+.table () {
+  table {
+    margin: 1em 0;
+    width: 100%;
```
schwers: :no_bell: Does this fix the issue where really large tables overflow and cause horizontal scroll on the page?

ajacksified: It does not.

PR #553, opened 2016-03-29
gulp: enable source maps in prod

curioussavage:
Also fixes issue with build not running after an error

:eyeglasses: @ajacksified @nramadas @schwers 

please test!!

I gave up on the refactor and just tried using the plugins/'end' event I discovered in our existing build and with a little tweaking it worked!


ajacksified:
This looks good for prod, but it isn't building sourcemaps in dev. :nail_care:

curioussavage:
nope the source maps are inline just like they are now

ajacksified:
Nevermind - you get sourcemaps if you run `npm run watch` or `npm run dev`, but not in prod mode (`npm run build`). :fish:

curioussavage:
closing this as there was still an issue and we are really close to webpack/new framework

PR #556, opened 2016-03-30
Commment: disable buttons while a request is active

curioussavage:
Primarily for new comments but I decided to add it disable for other actions too


curioussavage:
:eyeglasses: @nramadas or @schwers

schwers:
:fish: :fishit: :shipit:

PR #558, opened 2016-03-30
messages: show reply button for messages

curioussavage:
message.replies is always an array now

:eyeglasses: @ajacksified


ajacksified:
:fish:

PR #560, opened 2016-03-31
Make it so nightmode can be activated via a query-param

schwers:


src/server/index.jsx
```diff
@@ -72,9 +72,13 @@ function setTheme(ctx, app) {
     return;
   }
 
-  const theme = (ctx.cookies.get('theme') || '').toString() === NIGHTMODE
+  let theme = (ctx.cookies.get('theme') || '').toString() === NIGHTMODE
     ? NIGHTMODE : DAYMODE;
 
+  if (ctx.query.nightmode === 'true') {
```
ajacksified: `ctx.query.theme === NIGHTMODE`. Also add in the opposite case (`ctx.query.theme === DAYMODE`). Maybe `if ([DAYMODE, NIGHTMODE].includes(ctx.query.theme) { theme = ctx.query.theme }` or something.

schwers:
:eyeglasses: @ajacksified

ajacksified:
:fish:

PR #563, opened 2016-04-01
nightmode fixes plus fix subreddit select in submit

curioussavage:
:eyeglasses: @nramadas 

just some quick fixes for night mode, plus 2 little fixes for submit


nramadas:
:fish:

PR #565, opened 2016-04-04
Change ?ref on "view desktop" to mweb_navbar

ajacksified:
:eyeglasses: @schwers and @drunken-economist 

@drunken-economist - made this change before we discussed moving mweb to utm, so wanted to get your input before this goes in.


drunken-economist:
:fish: 
Thanks for the change. We have a midas module in the works for UTM parsing, and not having to add `ref` stuff to it is nice

PR #567, opened 2016-04-05
Add some more logic to smart banner

nramadas:


src/lib/smartBannerState.es6.js
```diff
@@ -1,21 +1,114 @@
...
 
+const PAGE_PERCENTAGES = {
+  'comments.subreddit': 5,
+};
+
+const USE_TUNE = 100;
+
+const checkDeviceType = (allowedAgents, userAgentString) => {
+  return !!allowedAgents
+    .map(a => userAgentString.indexOf(a) > -1)
```
schwers: You can use `Array.some` instead of map -> filter -> length check

nramadas: good point

src/lib/smartBannerState.es6.js
```diff
@@ -1,21 +1,114 @@
...
+
+  // 4) Check the user agent
+  if (!checkDeviceType(ALLOWED_DEVICES, userAgent)) { return BASE_VAL; }
+
+  // 5) only show the banner to people from certain countries
+  if (!ALLOWED_COUNTRIES.has(country)) { return BASE_VAL; }
+
+  // Create a bucket; a few rules are going to depend on that
+  let userId = '';
+  if (loid || data.user) { userId = loid || data.user.id; }
```
schwers: :no_bell: When do we have neither?

nramadas: i don't know, and I don't think we ever will. But, i don't want to take any chances either

src/lib/smartBannerState.es6.js
```diff
@@ -1,21 +1,114 @@
...
+  const userIdSum = userId.split('').reduce((sum, chr) => sum + chr.charCodeAt(0), 0);
+  const bucket = userIdSum & 100;
```
schwers: Should this be `userIdSum % 100`?

nramadas: i'm an idiot

schwers: False

schwers:
inb4 we need to add page source to TUNE urls

assets/js/client.es6.js
```diff
@@ -99,28 +90,19 @@ function initialize(bindLinks) {
...
-
-    if (featureEnabled && downloadUrl && inBucket && inCountry) {
-      const { IMPRESSION, CLICK } = downloadUrl;
-      initBanner(IMPRESSION, CLICK);
-    }
+    const {
+      show,
+      impressionUrl,
+      clickUrl,
+    } = shouldShowBanner({
```
schwers: :no_bell: Super-Nit:tm: `shouldShowBanner` as name _feels_ like it should be returning a bool versus an object that describes the banner-showing-state. But I don't have any replacement suggestions atm

schwers:
:nail_care: Just a small math nit

nramadas:
:haircut:

schwers:
:fish:

PR #568, opened 2016-04-06
redirects: add /live and /rules

curioussavage:
:eyeglasses: @ajacksified or @schwers 

this is mostly for the /live because of this https://reddit.atlassian.net/browse/MOBILEWEB-405 issue.

I'm not sure what will fix that, but either way I think its not a bad thing to have this redirect there. And it sucks that there is an active thread now and its pointing to an error page for mobile

The rules endpoint just mirrors what r2 does - send this url to the content policy


ajacksified:
awesome! this burns me all the time on r/politics.

🐟

PR #570, opened 2016-04-07
Support Blocks

schwers:
Depends on https://github.com/reddit/snoode/pull/98


schwers:
There's some tricky/hacking manipulation of <Comment />'s state because of how replies are loaded in, that part could use some extra thoughtful review. I think it's a good example of why we should continue considering redux as then we'd have better access to the full comment tree

schwers:
:eyeglasses: @ajacksified @nramadas

schwers:
We should add blocks in 2X soon, it's probably okay to keep this open for reference's sake, but we don't need it yet

phil303:
@schwers can we kill this?

schwers:
@phil303 I'm going to close it for now, I'll make sure there's a ticket for this

PR #573, opened 2016-04-08
routes: remove some endpoints

curioussavage:
remove routes that other applications will
never need to know about from routes


ajacksified:
:fish:

PR #577, opened 2016-04-14
events: add default arg to stripPrefix method

curioussavage:
:eyeglasses: @ajacksified


curioussavage:
:eyeglasses: @schwers

ajacksified:
🐟

PR #580, opened 2016-04-19
Add unhandledrejection listener on server and client

curioussavage:
this can help us find promises that need
rejection to be handled. currently only works
in chrome on the browser

👓  @ajacksified @schwers @nramadas 

I think this is really handy, just like `uncaughtException` its a nice helper in dev. only available in chrome right now though :(

see:
http://www.2ality.com/2016/04/unhandled-rejections.html?utm_source=javascriptweekly&utm_medium=email


curioussavage:
would be nice to have this in new framework too

ajacksified:
🎉

ajacksified:
:fish:

schwers:
:fish:

PR #585, opened 2016-05-02
Fix compact query param

curioussavage:
👓  @ajacksified  or @schwers 

Looks like there is no code that sets a compact cookie with a 'false' value anymore. With git blame it looks like this was intended to fix some clients that still had them. Changing this enables you to switch from compact back to card view by url query parameter.


ajacksified:
:fish:

schwers:
🐟

PR #587, opened 2016-05-02
Use server html

prashtx:
We have a bug in the rendering of expandContent, where we overzealously strip out some tags (notably del). We want to move away from client-rendered HTML in general, favoring HTML fields provided by the
server.  By switching to the correctly-rendered API fields, we can avoid the rendering bug and remove some dependencies, including the 3rd-party library that led to the original rendering bug.

Depends on a Snoode change (https://github.com/reddit/node-api-client/pull/103)

### DIFF
--- a/buildTasks/js.js
+++ b/buildTasks/js.js
@@ -56,7 +56,6 @@ module.exports = function buildJS(gulp, options) {
       .require('moment')
       .require('react')
       .require('react-dom')
-      .require('reddit-text-js')
       .require('superagent');
 
     bundler.add(entryFile);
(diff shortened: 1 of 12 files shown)


prashtx:
👓 @schwers @ajacksified 

Waiting on mweb staging, and then I'll push the corresponding version of Snoode to staging along with a verison of this reddit-mobile change that specifies Snoode by commit hash.

ajacksified:
🎉 :fish:

buildTasks/js.js
```diff
@@ -78,7 +77,7 @@ module.exports = function buildJS(gulp, options) {
     bundler
       .transform(babelify.configure({
         plugins,
-        ignore: /.+node_modules\/(moment|q|react|reddit-text-js|superagent|lodash|snuownd)\/.+/i,
+        ignore: /.+node_modules\/(moment|q|react|superagent|lodash|snuownd)\/.+/i,
```
schwers: can we remove `snuownd` as well?

schwers:
:fish:

PR #593, opened 2016-05-17
Enable shell-rendering

ajacksified:
👓 @schwers 

Depends on https://github.com/reddit/node-platform/pull/1

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@r/api-client": "~3.8.1",
     "@r/build": "^0.6.0",
     "@r/middleware": "0.5.4",
-    "@r/platform": "^0.1.0",
+    "@r/platform": "~0.2.0",
     "@r/private": "~0.1.0",
     "@r/redux-state-archiver": "~0.0.5",
     "atob": "2.0.3",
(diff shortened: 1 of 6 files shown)


src/app/actions/commentsPage.js
```diff
@@ -26,7 +26,9 @@ export const fetchCommentsPage = commentsPageParams => async (dispatch, getState
   const commentsPageId = paramsToCommentsPageId(commentsPageParams);
   const commentsPage = state.commentsPages[commentsPageId];
 
-  if (commentsPage) { return; }
+  const { shell } = state;
+
+  if (commentsPage || shell) { return; }
```
schwers: 🔕  I'm a little unsure if the data-fetching actions should take shell into account, versus the route handlers. The benefit would be data-fetching code could then be concerned only with how to fetch data, and less about 'when'. Then handlers could have better control of what happens when it's a shell page, say if we wanted partial shell'ing. But I don't have strong feelings either way

schwers:
Works, code looks good. I do have a question on https://github.com/reddit/node-platform/pull/1 that could impact this, so I'll hold the fish for now

ajacksified:
💇

ajacksified:
💇

schwers:
:fish:

PR #595, opened 2016-05-17
Add relevancy experiments

prashtx:
Support 3 relevancy variants (top, engaging, related) on the comment
threads for r/gaming. When a non-control variant is active, show only 3
comments (with an option to see the rest), and show 3 items of relevant
content below the comments.

Support "and" and explicit "or" feature config constructs, so that we
can condition on combinations of experiment variants, url flags,
subreddits, etc.

Support experiment variants in the feature flag component.

Add a RelevantContent component to encapsulate the content for the 3
variants.

We send bucketing events when determining which variant we're in, and we
send click events for clicks in the experiment area.


prashtx:
👓 @schwers @ajacksified

assets/less/components/RelevantContent.less
```diff
@@ -0,0 +1,150 @@
+.RelevantContent {
+  padding: @grid-size;
+  .dayMode & {
```
schwers: You can use the arguments `@inDayMode` and `@inNightMode` in the call to `themeify` below here. [example](https://github.com/reddit/reddit-mobile/blob/master/assets/less/components/stalagmite-stalactite.less#L20)

assets/less/components/RelevantContent.less
```diff
@@ -0,0 +1,150 @@
...
+    background-color: @pale-grey;
+  }
+  .nightMode & {
+    background-color: @black;
+  }
+
+  .themeify({
+    border-top: @theme-border;
+  });
+  .dayMode & .RelevantContent-action,
```
schwers: These styles don't look they depend on theme variables, could the the selector be simplified to `.RelevantContent-action` ?

assets/js/trackingEvents.es6.js
```diff
@@ -453,6 +455,70 @@ function trackingEvents(app) {
...
+      loid_created,
+      referer_page_type, // "listing", "link", "self", "comment"
+      // fullname (prefix+base-36 id) of the content at the target location
+      target_fullname,
+      target_url, // clickthrough url of the click
+      // human-readable name of the target (only applicable to subreddits)
+      target_name,
+      target_type, // "subreddit", "link", "self", "subscribe"
+      user_id, // base-10 user id
+      user_name // numan-readable name of the user sending the event
```
curioussavage: little typo here. like the comments 👍

assets/less/components/RelevantContent.less
```diff
@@ -0,0 +1,150 @@
...
+}
+
+.RelevantContent .Post,
+.RelevantContent .SearchPage__community {
+  border-width: 2px;
+  border-style: solid;
+  margin-bottom: @grid-size;
+  padding-left: 2 * @grid-size;
+
+  .dayMode & {
```
schwers: These could also be in the `@inNightMode` or `@inDayMode` blocks. Mostly a style thing, as I think its nice to have all the themes in `.themeify` for scoping, and then your styles don't need to be aware of the selectors for night vs day theme

src/routes.jsx
```diff
@@ -158,7 +159,7 @@ export function userData(ctx, app) {
 
   if (ctx.props.token) {
     const userOptions = Object.assign({}, apiOptions, {
-      user: 'me',
```
schwers: Just a reminder to go through and check for trailing commas, and other linter errors / warnings

assets/js/trackingEvents.es6.js
```diff
@@ -453,6 +455,70 @@ function trackingEvents(app) {
...
+      loid,
+      loid_created: loidcreated,
+      loid_new: false
+    };
+    
+    eventSend('bucketing_events', 'bucket', payload);
+  });
+
+  app.on('click:experiment', function (data) {
+    const loid = cookies.get('loid');
```
schwers: 🔕  Cookies.get doesn't seem like the best way to get this. It looks like `<RelevantContent />` is being rendered by the `<ListingPage />` which should be getting `loid` and `loidcreated` as props. It looks like that doens't happen when you're authenticated though, is that why this is cookies? If thats the case, do you have a sense of how hard it would be to pull in loid there?

src/featureflags.es6.js
```diff
@@ -7,6 +8,33 @@ const { BETA, SMARTBANNER } = constants.flags;
 const config = {
   [BETA]: true,
   [SMARTBANNER]: true,
+  experimentRelevancyTop: {
```
schwers: How would you feel about moving the experiment names into constants like `BETA` and `SMARTBANNER` ?

src/views/components/listings/RelevantContent.jsx
```diff
@@ -0,0 +1,238 @@
...
+      };
+      return (
+        <article ref='rootNode' className='Post' key={ id }>
+          <div className='Post__header-wrapper' onClick={ e => this.goToPost(e, url, name, i + 1) }>
+            <PostContent
+              post={ postWithFallback }
+              single={ false }
+              compact={ true }
+              expandedCompact={ false }
+              onTapExpand={ false }
```
schwers: This should be a function, should be safe to just turn this into `function()[}` (to avoid the linter warning for no binding in jsx props` or make a constant `const nop = () => {}`

src/views/components/listings/RelevantContent.jsx
```diff
@@ -0,0 +1,238 @@
...
+                url: c.url,
+                id: c.name,
+                name: c.title,
+                linkName: `${linkType} subreddit ${i + 1}`,
+                linkIndex: i + 1
+              };
+              return (
+                <div className='SearchPage__community' key={ c.id }>
+                  <div className='CommunityRow'>
+                    <a className='CommunityRow__icon' href='#' onClick={ e => this.goToSubreddit(e, clickData) }>
```
schwers: 🔕 If you want to have the styles as consistent as possible with Search results and topnav, there's a few more edge cases for rendering the subreddit icon. It might be easier to add an `onClick` param to `renderIcon` in [CommunityRow](https://github.com/reddit/reddit-mobile/blob/master/src/views/components/communities/CommunityRow.jsx#L10)

src/views/pages/listing.jsx
```diff
@@ -219,8 +233,39 @@ class ListingPage extends BasePage {
       draw a loading state.
     */
     if (Array.isArray(comments)) {
-      commentsList = comments.map((comment, i) => {
-        const key = `comment-${i}`;
+      let commentsTrees = comments;
+
+      function limitTrees(limit, trees) {
```
schwers: 🔕 This would be cool to have in `api-client@3x`'s comment tree utils

src/views/pages/listing.jsx
```diff
@@ -284,6 +337,28 @@ class ListingPage extends BasePage {
...
+          width={ constants.POST_DEFAULT_WIDTH }
+          subredditName={ subredditName }
+          subreddit={ subreddit }
+          listingId={ listingId }
+          isSelfText={ isSelfText }
+          app={ app }
+        />
+      );
+    } else {
+      relevantContent = (
```
schwers: Is there ever a case where the user isn't in a relevancy experiment?

schwers:
💅

assets/js/trackingEvents.es6.js
prashtx: that was a test for humans. you've passed.

src/views/pages/listing.jsx
prashtx: In case we're not in one of the variants, we still resolve the promise that backs `this.state.data.relevant` with an empty object. We don't know that we're in a null/control variant until the user data request completes, so `relevant` has to be asynchronous, and this lets us differentiate between "not yet known" (`relevant === undefined`) and "nothing interesting to render" (`relevant` is a  `{}`).

assets/js/trackingEvents.es6.js
schwers: :upvote:

src/views/pages/listing.jsx
schwers: Ah I see, okay.

assets/js/trackingEvents.es6.js
prashtx: It looks like we leave them out of the app context for logged-in users in appMethods.es6.js (https://github.com/reddit/reddit-mobile/blob/master/assets/js/appMethods.es6.js#L122), but I thought there was some issue for logged-out users, too. I'll try modifying that piece to see if it's an easy fix, otherwise it might become invasive.

src/views/components/listings/RelevantContent.jsx
prashtx: There were some intentional style differences between the below-content relevancy stuff and the search/topnav cousins. There's enough similarity to lean on the existing code and styles for this experiment, but I think we might want to refactor a bit as this experimentation progresses. Then we can adjust the presentation of the "relevant content" without stepping on the other pieces.

prashtx:
💇

assets/less/components/RelevantContent.less
```diff
@@ -0,0 +1,150 @@
...
+    border-top: @theme-border;
+  },
+  @inDayMode: {
+    background-color: @pale-grey;
+  },
+  @inNightMode: {
+    background-color: @black;
+  });
+
+  .RelevantContent-action {
```
schwers: 🔕 We've been trying to write any new LESS in BEM (https://css-tricks.com/forums/topic/less-bem/) when possible [PostHeader](https://github.com/reddit/reddit-mobile/blob/master/assets/less/components/listings/PostHeader.less) is a more complicated example. This looks fine for now, but just something that might be a useful reference.

assets/less/components/RelevantContent.less
```diff
@@ -0,0 +1,150 @@
...
+    margin-left: @grid-size;
+    margin-right: @grid-size;
+    padding-top: @grid-size;
+    padding-bottom: @grid-size;
+  }
+
+}
+
+.RelevantContent-header {
+  font-size: @font-size-h4;
```
schwers: 🔕 We've been moving away from the boostrap variables / styles as much as possible. This will probably have to change when it's converted to 2x

src/featureflags.es6.js
```diff
@@ -47,11 +80,77 @@ flags.addRule('url', function(query) {
   return Object.keys(parsedQuery).includes(query);
 });
 
+flags.addRule('and', (function (flags) {
```
schwers: 🔕 Nice, these are nice additions. Might be worth building it into node-flags

schwers:
:fish: :party:

prashtx:
👓 @schwers for the late-breaking data change, please (https://github.com/reddit/reddit-mobile/pull/595/commits/46016e777ce7d68451812b2d1a84c52cb1588205)

/cc @drunken-economist

schwers:
:fish:

prashtx:
6f9a5155c68e39303eceef59f4a420555686892b

PR #597, opened 2016-05-19
Ensure `subredditName` is supplied to `PostAndCommentList.jsx`

dwick:
Not having this property was causing the site to always be
`_FRONTPAGE_NAME` in `getSite`.

👓 @schwers


src/views/components/listings/Listing.jsx
```diff
@@ -22,6 +22,7 @@ export default class Listing extends BaseComponent {
     firstPage: T.number,
     page: T.number,
     hideSubredditLabel: T.bool,
+    subredditName: T.string.isRequired,
```
dwick: will non-subreddit listings not have this?

schwers:
:fish:

PR #598, opened 2016-05-20
Subreddit headers

schwers:
depends on https://github.com/reddit/node-api-client/pull/114/files

### DIFF
--- a/src/app/reducers/subscribedSubreddits.js
+++ b/src/app/reducers/subscribedSubreddits.js
@@ -7,7 +7,7 @@ import * as subscribedSubredditsActions from 'app/actions/subscribedSubreddits';
 import * as apiResponseActions from 'app/actions/apiResponse';
 import { newSubscribedSubredditsModel } from 'app/models/SubscribedSubreddits';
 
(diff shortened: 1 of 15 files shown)


schwers:
👓 @ajacksified

src/app/reducers/subscribedSubreddits.js
```diff
@@ -1,13 +1,43 @@
...
 
+const { SUBREDDIT } = models.ModelTypes;
 
 const DEFAULT = newSubscribedSubredditsModel();
 
+const updateStateFromSubreddits = (state, subreddits) => {
+  const keys = Object.keys(subreddits);
+  if (!keys.length) { return state; }
+
+  const nextSubreddits = Object.assign({}, state.subreddits);
```
ajacksified: Any reason in particular to not use `{ ...state.subreddits }` for brevity?

ajacksified:
Just a question, but :fish:

src/app/reducers/subscribedSubreddits.js
schwers: good catch, updating

PR #599, opened 2016-05-20
comment replies

ajacksified:
UI in place; API stubbed.

👓 @schwers

### DIFF
--- a/src/app/router/handlers/CommentsPage.js
+++ b/src/app/router/handlers/CommentsPage.js
@@ -1,7 +1,10 @@
 import { BaseHandler, METHODS } from '@r/platform/router';
+import * as platformActions from '@r/platform/actions';
 import { models } from '@r/api-client';
+
 import { cleanObject } from 'lib/cleanObject';
 import * as commentsPageActions from 'app/actions/commentsPage';
+import * as commentActions from 'app/actions/comment';
 import { fetchUserBasedData } from './handlerCommon';
 
 const { POST_TYPE } = models.ModelTypes;
@@ -36,6 +39,7 @@ export default class CommentsPage extends BaseHandler {
     });
   }
 
+
   async [METHODS.GET](dispatch, getState/*, utils*/) {
     const state = getState();
     if (state.platform.shell) { return; }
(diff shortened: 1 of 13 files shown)


src/app/router/handlers/CommentsPage.js
```diff
@@ -45,4 +49,30 @@ export default class CommentsPage extends BaseHandler {
...
+    if (!state.session.isValid) {
+      return dispatch(platformActions.setPage('/login'));
+    }
+
+    try {
+      dispatch(commentActions.reply(thingId, text));
+
+      // Go back to the state before the comment form was opened, if we can go
+      // back. Otherwise, redirect to the redirectUrl passed in.
+      if (history && state.platform.history.length > 2) {
```
ajacksified: I think we should build back functionality into platform that works on both server and client - which also means implementing referrer in platform.

src/app/router/handlers/CommentsPage.js
```diff
@@ -45,4 +49,30 @@ export default class CommentsPage extends BaseHandler {
...
+    if (!state.session.isValid) {
+      return dispatch(platformActions.setPage('/login'));
+    }
+
+    try {
+      dispatch(commentActions.reply(thingId, text));
+
+      // Go back to the state before the comment form was opened, if we can go
+      // back. Otherwise, redirect to the redirectUrl passed in.
+      if (history && state.platform.history.length) {
```
ajacksified: We should build `back` functionality into platform that works on both server and client - which also means implementing referrer in platform.

src/app/router/handlers/CommentsPage.js
```diff
@@ -45,4 +49,30 @@ export default class CommentsPage extends BaseHandler {
...
+        return history.go(-1);
+      } else {
+        // todo fix after implementing referrer
+        // dispatch(platformActions.navigateToUrl(METHODS.GET, '/'));
+      }
+    } catch (e) {
+      console.log('Error commenting');
```
ajacksified: We should also add error handling to future sprint planning; toasts would be nice.

src/app/router/handlers/CommentsPage.js
schwers: Agreed, Subreddit subscribe needs it as well

src/app/router/handlers/CommentsPage.js
```diff
@@ -45,4 +49,30 @@ export default class CommentsPage extends BaseHandler {
     dispatch(commentsPageActions.fetchCommentsPage(commentsPageParams));
     fetchUserBasedData(dispatch);
   }
+
+  async [METHODS.POST](dispatch, getState/*, utils*/) {
+    const { thingId, text } = this.bodyParams;
+
+    const state = getState();
+
+    if (!state.session.isValid) {
```
schwers: This check isn't quite right, `await`'s inside functions called in `dispatchBeforeSession` don't actually 'wait' before the handler is run. So you can have a valid `accessToken` but be in the middle of refreshing. This should probably use `waitForState`

src/app/router/handlers/CommentsPage.js
schwers: toasts++

schwers:
:fish:

PR #603, opened 2016-05-31
2X Subreddit sidebar and subnav. Converts lodash imports to easier syntax too

schwers:


schwers:
👓 @ajacksified

src/app/pages/SubredditAbout.jsx
```diff
@@ -0,0 +1,36 @@
...
+      href={ `/r/${subredditName}` }
+    >
+      { `r/${subredditName}` }
+    </BackAnchor>
+  </div>
+);
+
+export const SubredditAboutPage = props => {
+  const { subredditName } = props.urlParams;
+  console.log('subreddit page sees', subredditName, props);
```
ajacksified: 🚫

ajacksified:
A page you don't have access to - such as `http://localhost:8888/r/ajacksified_qa/about` - just shows the loading icon forever. Maybe a "Sorry, you cannot access this subreddit" message, with a "try logging in" link if you don't have a user?

We don't get info about whether it's because it's quarantined, private, or whatever, but that would be better than nothing.

Also, just the console.log above.

💅

schwers:
💇

src/app/reducers/subredditRequests.js
```diff
@@ -24,6 +24,13 @@ export default (state=DEFAULT, action={}) => {
       return merge(state, { [name]: { loading: false } });
     }
 
+    case subredditActions.FAILED_SUBREDDIT: {
+      const { name } = action;
+      const currentRequest = state[name];
+      if (!(currentRequest && currentRequest.loading)) { return state; }
+
+      return merge(state, { [name]: { loading: false, failed: true } });
```
ajacksified: how do you feel about making `failed` a typed error with a more descriptive message? `{ status: 403, message: 'unauthorized' }`, or something like that.

schwers: I like the idea, the status codes make me hesitant. Quarantined subreddits 404 (when they should maybe 403?) so it seems like for now we would still just be checking for the presence of an error. Would you be okay with me adding it in a separate commit that has error pages / "please try reloading" messages?

ajacksified:
:fish: - we can do better errors in a different pr

PR #605, opened 2016-05-31
Add a static json endpoint for supporting deep linking on iOS

uzi:
Also a quick update to the robots.txt -- blacklisting
another crawler to match what's on www.reddit.com.

:eyeglasses: @schwers @prashtx

### DIFF
--- a/src/server/routes.es6.js
+++ b/src/server/routes.es6.js
@@ -31,6 +31,10 @@ const serverRoutes = function(app) {
       User-agent: 008
       Disallow: /
 
+      # 80legs' new crawler
+      User-agent: voltron
+      Disallow: /
+
       User-Agent: bender
       Disallow: /my_shiny_metal_ass
 
--- a/src/server/routes.es6.js
+++ b/src/server/routes.es6.js
@@ -3,6 +3,24 @@ import crypto from 'crypto';
 
 import constants from '../constants';
 
+// Static file needed to support deep links on iOS
+const appleAppSiteAssociation = JSON.stringify({
+  webcredentials: {
+    apps: [
+      '2TDUX39LX8.com.reddit.Reddit',
+    ],
+  },
+  applinks: {
+    apps: [],
+    detail: [
+      {
+        appID: '2TDUX39LX8.com.reddit.Reddit',
+        paths: [ '/', '/r/*', '/u/*', '/user/*' ],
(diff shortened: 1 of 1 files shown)


uzi:
This is for https://reddit.atlassian.net/browse/MOBILEWEB-427

src/server/routes.es6.js
```diff
@@ -3,6 +3,25 @@ import crypto from 'crypto';
 
 import constants from '../constants';
 
+// Static file needed to support deep links on iOS
+const appleAppSiteAssociationObj = {
+  'webcredentials': {
```
prashtx: 🔕 We don't need quotes around the keys here, but apparently our lint rules don't care, so I suppose it's not a big deal.

src/server/routes.es6.js
```diff
@@ -36,6 +59,11 @@ const serverRoutes = function(app) {
     `;
   });
 
+  router.get('/apple-app-site-association', function * () {
+    this.body = appleAppSiteAssociation;
+    this.type = 'application/json';
```
prashtx: Will this add the `charset=utf-8` to the content type? If not, maybe it's better to set `this.body` to the object and let the rest of the stack take care of stringification/content-type/etc. Presumably that also gives us control over having pretty-print stringification in dev and concise stringification in production.

prashtx:
💅

src/server/routes.es6.js
schwers: 🔕 Style wise, we normally don't avoid quotes around object keys (e.g.)[https://github.com/reddit/reddit-mobile/blob/master/src/propTypes.es6.js]. I'll defer to you're thinking on this because it might be something we update often via copy-paste new versions, in which case keeping quotes is probably easier.

uzi:
:haircut:

prashtx:
🐟

schwers:
:fish:

src/server/routes.es6.js
```diff
@@ -13,6 +31,10 @@ const serverRoutes = function(app) {
       User-agent: 008
       Disallow: /
 
+      # 80legs' new crawler
```
kjoconnor: Can you make this a separate commit?

uzi: Sure

PR #607, opened 2016-06-01
Reducer tests

phil303:
Pretty self explanatory
:eyeglasses: @schwers

### DIFF
--- a/src/app/actions/accounts.js
+++ b/src/app/actions/accounts.js
@@ -7,8 +7,8 @@ const { AccountsEndpoint } = endpoints;
 export const FETCHING_ACCOUNT = 'FETCHING_ACCOUNT';
 export const fetching = (options) => ({ type: FETCHING_ACCOUNT, ...options });
 
-export const RECEIEVED_ACCOUNT = 'RECEIEVED_ACCOUNT';
-export const receieved = (options, result) => ({ type: RECEIEVED_ACCOUNT, ...options, result });
+export const RECEIVED_ACCOUNT = 'RECEIVED_ACCOUNT';
+export const received = (options, result) => ({ type: RECEIVED_ACCOUNT, ...options, result });
 
 export const fetch = (options) => async (dispatch, getState) => {
   const { name, loggedOut } = options;
@@ -24,5 +24,5 @@ export const fetch = (options) => async (dispatch, getState) => {
 
   const { results } = apiResponse;
   const result = results.length ? results[0] : {};
-  dispatch(receieved(options, result));
+  dispatch(received(options, result));
 };
--- a/src/app/reducers/accountRequests.test.js
+++ b/src/app/reducers/accountRequests.test.js
@@ -1,4 +1,5 @@
 import createTest from '@r/platform/createTest';
+
 import accountRequests from './accountRequests';
 import * as accountActions from 'app/actions/accounts';
 import * as loginActions from 'app/actions/login';
--- a/src/app/reducers/accountRequests.test.js
+++ b/src/app/reducers/accountRequests.test.js
@@ -1,0 +1,57 @@
+import createTest from '@r/platform/createTest';
+import accountRequests from './accountRequests';
+import * as accountActions from 'app/actions/accounts';
+import * as loginActions from 'app/actions/login';
+
+const REQUIRED_KEYS = ['id', 'loading', 'result'];
+
+createTest({ reducers: { accountRequests } }, ({ getStore, expect }) => {
+  describe('accountRequests', () => {
+
+    describe('LOGGED_IN and LOGGED_OUT', () => {
+      it('should return default on log in', () => {
+        const { store } = getStore();
+        store.dispatch(loginActions.loggedIn());
+
+        const { accountRequests } = store.getState();
+        expect(accountRequests).to.eql({});
+      });
+
+      it('should return default on log out', () => {
+        const { store } = getStore();
+        store.dispatch(loginActions.loggedOut());
+
+        const { accountRequests } = store.getState();
+        expect(accountRequests).to.eql({});
+      });
+    });
+
+    describe('FETCHING_ACCOUNT', () => {
+      it('should add an account optimistically', () => {
+        const ACCOUNT = { id: 'foobar', loading: true, result: {} };
+
+        const { store } = getStore();
+        store.dispatch(accountActions.fetching({ name: 'foobar' }));
+
+        const { accountRequests } = store.getState();
+        expect(accountRequests).to.have.keys('foobar');
+        expect(accountRequests.foobar).to.have.all.keys(REQUIRED_KEYS);
+        expect(accountRequests.foobar).to.eql(ACCOUNT);
+      });
+    });
+
+    describe('RECEIVED_ACCOUNT', () => {
+      it('should update an account when request if finished', () => {
+        const RESULT = { type: 'account', uuid: 'me' };
+
+        const { store } = getStore();
+        store.dispatch(accountActions.received({ name: 'foobar' }, RESULT));
+
+        const { accountRequests } = store.getState();
+        expect(accountRequests).to.have.keys('foobar');
+        expect(accountRequests.foobar).to.have.all.keys(REQUIRED_KEYS);
+        expect(accountRequests.foobar.result).to.eql(RESULT);
+      });
+    });
+  });
+});
--- a/src/app/reducers/collapsedComments.test.js
+++ b/src/app/reducers/collapsedComments.test.js
@@ -1,0 +1,59 @@
+import createTest from '@r/platform/createTest';
+
+import collapsedComments from './collapsedComments';
+import * as commentActions from 'app/actions/comment';
+import * as loginActions from 'app/actions/login';
+
+createTest({ reducers: { collapsedComments } }, ({ getStore, expect }) => {
+  describe('collapsedComments', () => {
+    describe('LOGGED_IN and LOGGED_OUT', () => {
+      it('should return default on log in', () => {
+        const { store } = getStore();
+        store.dispatch(loginActions.loggedIn());
+
+        const { collapsedComments } = store.getState();
+        expect(collapsedComments).to.eql({});
+      });
+
+      it('should return default on log out', () => {
+        const { store } = getStore();
+        store.dispatch(loginActions.loggedOut());
+
+        const { collapsedComments } = store.getState();
+        expect(collapsedComments).to.eql({});
+      });
+    });
+
+    describe('TOGGLE_COLLAPSE', () => {
+      const ID = '1';
+
+      it('should set a comments collapse state to true when the actions collapse is true', () => {
+        const { store } = getStore();
+        store.dispatch(commentActions.toggleCollapse(ID, true));
+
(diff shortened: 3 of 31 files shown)


src/app/reducers/commentsPages.test.js
```diff
@@ -0,0 +1,182 @@
...
+          params: { id: POST_ID },
+          postId: POST_ID,
+          loading: true,
+          loadingMoreChildren: {},
+          results: [],
+        };
+
+        // inject the cached comments page into the store ahead of time
+        const { store } = getStore();
+        store.getState().commentsPages = { '1': CACHED_COMMENTS_PAGE };
```
nramadas: couldn't you use the state initialization of getStore here?

phil303: This is a fairly invasive test and i'm not completely sold we should keep it . It's testing to make sure the reducer checks for cached data and doesn't do any unnecessary updates to the store. Therefore it's checking for identity - mocha's `equal` instead of value - mocha's `eql`.

phil303: And now that I think about it I can do it that way. I'll update it as such but I'm hoping someone convinces me this is a dumb test too

src/app/reducers/commentsPages.test.js
```diff
@@ -0,0 +1,182 @@
...
+        const LOADING_COMMENTS_PAGE = {
+          id: '1',
+          params: { id: POST_ID },
+          postId: POST_ID,
+          loading: true,
+          loadingMoreChildren: {},
+          results: [],
+        };
+        const RESULTS = [
+          { type: 'comment', uuid: 't1_comment1', paginationId: 't1_comment1' },
```
ajacksified: 🔕 it may be good to use a real `Record` object for these, so that as the API evolves we don't break all of our tests. It's also a slightly more concise syntax:

```
const RESULTS = [
  new Record('comment', 't1_comment1', 't1_comment1'),
  ...
];
```

We could be _even more_ pedantic and use the `TYPES` from https://github.com/reddit/node-api-client/blob/2ca06bf4ebf7bc7a15cd08c07cea897e1f3c1dff/apis/base.es6.js#L13 instead of hardcoding `t1`, but that's probably a bit over the top for something that's probably never going to change, unless you wanted to implement a `stubRecord(type, id=fakeid, paginationId=null)` method to build fake stubs in api-client.

src/app/reducers/accountRequests.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+        expect(accountRequests).to.eql({});
+      });
+    });
+
+    describe('FETCHING_ACCOUNT', () => {
+      it('should add an account optimistically', () => {
+        const ACCOUNT = { id: 'foobar', loading: true, result: {} };
+        const { store } = getStore();
+        store.dispatch({
```
ajacksified: [prefer newlines around multiline blocks](https://github.com/reddit/styleguide/tree/master/javascript#18.7) when not the first block of its indentation level

src/app/reducers/commentsPages.test.js
```diff
@@ -0,0 +1,182 @@
+import createTest from '@r/platform/createTest';
+import merge from '@r/platform/merge';
+
+import commentsPages from './commentsPages';
+import * as commentsPageActions from 'app/actions/commentsPage';
+import * as loginActions from 'app/actions/login';
+import * as apiResponseActions from 'app/actions/apiResponse';
+
+
```
ajacksified: extra newline 😱

src/app/reducers/accountRequests.test.js
phil303: oh interesting. Will update!

src/app/reducers/commentsPages.test.js
schwers: FWIW `paginationId` is defaulted to `uuid` (src)[https://github.com/reddit/node-api-client/blob/3X/apiBase/Record.es6.js] so if you were using `Record`s it would be even more concise. e.g.
```javascript
const RESULTS = [
  new Record('comment', 't1_comment'),
  ...
];
```

src/app/reducers/commentsPages.test.js
phil303: oh yea there's a few of these. That's a personal habit of mine to separate the imports from the business logic. I can remove.

src/app/reducers/collapsedComments.test.js
```diff
@@ -0,0 +1,57 @@
+import createTest from '@r/platform/createTest';
+
+import collapsedComments from './collapsedComments';
+import * as commentActions from 'app/actions/comment';
+import * as loginActions from 'app/actions/login';
+
+createTest({ reducers: { collapsedComments } }, ({ getStore, expect }) => {
+  describe('collapsedComments', () => {
+    describe('LOGGED_IN and LOGGED_OUT', () => {
+      it('should return default on log in', () => {
```
ajacksified: What is the purpose of these loggedin/out tests? The collapse reducer isn't affected by login state.

src/app/reducers/commentsPages.test.js
phil303: Yea. Using `Record` is the right approach - a lot of this was an exercise in understanding our codebase and there's probably a few spots like this. I'll update.

src/app/reducers/commentsPages.js
```diff
@@ -29,16 +29,18 @@ export default (state=DEFAULT, action={}) => {
...
-    case commentsPageActions.RECEIEVED_COMMENTS_PAGE: {
+    case commentsPageActions.RECEIVED_COMMENTS_PAGE: {
       const { commentsPageId, commentsPageResults } = action;
       const currentCommentsPage = state[commentsPageId];
       if (!currentCommentsPage) { return state; }
 
       return merge(state, {
         [commentsPageId]: {
           loading: false,
+          // TODO: what happens if a user adds a comment before results come back?
```
ajacksified: Good point - we should probably use the `concat` strategy of `merge` here: https://github.com/reddit/node-platform/blob/master/lib/modules/merge.js

src/app/reducers/commentsPages.js
```diff
@@ -29,16 +29,18 @@ export default (state=DEFAULT, action={}) => {
...
           results: commentsPageResults,
         },
+        // TODO: this feels inherently race condition-ish
```
ajacksified: agreed, this is a hack (I wrote it) - it would be better to send the current comment page ID with the newModel action instead, so it can update its page.

src/app/reducers/collapsedComments.test.js
phil303: uh, you're right. i can't figure out where the heck I got this.

src/app/reducers/commentsPages.test.js
```diff
@@ -0,0 +1,182 @@
...
+        expect(commentsPages).to.eql({});
+      });
+    });
+
+    describe('FETCHING_COMMENTS_PAGE', () => {
+      it('should try the cache for a pre-existing commentsPage first', () => {
+        const POST_ID = 't3_12345';
+        const CACHED_COMMENTS_PAGE = {
+          id: '1',
```
ajacksified: it would be best to put the `'1'` into a const and reference it everywhere below (for this and all tests where data such as ids are re-used), so that tests are more refactorable

src/app/reducers/compact.test.js
```diff
@@ -0,0 +1,43 @@
+import createTest from '@r/platform/createTest';
+
+import * as compactAction from 'app/actions/compact';
+import * as loginActions from 'app/actions/login';
+import compact from './compact';
+
+createTest({ reducers: { compact } }, ({ getStore, expect }) => {
+  describe('compact', () => {
+    describe('LOGGED_IN and LOGGED_OUT', () => {
```
ajacksified: (like my other comment, why logged in/out?)

src/app/reducers/collapsedComments.test.js
schwers: Technically it does (https://github.com/reddit/reddit-mobile/commit/f142b4593de4a7de9511f7f57bfd826704db87db). When I did auth stuff I added the log in / out clearing to all of the reducers that were there. In this case I could see the argument for preserving what's collapsed when you log-in. But clearing on log-out seems like the right idea to me personally, as I think it would be weird to log-out and see any remnants of what I did logged in other than links being a visited color (as that's standard browser behavior).

ajacksified: Oh, you're right - I didn't notice that when looking through the code. Disregard!

src/app/actions/accounts.js
```diff
@@ -7,8 +7,8 @@ const { AccountsEndpoint } = endpoints;
 export const FETCHING_ACCOUNT = 'FETCHING_ACCOUNT';
 export const fetching = (options) => ({ type: FETCHING_ACCOUNT, ...options });
 
-export const RECEIEVED_ACCOUNT = 'RECEIEVED_ACCOUNT';
```
schwers: 🔕 Well, this is embarrassing. My editor tends to aggressively autocomplete across project files so I must have spelled received incorrectly early on and and just blissfully autocompleted it in multiple places. Thanks for fixing this!

ajacksified: never had that problem with vim :trollface:

src/app/reducers/collapsedComments.test.js
phil303: ah thank god. thought i was losing it

src/app/reducers/commentsPages.test.js
```diff
@@ -0,0 +1,182 @@
...
+          loadingMoreChildren: {},
+          results: [],
+        };
+
+        const { store } = getStore({
+          commentsPages: { '1': CACHED_COMMENTS_PAGE },
+        });
+
+        store.dispatch({
+          type: commentsPageActions.FETCHING_COMMENTS_PAGE,
```
schwers: Testing philosophy question: is there are reason to build the actions manually rather than use the action creators? e.g.
```javascript
store.dispatch(commentPageActions.fetching('1', { id: POST_ID }));
``` 
It seems like if we used the creators and then had separates tests to verify they're building the correct action objects the tests would be more concise and refactorable.

ajacksified: :+1:

phil303: Yea that's a good point. It'd exercise more code as well and that's only a good thing. TBH, I just started writing the tests and didn't slow down to think about that.

src/app/reducers/commentsPages.js
phil303: @ajacksified Not a huge fan of leaving TODOs in code but I want to address this eventually and not in this patch. I can remove and add it to my own todo list or I can leave it. Do you have a preference here?

phil303:
:haircut:

src/app/reducers/commentsPages.js
ajacksified: Yeah, a fix should be a separate PR.

ajacksified:
looks good to me. 🐟

thanks for splitting each reducer test into a separate commit! makes reviewing very nice.

PR #608, opened 2016-06-02
Add next-content experiments

prashtx:
Support 4 next-content variants (middle carousel, bottom, banner, top
3). The "middle" variant shows a swipeable set of the top unread posts
from the current subreddit, positioned between the current post and the
comments. The "bottom" variant shows the top unread post from the
current subreddit, fixed to the bottom of the viewport. The "banner"
variant behaves like "bottom" but appears only after the user scrolls
and sits just above the bottom of the viewport. The "top3" variant is
identical to the "top3" variant of the relevancy_mweb experiment. The
experiment is not restricted to a particular subreddit.

Track the 10 most recently viewed posts, so that we can show new content
to logged-out and logged-in users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "react-addons-test-utils": "^15.0.1",
     "react-dom": "^15.0.1",
     "react-motion": "0.4.2",
+    "react-touch": "0.4.0",
     "scmp": "1.0.0",
     "semver": "5.0.3",
     "simple-oauth2": "0.2.1",
--- a/src/featureflags.es6.js
+++ b/src/featureflags.es6.js
@@ -8,6 +8,10 @@ const {
   VARIANT_RELEVANCY_TOP,
   VARIANT_RELEVANCY_ENGAGING,
   VARIANT_RELEVANCY_RELATED,
+  VARIANT_NEXTCONTENT_BOTTOM,
+  VARIANT_NEXTCONTENT_MIDDLE,
+  VARIANT_NEXTCONTENT_BANNER,
+  VARIANT_NEXTCONTENT_TOP3,
 } = constants.flags;
 
 const config = {
@@ -42,6 +46,34 @@ const config = {
         variant: 'relevancy_mweb:related',
       }],
     },
+  },
+  [VARIANT_NEXTCONTENT_BOTTOM]: {
+    url: 'experimentnextcontentbottom',
+    and: [{
+      variant: 'nextcontent_mweb:bottom',
+      loggedin: false,
+    }],
+  },
+  [VARIANT_NEXTCONTENT_MIDDLE]: {
+    url: 'experimentnextcontentmiddle',
+    and: [{
+      variant: 'nextcontent_mweb:middle',
+      loggedin: false,
+    }],
+  },
+  [VARIANT_NEXTCONTENT_BANNER]: {
+    url: 'experimentnextcontentbanner',
+    and: [{
+      variant: 'nextcontent_mweb:banner',
+      loggedin: false,
+    }],
+  },
+  [VARIANT_NEXTCONTENT_TOP3]: {
+    url: 'experimentnextcontenttop3',
+    and: [{
+      variant: 'nextcontent_mweb:top3',
+      loggedin: false,
+    }],
   },
 };
 
@@ -111,7 +143,7 @@ flags.addRule('userAgentSubstr', function(agents) {
 });
 
 flags.addRule('subreddit', function (name) {
-  return this.props.subredditName && this.props.subredditName === name;
+  return this.props.subredditName && this.props.subredditName.toLowerCase() === name.toLowerCase();
 });
 
 flags.addRule('variant', function (name) {
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -33,6 +33,7 @@ import { SORTS } from './sortValues';
 import isFakeSubreddit, { randomSubs } from './lib/isFakeSubreddit';
 import makeRequest from './lib/makeRequest';
 import features from './featureflags';
+import { getVisitedPosts, setVisitedPosts } from './lib/visitedPosts';
 
 const config = defaultConfig;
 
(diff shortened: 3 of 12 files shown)


prashtx:
I'm still waiting on some event payload details and feedback from design, but I wanted to put this out for initial review. Temporary spots are marked with `XXX` (feature flag details to allow easier review; event payload values).

👓 @ajacksified @uzi

prashtx:
Working with @phil303 to resolve the peer dependency conflict that results from my inclusion of react-touch.

uzi:
:eyeglasses: @schwers -- adding a pair of eyes to replace Jack's

src/featureflags.es6.js
```diff
@@ -112,7 +148,7 @@ flags.addRule('userAgentSubstr', function(agents) {
 });
 
 flags.addRule('subreddit', function (name) {
-  return this.props.subredditName && this.props.subredditName === name;
+  return this.props.subredditName && this.props.subredditName.toLowerCase() === name.toLowerCase();
```
schwers: 👍  Good idea

src/routes.jsx
```diff
@@ -428,6 +430,20 @@ function routes(app) {
       commentId: ctx.params.commentId,
     });
 
+    if (localStorageAvailable()) {
+      const visitedString = global.localStorage.getItem(constants.VISITED_POSTS_KEY);
```
schwers: Could you move this into a helper function, something like `recordVisitedPost(ctx.params.listingId)`? These routes are getting pretty big. Might be worth while to have helpers for `getVisitedPosts(): -> [PostId]` or `setVisitedPosts(posts:[PostId])`.  Should we be clearing these on log-out?

src/views/components/listings/RelevantContent.jsx
```diff
@@ -37,11 +55,69 @@ export default class RelevantContent extends BaseComponent {
...
+      if (this.props.feature.enabled(VARIANT_NEXTCONTENT_BANNER)) {
+        this.props.app.off(constants.SCROLL, this.maybeShowBanner);
+      }
+    }
+  }
+
+  maybeShowBanner() {
+    if (window && window.scrollY > this.scrollY) {
+      this.props.app.off(constants.SCROLL, this.maybeShowBanner);
+      window.requestAnimationFrame(() => {
```
schwers: 🔕 I was under the impression react batched state updates to animation frames and this wasn't necessary. Was there a performance boost or some other reason I'm overlooking for this? (Also do we have a polyfill for `requestAnimationFrame`?)

src/views/components/listings/RelevantContent.jsx
```diff
@@ -64,19 +140,20 @@ export default class RelevantContent extends BaseComponent {
...
+      eventType,
       loid,
       loidcreated,
-      experimentName: 'relevancy_mweb',
+      experimentName,
       linkIndex,
-      linkName: `top post ${linkIndex}`,
-      refererPageType: isSelfText ? 'self' : 'link',
-      refererUrl: window.location.href,
+      // linkName: `top post ${linkIndex}`,
```
schwers: Is this comment showing what `linkName` looks like? Maybe un-needed now?

src/views/components/listings/RelevantContent.jsx
phil303: Yea, I don't think this will work as intended. `setState` is potentially asynchronous as it tries to batch updates.

phil303: https://facebook.github.io/react/docs/component-api.html

prashtx: This is probably unnecessary. As I coded this up, there was a chance more stuff would happen on scroll, so `requestAnimationFrame` would've been handy. As it stands now, I think we can cut it.

src/featureflags.es6.js
```diff
@@ -40,6 +44,38 @@ const config = {
...
     },
   },
+  [VARIANT_NEXTCONTENT_BOTTOM]: {
+    url: 'experimentnextcontentbottom',
+    subreddit: 'images',
+    and: [{
+      variant: 'nextcontent_mweb:bottom',
+      loggedin: false,
+    }],
+  }, // XXX
```
schwers: Why `XXX`?

src/routes.jsx
prashtx: :+1: on moving to a helper.

Desktop seems to retain recently-visited subs on logout, so I think it's reasonable to retain recent posts. Since we use them to show the user new content, it'll be helpful behavior.

src/featureflags.es6.js
prashtx: The specific feature-flag configs are temporary, so I can demonstrate the different variants more easily (deployed as such to mweb-staging). The actual experiment will target logged-out users across all communities. `XXX` is just an easy way to have the linter/silver search help me find all these pre-deploy spots. There's another one for a value in the event payload, with is TBD.

schwers: Ah, neat. I've always used `TODO`, didn't realize `XXX` was a similar convention and also gets highlighted / shows up in linters. 👍

src/routes.jsx
schwers: Okay cool

src/views/pages/listing.jsx
```diff
@@ -243,7 +243,26 @@ class ListingPage extends BasePage {
...
+       this.state.feature.enabled(constants.flags.VARIANT_NEXTCONTENT_TOP3));
+
+    let footerClass = '';
+    if (this.state.feature.enabled(constants.flags.VARIANT_NEXTCONTENT_BOTTOM) ||
+        this.state.feature.enabled(constants.flags.VARIANT_NEXTCONTENT_BANNER)) {
+      footerClass = 'with-footer-nextcontent';
+    }
+
+    const relevantBottom =
+      (this.state.feature.enabled(constants.flags.VARIANT_RELEVANCY_TOP) ||
```
schwers: 🔕 Maybe we should add `feature.anyEnabled` to `node-flags` or do something like `some(map(this.state.feature.enabled([...features...])))`

prashtx: Good suggestion. I'm using the `some([...features...], variant => feature.enabled(variant))` approach for the 2X port.

prashtx:
💇

schwers:
:fish:

package.json
```diff
@@ -51,6 +51,7 @@
     "react-addons-test-utils": "^15.0.1",
     "react-dom": "^15.0.1",
     "react-motion": "0.4.2",
+    "react-touch": "0.4.0",
```
schwers: 🔕 Just small reminder to update deps when this goes out

PR #610, opened 2016-06-06
Remove country filter from the smart banner

uzi:
Native mobile is launched in all countries now,
so remove the country filter as it's no longer needed.


phil303:
:fish:

schwers:
:fish:

PR #613, opened 2016-06-09
Add some metadata routes from 1X

prashtx:
We need to support `/robots.txt`, `/routes`, and the apple app site
assocation routes.


prashtx:
👓 @schwers, @nramadas

schwers:
:fish:

PR #614, opened 2016-06-09
Add feature flag support

prashtx:
Add support for feature flags that are currently computed as a function of state.

Outside of views, `app/featureFlags` can be used directly. For views, `app/selectors/features` is a small wrapper to help get the feature flag component into the props.

We need to wire the variant evaluation up to bucketing events when the event system is ready. Because we want to check the variants and trigger the events as late as possible (as close to when the user sees a different behavior, and only when the user would be presented with experimental behavior), we need to avoid precomputing variants and making them part of the state.

### DIFF
--- a/src/app/featureFlags.js
+++ b/src/app/featureFlags.js
@@ -1,0 +1,73 @@
+import Flags from '@r/flags';
+
+import getSubreddit from 'lib/getSubredditFromState';
+
+import { flags as flagConstants } from './constants';
+
+const {
+  BETA,
+  SMARTBANNER,
+} = flagConstants;
+
+const config = {
+  [BETA]: true,
+  [SMARTBANNER]: true,
+};
+
+const flags = new Flags(config);
+
+function extractUser(ctx) {
+  const { state } = ctx;
+  if (!state || !state.user || !state.accounts) {
+    return;
+  }
+  return state.accounts[state.user.name];
+}
+
+flags.addRule('loggedin', function(val) {
+  return (!!this.state.user && !this.state.user.loggedOut) === val;
+});
+
+flags.addRule('users', function(users) {
+  const user = extractUser(this);
+  return users.includes(user.name);
+});
+
+flags.addRule('employee', function(val) {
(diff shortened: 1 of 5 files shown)


prashtx:
👓 @schwers @nramadas

src/app/featureFlags.js
```diff
@@ -0,0 +1,73 @@
...
+
+flags.addRule('url', function(query) {
+  // turns { feature_thing: true, wat: 7 } into { thing: true }
+  const parsedQuery = Flags.parseConfig(this.state.platform.currentPage.queryParams);
+  return Object.keys(parsedQuery).includes(query);
+});
+
+flags.addRule('subreddit', function (name) {
+  const subreddit = getSubreddit(this.state);
+  if (!subreddit) {
```
nramadas: if `getSubreddit` returned an empty string instead of undefined, you wouldn't need this check here.

src/lib/getSubredditFromState.js
```diff
@@ -0,0 +1,14 @@
+export default function getSubreddit(state) {
+  if (state.platform.currentPage.urlParams.subredditName) {
+    return state.platform.currentPage.urlParams.subredditName;
+  }
+
+  const { commentsPages = {}} = state;
+  const { current } = commentsPages;
+  if (!current) { return; }
+  const results = commentsPages[current].results;
```
nramadas: lodash's `has` could clean some of this up:
```
if (!has(state, ['commentsPages', 'current'])) {
  return '';
}

if (!has(state, ['commentsPages', state.commentsPages.current, 'results'])) {
  return '';
}
```

nramadas:
:haircut:

nramadas:
:nail_care:

src/app/featureFlags.js
prashtx: Discussed IRL. Decided to leave explicit non-string return type but to use `null` instead of `undefined`.

prashtx:
💇

nramadas:
:fish:

PR #616, opened 2016-06-10
Avoid showing replies in inbox mode

prashtx:
The mobile web messages view has an "All" view, which should not show a
user's own replies, and a "Messages" view, which should show a user's
reply (nested under the original message). The user can compose a reply
from both views, but we should only add the submitted message to the
view if we're in the "Messages" view.


prashtx:
👓 @schwers @nramadas

curioussavage:
🐟  for me. not ideal but with niranjan implementing the new designs it seems fine

PR #617, opened 2016-06-10
Default to compact mode on the client

phil303:
We're only defaulting to compact mode when the user comes in through reddit.com (via a url parameter) but not when they come in directly through m.reddit.com.

:eyeglasses: @schwers

### DIFF
--- a/src/server/index.jsx
+++ b/src/server/index.jsx
@@ -93,7 +93,8 @@ function setCompact(ctx, app) {
     return;
   }
 
-  let compact = (ctx.cookies.get('compact') || '').toString() === 'true';
+  // default to compact mode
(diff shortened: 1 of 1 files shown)


uzi:
:fish:

uzi:
The only thing I can see with changing this is to do a 99% experiment ... if we wanted to go that route here as well.  But fine to swing this to 100% and then scale back the 1% if we're running the experiment than to hold off.

schwers:
:fish:

schwers:
ping @phil303 should we land this patch soon?

phil303:
@schwers This patch should be closed. Easier to explain in person

PR #620, opened 2016-06-14
Fix comment caron icon vertical alignment

schwers:
This fixes the comment caron icon being a few pixels off vertically

Before
image

After
image

👓 @phil303 or @nramadas


phil303:
:fish:

PR #623, opened 2016-06-20
Comment page: disable reply for banned users

curioussavage:
Also refactored to show the same alert used to notify users
that they are on a permalink page for banned users and when
the post is locked or archived

👓  @schwers 

I saw [this](https://www.reddit.com/r/mobileweb/comments/4ovwre/better_handling_of_subreddit_bans_and_disabled/) in the channel last night and it looked easy so here we are.

As you can see here in the screenshot I changed it to use the same alert we use for permalink pages. I think it is much more noticeable. Let me know what you think.
screen shot 2016-06-20 at 11 36 33 am

### DIFF
--- a/src/views/components/LinkTools.jsx
+++ b/src/views/components/LinkTools.jsx
@@ -126,6 +126,8 @@ export default class LinkTools extends React.Component {
   }
 
   renderTools() {
+    const text = this.props.disableReply ? '' : 'Write a comment';
+
     return (
       <div className='LinkTools__tools'>
         { this.renderSort() }
@@ -133,7 +135,7 @@ export default class LinkTools extends React.Component {
           className='LinkTools__comment'
           onClick={ this.toggleForm }
         >
-          Write a comment
+          { text }
         </div>
       </div>
     );
--- a/src/views/components/LinkTools.jsx
+++ b/src/views/components/LinkTools.jsx
@@ -47,8 +47,7 @@ export default class LinkTools extends React.Component {
   }
 
   toggleForm() {
-    if (this.props.isArchived) { return; }
-    if (this.props.isLocked) { return; }
+    if (this.props.disableReply) { return; }
     if (this.props.app.needsToLogInUser()) { return; }
 
     this.setState({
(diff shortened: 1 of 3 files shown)


src/views/pages/listing.jsx
```diff
@@ -195,6 +195,42 @@ class ListingPage extends BasePage {
...
+      message = (
+        <p>
+          <span>You are viewing a single comment's thread. </span>
+          <a href={ listing.permalink }>View the rest of the comments</a>
+        </p>
+      );
+    }
+
+    if (subreddit && subreddit.user_is_banned) {
+      message = <p>You are banned from commenting</p>;
```
schwers: 🔕 Should we mention "from this subreddit" ?

schwers: 🔕 or "community"

src/views/components/LinkTools.jsx
```diff
@@ -127,22 +126,14 @@ export default class LinkTools extends React.Component {
...
-
     return (
       <div className='LinkTools__tools'>
         { this.renderSort() }
         <div
           className='LinkTools__comment'
           onClick={ this.toggleForm }
         >
-          { text }
+          Write a comment
```
schwers: I'm not sure `Write a comment` is the best copy to use in all cases. Now that the alert is handling the messaging for why you can't comment, maybe this copy shouldn't be here when commenting is disabled? It also might just seem strange to me, what do you think?

src/views/pages/listing.jsx
```diff
@@ -409,14 +440,15 @@ class ListingPage extends BasePage {
...
+                disableReply={ disableReply }
                 isArchived={ listing.archived }
                 isLocked={ listing.locked }
                 sort={ sort }
                 onNewComment={ this.onNewComment }
                 onSortChange={ this.handleSortChange }
               />
             </div>
-            { singleComment }
+            { threadNotification }
```
schwers: This name is so much better, thanks! 💯

schwers:
💅

src/views/components/LinkTools.jsx
curioussavage: I think I agree. especially since we don't present it like a button I see no problem making it invisible since we would probably want to make it look disabled if it looked like a button. hiding it seems like the best thing

curioussavage:
💇

src/views/components/comment/Comment.jsx
```diff
@@ -584,6 +586,16 @@ export default class Comment extends BaseComponent {
     );
   }
 
+  renderBannedReplyForm() {
+    return (
+      <div className='Comment__replyForm'>
+        <div className='Comment__replyFormLocked'>
+          You are currently banned
```
schwers: Should this mention community as well?

schwers:
:fish:

PR #624, opened 2016-06-21
Refactor data fetching, move replies, add better tests

schwers:
In current 2X code, async actions that fetch data from the api generally follow this pattern:
```javascript
const fetch = params => async (dispatch, getState) => {
  const id = paramsToId(params);
  const currentRequest = getState().someRequests[id];
  if (currentRequest) { return; }

  dispatch(fetching(id));

  const res = await Endpoint.get(apiOptionsFromState(getState(), params);
  dispatch(receivedApiResponse(res));
  dispatch(received(id, res.results));
};
```

This last bit of dispatching multiple actions isn't quite idiomatic redux and due to a lack of batched actions causes us to run selectors (and possibly re-render) twice. The largest thing this patch does is refactor away `dispatch(receievedAPIResponse(res))`. In doing so some of the naming conventions in action files are made more consistent, the generalized `reducer/apiResponse.js` is removed, in its place are reducers and tests for each specific type of Model (e.g. `posts.js`). 

There's also some logical cleanup of where two actions live: toggling subreddit subscription now lives in `action/subscribedSubreddits.js` with the actions to fetch all of your subscribed subreddits. The action to reply to a post or comment now lives in `actions/reply.js` because even though replies are comments, the action it self acts on comments OR posts. 

Tests are added for the new reducers and updated to reflect the refactor. Note that logged in / logged out tests now all prefill the state of their test's reducer so that we really know log in / log out clears the state to default.

👓 @phil303 @nramadas


src/app/reducers/accountRequests.js
```diff
@@ -1,4 +1,5 @@
 import merge from '@r/platform/merge';
+
 import * as accountActions from 'app/actions/accounts';
 import { newAccountRequest } from 'app/models/AccountRequest';
```
phil303: i'm not sure these are even necessary. It just feels like indirection.

src/app/reducers/accounts.js
```diff
@@ -0,0 +1,22 @@
...
+export default function(state=DEFAULT, action={}) {
+  switch (action.type) {
+    case loginActions.LOGGED_IN:
+    case loginActions.LOGGED_OUT: {
+      return DEFAULT;
+    }
+
+    case accountActions.RECEIVED_ACCOUNT: {
+      const { accounts } = action.apiResponse;
+      console.log('accounts are:', accounts);
```
phil303: print statement

schwers: shucks, that should really show up in `npm run lint` :/

src/app/reducers/accountRequests.js
schwers: Just for my understanding, do you mean the object creation should just live in the reducer? Or something else?

phil303: yea just `return { id: name, loading: true }`

phil303: I don't feel strongly. Maybe just moving the function into the file would ease the feeling of misdirection.

src/app/reducers/accounts.test.js
```diff
@@ -0,0 +1,57 @@
+import createTest from '@r/platform/createTest';
+
+import accounts from './accounts';
+import * as accountActions from 'app/actions/accounts';
+import * as loginActions from 'app/actions/login';
+
+createTest({ reducers: { accounts } }, ({ getStore, expect }) => {
+  describe('accoutns', () => {
```
phil303: nit misspelled accounts

src/app/reducers/accounts.test.js
```diff
@@ -0,0 +1,57 @@
...
+    describe('LOGGED_IN and LOGGED_OUT', () => {
+      it('should return default on log ou', () => {
```
phil303: misspelled out

src/app/reducers/subreddits.test.js
```diff
@@ -0,0 +1,127 @@
...
+        }));
+
+        const { subreddits } = store.getState();
+        expect(subreddits).to.eql({
+          [SUBREDDIT.uuid]: SUBREDDIT,
+        });
+      });
+    });
+
+    describe('RECEIEVED_SUBSCRIBED_SUBREDDITS', () => {
```
phil303: misspelled `RECEIVED`

phil303:
Overall I'm really happy with how this came out! Thanks for doing it. I think it'll make the code much easier to follow. Some nitty stuff but :fish:!

schwers:
Okay, thanks for the review. I'll fix the typos now and address the models in another pr today

PR #625, opened 2016-06-21
Relevancy experiment initial code

prashtx:
Add support for a variant of the next-content experiment. We will likely
tweak the variant and add others, but it's useful to have a basic
experiment present, so we can properly hook up the feature flag and
event code.

### DIFF
--- a/src/app/actions/commentsPage.js
+++ b/src/app/actions/commentsPage.js
@@ -1,8 +1,20 @@
+import { some } from 'lodash/collection';
+
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+import getSubreddit from 'lib/getSubredditFromState';
+import { cleanObject } from 'lib/cleanObject';
+import features from 'app/featureFlags';
 import { endpoints } from '@r/api-client';
 import { paramsToCommentsPageId } from 'app/models/CommentsPage';
+import { paramsToPostsListsId } from 'app/models/PostsList';
+import { fetching as fetchingPosts, received as receivedPosts } from 'app/actions/postsList';
+import { flags } from 'app/constants';
 
-const { CommentsEndpoint } = endpoints;
+const { CommentsEndpoint, PostsEndpoint } = endpoints;
+
+const {
+  VARIANT_NEXTCONTENT_BOTTOM,
+} = flags;
 
 export const FETCHING_COMMENTS_PAGE = 'FETCHING_COMMENTS_PAGE';
 
@@ -18,6 +30,13 @@ export const received = (commentsPageId, apiResponse) => ({
   type: RECEIVED_COMMENTS_PAGE,
   commentsPageId,
   apiResponse,
+});
+
+export const VISITED_COMMENTS_PAGE = 'VISITED_COMMENTS_PAGE';
+
+export const visitedCommentsPage = (postId) => ({
+  type: VISITED_COMMENTS_PAGE,
+  postId,
 });
 
 export const fetchCommentsPage = commentsPageParams => async (dispatch, getState) => {
@@ -36,3 +55,31 @@ export const fetchCommentsPage = commentsPageParams => async (dispatch, getState
   const apiResponse = await CommentsEndpoint.get(apiOptions, commentsPageParams);
   dispatch(received(commentsPageId, apiResponse));
 };
+
+export function relevantContentPostsParams({ subredditName }) {
+  return cleanObject({
+    sort: 'hot',
+    subredditName,
+  });
+}
+
+export const fetchRelevantContent =
+  () => async (dispatch, getState, { waitForState }) => {
+    await waitForState(getSubreddit, async () => {
+      const state = getState();
+      const feature = features.withContext({ state });
+      const subredditName = getSubreddit(state);
+      if (some([VARIANT_NEXTCONTENT_BOTTOM], variant => feature.enabled(variant))) {
+        const postsParams = relevantContentPostsParams({ subredditName });
+        const postsListId = paramsToPostsListsId(postsParams);
+        const postsList = state.postsLists[postsListId];
+
+        if (postsList) { return; }
+
+        dispatch(fetchingPosts(postsListId, postsParams));
+        const apiOptions = apiOptionsFromState(state);
+        const apiResponse = await PostsEndpoint.get(apiOptions, postsParams);
+        dispatch(receivedPosts(postsListId, apiResponse));
+      }
+    });
+  };
--- a/src/app/components/RelevantContent/index.jsx
+++ b/src/app/components/RelevantContent/index.jsx
@@ -1,0 +1,221 @@
+import './styles.less';
+
+import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+import { navigateToUrl } from '@r/platform/actions';
+import { METHODS } from '@r/platform/router';
+
+import filter from 'lodash/fp/filter';
+import first from 'lodash/fp/first';
+import flow from 'lodash/fp/flow';
+
+import getSubreddit from 'lib/getSubredditFromState';
+import mobilify from 'lib/mobilify';
+
+import { featuresSelector } from 'app/selectors/features';
+import { flags } from 'app/constants';
+
+import PostContent from 'app/components/Post/PostContent';
+
+import { paramsToPostsListsId } from 'app/models/PostsList';
+import { relevantContentPostsParams } from 'app/actions/commentsPage';
+
+import {
+  isPostDomainExternal,
+  cleanPostHREF,
+} from 'app/components/Post/postUtils';
+
+const {
+  VARIANT_NEXTCONTENT_BOTTOM,
+} = flags;
+
+
+const T = React.PropTypes;
+
+function NextPost(props) {
+  const {
+    topPosts,
+    width,
+    postId,
+    visited,
+    goToPost,
+    loid: { loid, loidCreated },
+    isSelf: isSelfText,
+  } = props;
+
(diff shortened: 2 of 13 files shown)


prashtx:
👓 @schwers @nramadas

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,236 @@
...
+            toggleEditing={ false }
+            saveUpdatedText={ false }
+            forceHTTPS={ true }
+            isDomainExternal={ externalDomain }
+            renderMediaFullbleed={ true }
+            showLinksInNewTab={ false }
+          />
+          <header className='NextContent__header'>
+            <div className='NextContent__post-descriptor-line'>
+            <a
```
schwers: nit: this should be indented

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,236 @@
...
+      isSelfText,
+      loid,
+      loidCreated,
+    });
+  });
+  const noop = (e => e.preventDefault());
+
+  const variant = 'bottom';
+  const descriptor = (
+    <a
```
schwers: If this isn't linking anywhere, could it be a span instead?

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,236 @@
...
+      topPosts,
+      feature,
+      visited,
+      loid,
+    };
+  },
+);
+
+const mapDispatchToProps = (dispatch) => ({
+  goToPost: ({
```
schwers: I wonder if this should rely on an `Anchor` tag instead? The even tracker middleware could handle the tracking action.

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,236 @@
...
+            { descriptor }
+          </header>
+        </div>
+        <div className='NextContent__next-link'>
+          <a
+            href='#'
+            onClick={ noop }
+          >
+            { actionText }
+            <span className='icon-nav-arrowforward icon'></span>
```
schwers: This should be a self-closing tag, we have it in the [style guide react tag section](https://github.com/reddit/styleguide/tree/master/javascript/react#tags) but haven't updated the linter fully yet

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,236 @@
...
+      />
+    );
+  }
+}
+
+const makeSelector = createSelector(
+  (_, props) => props.listingId,
+  (state, props) => state.posts[props.listingId],
+  (state) => state.posts,
+  featuresSelector,
```
schwers: 🔕 It might be worth adding some memoization to `featuresSelector` so it doesn't trigger unnecessary re-renders

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,236 @@
...
+              onClick={ noop }
+              target={ linkExternally ? '_blank' : null }
+            >
+              { title }
+            </a></div>
+            { descriptor }
+          </header>
+        </div>
+        <div className='NextContent__next-link'>
+          <a
```
schwers: Same question regarding use of `a` vs `span`

src/app/components/RelevantContent/styles.less
```diff
@@ -0,0 +1,147 @@
...
+    font-size: @font-size-base;
+    line-height: 1.5;
+    text-align: center;
+    text-transform: uppercase;
+    color: @blue;
+    margin-top: (@nextcontent-height - (1.5 * @font-size-base)) / 2;
+    margin-bottom: (@nextcontent-height - (1.5 * @font-size-base)) / 2;
+    margin-right: @grid-size;
+    margin-left: @grid-size;
+    a, a:hover, a:visited {
```
schwers: Could we use a class here instead of applying to all `a` tags?

src/app/actions/commentsPage.js
```diff
@@ -36,5 +55,35 @@ export const fetchCommentsPage = commentsPageParams => async (dispatch, getState
...
+    await waitForState(getSubreddit, async () => {
+      const state = getState();
+      const feature = features.withContext({ state });
+      const subredditName = getSubreddit(state);
+      if (some([VARIANT_NEXTCONTENT_BOTTOM], variant => feature.enabled(variant))) {
+        const postsParams = relevantContentPostsParams({ subredditName });
+        const postsListId = paramsToPostsListsId(postsParams);
+        const postsList = state.postsLists[postsListId];
+
+        if (!postsList) {
```
schwers: small nit, for consistency with other async actions, this should return early if you have `postsList`

src/app/components/RelevantContent/styles.less
```diff
@@ -0,0 +1,147 @@
...
+      vertical-align: bottom;
```
schwers: Just my opinion on style, but you might be able to do something like
```less
&, .icon {
  text-decoration: none;
  &.themeify({
    color: @blue;
  }
}

.icon {
   font-size: @font-size-base;
   line-height: 1.5;
}
```
to remove some of the the duplication and show what's common.

schwers:
Just a reminder to add tests for the new actions / reducer. 💅

src/app/components/RelevantContent/index.jsx
prashtx: The experiment events will require separate actions, which `Anchor` doesn't currently dispatch. We can't issue those events for every navigation. The details of those will also depend on aspects of this component, particularly for future variants (for example, did the user click the 1st or 2nd link in a list).

It might be worth giving the `Anchor` component an optional custom dispatcher property, but I think that's out of scope for this change.

schwers: Just to be sure I understand, when you say we can't fire them on every navigation, are you referring the case of: click a link in the relevancy content -> reload that page? That would certainly prevent the tracking from all relying on the navigation action, but instead you could pass an `onClick` function to the `Anchor` component. It will get called before dispatching a navigation action ([source](https://github.com/reddit/node-platform/blob/master/lib/modules/components.js#L90)). This assumes the tracking could be dispatched before the navigation action.

This could also be a style preference on my part, but I guess I would have expected something like
```javascript
<Anchor href={ postHref } onClick={ () => fireExperimentTracking(postId) } />
```

prashtx: Summary of offline discussion: @schwers's suggestion totally makes sense, but we're reusing `PostContent`, which has anchor tags, which can't be nested under other anchor tags. So for now I'll stick with this implementation, but as we flesh out future variants, we should consider factoring out the useful parts of `PostContent`

Philosophically, some folks think anchors should be closer to "points" in the page rather than "areas", and we'd be using `Anchor` for the latter. But I think we'll want to investigate any SEO/crawler implications of wrapping the area in an `a` vs a clickable `div`.

prashtx:
💇

src/app/actions/commentsPage.js
```diff
@@ -36,5 +55,35 @@ export const fetchCommentsPage = commentsPageParams => async (dispatch, getState
...
+        const postsParams = relevantContentPostsParams({ subredditName });
+        const postsListId = paramsToPostsListsId(postsParams);
+        const postsList = state.postsLists[postsListId];
+
+        if (postsList) { return; }
+
+        dispatch(fetchingSubredditPosts(postsListId, postsParams));
+        const apiOptions = apiOptionsFromState(state);
+        const apiResponse = await PostsEndpoint.get(apiOptions, postsParams);
+        dispatch(receivedResponse(apiResponse));
```
schwers: Just a heads up, this will have to be rebased for the changes in https://github.com/reddit/reddit-mobile/pull/624 (already landed)

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,221 @@
...
+    topPosts,
+    width,
+    listingId,
+    visited,
+    goToPost,
+    loid: { loid, loidCreated },
+    isSelf: isSelfText,
+  } = props;
+
+  const safeAndNew = (link =>
```
schwers: Minor nit, could you change `link` -> `post` and `listingId` and -> `postId`? We've moved away from those names as much as possible for clarity.

schwers:
One minor nit and rebase 💅

prashtx:
Changed variable names and rebased

💇

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,221 @@
...
+        <div className='NextContent__post-wrapper'>
+          <PostContent
+            post={ postWithFallback }
+            single={ false }
+            compact={ true }
+            expandedCompact={ false }
+            onTapExpand={ function () {} }
+            width={ width }
+            toggleShowNSFW={ false }
+            showNSFW={ false }
```
schwers: 🔕 We pull preferences now so we could use `preferences.over18` if desired.

src/app/components/RelevantContent/index.jsx
```diff
@@ -0,0 +1,221 @@
...
+        isSelf={ post.isSelf }
+      />
+    );
+  }
+}
+
+const makeSelector = createSelector(
+  (_, props) => props.postId,
+  (state, props) => state.posts[props.postId],
+  (state) => state.posts,
```
schwers: Really really small nit I should have caught earlier: could you remove the parens here when they're unneeded? Style wise we've been converging on not using the parens in arrow syntax when they're unneeded (single arg functions)

schwers:
:fish: (just one small nit if you can get to, removing some parens)

PR #627, opened 2016-06-22
Enable post expanding

schwers:
This adds post expanding behavior from mweb 1x. 

There was a change to `aspect-ratio.less` [commit](https://github.com/reddit/art-gallery/commit/d4dbd5bea64e8dc7bde2c76301d3259ba78126c6) that was trying to remove padding to fix hit areas on comments pages. That fix should have been removing the height set the images rendered by `PostContent` because the difference in that height versus the `aspect-ratio` seems to be what was really causing issues. This needed to be reverted in this patch to fix expands, and actually card view in general.

This depends on https://github.com/reddit/node-api-client/pull/127 to ensure the failed https requests to  `http:self` disappear and fixes empty preview images

### DIFF
--- a/src/app/components/Post/PostContent/index.jsx
+++ b/src/app/components/Post/PostContent/index.jsx
@@ -155,22 +155,6 @@ export default class PostContent extends React.Component {
     return DEFAULT_ASPECT_RATIO;
   }
 
-  heightStyle(aspectRatio) {
-    if (this.props.single && aspectRatio) {
-      return `${1 / aspectRatio * this.props.width}px`;
-    }
-  }
-
-  baseStyleFromAspectRatio(aspectRatio) {
-    const style = {};
-    const heightStyle = this.heightStyle(aspectRatio);
-    if (heightStyle) {
-      style.height = heightStyle;
-    }
-
-    return style;
-  }
-
   render() {
     const isCompact = this.isCompact();
     const { single, post, isDomainExternal } = this.props;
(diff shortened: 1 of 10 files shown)


schwers:
👓  @nramadas or @uzi

src/app/reducers/expandedPosts.test.js
```diff
@@ -0,0 +1,55 @@
+import createTest from '@r/platform/createTest';
+
+import expandedPosts from './expandedPosts';
+import * as loginActions from 'app/actions/login';
+import * as postActions from 'app/actions/posts';
+
+createTest({ reducers: { expandedPosts }}, ({ getStore, expect }) => {
+  describe('exapndedPosts', () => {
```
uzi: `exapndedPosts`

src/app/components/Post/PostContent/index.jsx
```diff
@@ -307,11 +291,16 @@ export default class PostContent extends React.Component {
     }
 
     const { onTapExpand } = this.props;
+    const callOnTapExpand = (e) => {
+      e.preventDefault();
+      onTapExpand();
+    };
+
```
uzi: Are we trying to get rid of putting the parenthesis around the (e) in this case since there's only the one arg?

uzi:
💇 Couple of nits

src/app/components/Post/PostContent/index.jsx
schwers: Yes, thanks for the reminder! I'll make a note of trying to find a linter rule for this

schwers:
💇

uzi:
:fish:

PR #628, opened 2016-06-22
WIP - Add messages

nramadas:
This is just a work in progress. I'm hoping this can go into the product while I wait on UX/design updates and move on to something else in the meantime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ bin/*
 build/
 .test/
 .DS_Store
-
+reddit-mobile.log
--- a/src/app/components/DirectMessage/index.jsx
+++ b/src/app/components/DirectMessage/index.jsx
@@ -1,0 +1,46 @@
+import './styles.less';
+import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+
+import DirectMessageHeader from './Header';
+import DirectMessageComposition from './Composition';
+
+const T = React.PropTypes;
+
+export function DirectMessage(props) {
+  return (
+    <div className='DirectMessage'>
+      <div className='DirectMessage__header'>
+        <DirectMessageHeader/>
+      </div>
+      <div className='DirectMessage__body'>
+        <DirectMessageComposition
+          username={ props.user.name }
+          recipient={ props.recipient }
+        />
+      </div>
+    </div>
+  );
+}
+
+DirectMessage.propTypes = {
+  user: T.object.isRequired,
+  recipient: T.string,
+};
+
+DirectMessage.defaultProps = {
+  recipient: '',
+};
+
+const selector = createSelector(
+  state => state.user,
+  user => ({ user })
+);
+
+const mergeProps = (stateProps, dispatchProps, ownProps) => ({
+  ...stateProps,
+  recipient: ownProps.queryParams.to,
+});
+
+export default connect(selector, null, mergeProps)(DirectMessage);
--- a/src/app/components/Messages/Comment.jsx
+++ b/src/app/components/Messages/Comment.jsx
@@ -1,0 +1,28 @@
+import './Comment.less';
+import React from 'react';
+
+import { short } from 'lib/formatDifference';
+
+const T = React.PropTypes;
+
+export default function MessagesComment(props) {
+  const { comment } = props;
+
+  return (
+    <div className='MessagesComment'>
+      <div className='MessagesComment__title'>
+        { comment.linkTitle }
+      </div>
+      <div className='MessagesComment__metaData'>
+        { `${ comment.author } \u2022 r/${ comment.subreddit } \u2022 ${ short(comment.createdUTC) }` }
+      </div>
+      <div className='MessagesComment__body'
+        dangerouslySetInnerHTML={ { __html: comment.bodyHTML } }
+      />
+    </div>
+  );
+}
+
+MessagesComment.propTypes = {
+  comment: T.object.isRequired,
+};
--- a/src/app/components/Messages/Comment.less
+++ b/src/app/components/Messages/Comment.less
@@ -1,0 +1,36 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/themes/themeify';
+
+.MessagesComment {
+  &__title {
+    font-size: 13px;
+    line-height: 1.5;
+    margin-bottom: 3px;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+
+    .themeify({
+      color: @theme-body-text-color;
+    });
+  }
+
+  &__metaData {
+    font-size: 11px;
+    line-height: 1.2;
+    margin-bottom: @grid-size;
+
+    .themeify({
+      color: @theme-meta-text-color;
+    });
+  }
+
+  &__body {
+    font-size: 11px;
+    line-height: 1.2;
+
+    .themeify({
+      color: @theme-body-text-color;
+    });
+  }
+}
--- a/src/app/components/Messages/Header.jsx
+++ b/src/app/components/Messages/Header.jsx
@@ -1,0 +1,16 @@
+import './Header.less';
+import React from 'react';
+// import { Anchor } from '@r/platform/components';
+
+export default function MessagesHeader() {
+  return (
+    <div className='MessageHeader'>
+      <div className='MessageHeader__title'>Inbox</div>
+      {/*Don't allow access to messager composition until that feature is ready
+      <Anchor
+        className='MessageHeader__compose icon icon-message'
+        href='/message/compose'
+      />*/}
+    </div>
+  );
+}
--- a/src/app/components/Messages/Message.jsx
+++ b/src/app/components/Messages/Message.jsx
@@ -1,0 +1,35 @@
+import './Message.less';
+import React from 'react';
+import { Anchor } from '@r/platform/components';
+
+import { short } from 'lib/formatDifference';
+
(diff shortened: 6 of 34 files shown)


.gitignore
```diff
@@ -7,4 +7,4 @@ bin/*
 build/
 .test/
 .DS_Store
-
+reddit-mobile.log
```
phil303: just curious, what is this?

src/app/components/DirectMessage/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  recipient: T.string,
+};
+
+DirectMessage.defaultProps = {
+  recipient: '',
+};
+
+const selector = createSelector(
+  state => state.user,
+  (user) => ({ user })
```
uzi: Are the parenthesis needed on either of these?

src/app/components/Messages/Comment.jsx
```diff
@@ -0,0 +1,28 @@
...
+export default function MessagesComment(props) {
+  const { comment } = props;
+
+  return (
+    <div className='MessagesComment'>
+      <div className='MessagesComment__title'>
+        { comment.linkTitle }
+      </div>
+      <div className='MessagesComment__metaData'>
+        { `${ comment.author } • r/${ comment.subreddit } • ${ short(comment.createdUTC) }` }
```
uzi: Is there a reason why we'd want to do something like the hex or html version of those dots instead of the unicode character?

src/app/components/Messages/Header.jsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <div className='MessageHeader'>
+      <div className='MessageHeader__title'>Inbox</div>
+      {/*<Anchor
+        className='MessageHeader__compose icon icon-message'
+        href='/message/compose'
+      />*/}
+    </div>
+  );
+}
```
uzi: Commented stuff in this file.  Remove or maybe we want to add comments about it?

src/app/components/Messages/Message.jsx
```diff
@@ -0,0 +1,35 @@
...
+  const { message } = props;
+
+  return (
+    <div className='MessagesMessage'>
+      <div className='MessagesMessage__header'>
+        <div className='MessagesMessage__title'>
+          { message.subject }
+        </div>
+        <div className='MessagesMessage__subtitle'>
+          { `${ message.author } • ${ short(message.createdUTC) }` }
```
uzi: Same unicode character question

src/app/components/Messages/Message.less
```diff
@@ -0,0 +1,65 @@
...
+
+  &__body {
+    font-size: 11px;
+    line-height: 1.2;
+
+    .themeify({
+      color: @theme-body-text-color;
+    });
+
+    a {
```
schwers: Can the styles in [text.less](https://github.com/reddit/reddit-mobile/blob/2X/src/app/less/components/text.less#L9) handle this?

src/app/components/Messages/index.jsx
```diff
@@ -0,0 +1,79 @@
...
+  mailType: T.string.isRequired,
+};
+
+const renderItem = (type, id, messages, comments, posts) => {
+  switch (type) {
+    case 'comment': return <MessagesComment comment={ comments[id] }/>;
+    case 'message': return <MessagesMessage message={ messages[id] }/>;
+    case 'post': return <MessagesPost post={ posts[id] }/>;
+    default: return null;
+  }
```
uzi: I was about to brain fart and ask if you needed breaks ... you're returning in all cases, so you don't.  Ummm... I usually like the formatting of

```
    case 'foo':
        return something();
    case 'bar':
        return somethingElse();
    ...
```

But that's a personal stylistic choice.  I just like whitespace.

src/app/components/DirectMessage/index.jsx
nramadas: they are needed on the second. i can remove them on the first.

src/app/components/DirectMessage/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+export function DirectMessage(props) {
+  return (
+    <div className='DirectMessage'>
+      <div className='DirectMessage__header'>
+        <DirectMessageHeader/>
+      </div>
+      <div className='DirectMessage__body'>
+        <DirectMessageComposition
+          username='nr4madas'
```
schwers: Should we add a todo for this component and un-hardcode your username?

src/app/components/Messages/Message.less
nramadas: i didn't know that was a thing! i'll see if i can use text.less

.gitignore
nramadas: not sure. it started appearing in my commits (and they seemed totally useless) so i'm gitignoring them for now. I'll spend some time later trying to figure out what it does.

src/app/components/DirectMessage/index.jsx
nramadas: yes. we definitely should.

phil303: replace it with your password plz

.gitignore
ajacksified: https://github.com/reddit/reddit-mobile/blob/8767a9747f4a5e03091b0f9729db1a298131a214/src/server/console.es6.js#L26

logs requests when you use the ascii console.

src/app/components/Messages/Comment.less
```diff
@@ -0,0 +1,42 @@
...
+
+  &__body {
+    font-size: 11px;
+    line-height: 1.2;
+
+    .themeify({
+      color: @theme-body-text-color;
+    });
+
+    a {
```
schwers: Same text.less question

src/app/components/Messages/Header.jsx
nramadas: added a comment

src/app/reducers/mail.js
```diff
@@ -0,0 +1,51 @@
...
+    case mailActions.PENDING: {
+      const { mailType } = action;
+      return merge(state, {
+        [mailType]: {
+          pending: true,
+          error: null,
+        },
+      });
+    }
+    case mailActions.SUCCESS: {
```
schwers: I think we should stick with `FETCHING` and `RECEIVED` convention the other files are using just for consistency. Happy to switch up the names but seems like that'd be good in a separate patch.

schwers:
💅

.gitignore
phil303: cool thanks for the info

src/app/router/handlers/Messages.js
```diff
@@ -0,0 +1,23 @@
...
+import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+import * as mailActions from 'app/actions/mail';
+
+export default class Messages extends BaseHandler {
+  async [METHODS.GET](dispatch, getState/*, utils*/) {
+    const { mailType } = this.urlParams;
+    const apiOptions = apiOptionsFromState(getState());
+
+    fetchUserBasedData(dispatch);
+    dispatch(mailActions.setInboxPending(mailType));
```
phil303: I have to look at the platform source again but would it make sense to move this into an action creator in `actions/mail.js`? The goal here being to maintain more consistency with other action files in our code base (at least the mweb one).

src/app/components/SettingsOverlayMenu/index.jsx
```diff
@@ -67,7 +67,7 @@ export const renderLoggedInUserRows = (user) => {
     <LinkRow
       key='inbox'
       text={ ['Inbox', badge] }
-      href='/message/inbox'
+      href='/message/messages'
       icon={ `icon-message icon-large ${inboxCount ? 'orangered' : 'blue'}` }
```
nramadas: sure

nramadas:
:haircut:

src/app/components/Messages/Post.jsx
```diff
@@ -0,0 +1,9 @@
+import './Post.less';
+import React from 'react';
+
+export default function MessagesPost() {
+  return (
+    <div className='MessagesPost'>
```
schwers: Could you add a note or todo here explaining what needs to be done in the Post case?

nramadas: sure

src/app/reducers/posts.js
```diff
@@ -26,7 +27,8 @@ export default function(state=DEFAULT, action={}) {
     case postsListActions.RECEIVED_POSTS_LIST:
     case hiddenActions.RECEIVED_HIDDEN:
     case savedActions.RECEIVED_SAVED:
-    case searchActions.RECEIVED_SEARCH_REQUEST: {
+    case searchActions.RECEIVED_SEARCH_REQUEST:
+    case mailActions.SUCCESS: {
```
schwers: It looks like not all of the `success` constants were converted to `received`. I get an empty inbox at the moment but it looks like its just because the reducers aren't pulling in the data

schwers:
Just two small things, looks good! 💅

lib/snooboots
```diff
@@ -0,0 +1 @@
+Subproject commit 50555e5c9894a91bc581338b584f9c8cfd982bc7
```
schwers: Whoops, I should have caught this earlier, but it looks like you're going to have to remove `lib/snooboots` and `lib/art-gallery`.

nramadas:
:haircut:

schwers:
:fish:

PR #629, opened 2016-06-27
Adds NSFW Interstitial and post blurring support

schwers:
Depends on https://github.com/reddit/node-api-client/pull/132 (git path hardcoded in package.json for testing purposes only)

NSFW subreddits will now have a NSFW interstitial if we haven't confirmed the user is over 18. If the user is logged in that confirmation will come from the preferences object. We weren't fetching it before but have to use the endpoint anyway to update that preference, so now we're fetching it as well. This sets up for using default comment sort and etc.

Posts in non-nsfw subreddits will now have the proper NSFW label and tap/click'ing them will toggle the blur.

👓 @nramadas @phil303

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test:watch": "npm run test -- -w"
   },
   "dependencies": {
-    "@r/api-client": "~3.14.6",
+    "@r/api-client": "~3.16.0",
     "@r/build": "^0.6.0",
     "@r/flags": "^1.5.0",
     "@r/middleware": "~0.6.1",
--- a/src/Server.js
+++ b/src/Server.js
@@ -18,6 +18,7 @@ import refreshproxy from 'server/session/refreshproxy';
 import dispatchSession from 'server/session/dispatchSession';
 import { dispatchInitialCompact } from 'server/initialState/dispatchInitialCompact';
 import { dispatchInitialLoid } from 'server/initialState/dispatchInitialLoid';
+import { dispatchInitialOver18 } from 'server/initialState/dispatchInitialOver18';
 import { dispatchInitialTheme } from 'server/initialState/dispatchInitialTheme';
 import metaRoutes from 'server/meta';
 
@@ -55,6 +56,7 @@ export function startServer() {
       dispatchInitialTheme(ctx, dispatch);
       dispatchInitialCollapsedComments(ctx, dispatch);
       dispatchInitialCompact(ctx, dispatch);
+      dispatchInitialOver18(ctx, dispatch);
     },
     preRouteServerMiddleware: [
       binFiles,
--- a/src/app/actions/preferences.js
+++ b/src/app/actions/preferences.js
@@ -1,0 +1,65 @@
+import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+import { endpoints } from '@r/api-client';
+
+const { PreferencesEndpoint } = endpoints;
+
+export const FETCHING = 'FETCHING_PREFERENCES';
+export const fetching = () => ({
+  type: FETCHING,
+});
+
+export const RECEIEVED = 'RECEIEVED_PREFERENCES';
+export const received = preferences => ({
+  type: RECEIEVED,
+  preferences,
+});
+
+export const FETCH_FAILED = 'FAILED_PREFERENCES_FETCH';
+export const fetchFailed = error => ({
+  type: FETCH_FAILED,
+  error,
+});
+
+export const fetch = () => async (dispatch, getState) => {
+  const state = getState();
+  // don't fetch if we're already doing so or if we're not logged in
+  const { preferencesRequest } = state;
+  if (preferencesRequest.pending || preferencesRequest.succeeded) { return; }
+  if (!state.session.accessToken) { return; }
+
+  try {
+    const preferences = await PreferencesEndpoint.get(apiOptionsFromState(state));
+    dispatch(received(preferences));
+  } catch (e) {
+    dispatch(fetchFailed(e));
+  }
+};
+
+export const patch = changes => async (dispatch, getState) => {
+  const state = getState();
+  const updatedPreferences = await PreferencesEndpoint.patch(apiOptionsFromState(state), changes);
+  dispatch(received(updatedPreferences));
+};
+
+export const IS_OVER_18 = 'IS_OVER_18';
+export const isOver18 = () => ({
+  type: IS_OVER_18,
+});
+
+export const setOver18 = () => async (dispatch, getState) => {
+  const state = getState();
+
+  // if the user isn't logged in, dispatch the IS_OVER_18 action and let reducers
+  // update the state appropriately
+  if (!state.session.accessToken) {
+    dispatch(isOver18());
+    return;
+  }
+
+  try {
+    await dispatch(patch({ over_18: true }));
+  } catch (e) {
+    // fallback to setting over 18 manually if the api call fails for somereason
+    dispatch(isOver18());
+  }
+};
--- a/src/app/actions/preferences.js
+++ b/src/app/actions/preferences.js
@@ -3,20 +3,20 @@ import { endpoints } from '@r/api-client';
 
 const { PreferencesEndpoint } = endpoints;
 
-export const FETCHING = 'FETCHING_PREFERENCES';
-export const fetching = () => ({
-  type: FETCHING,
+export const PENDING = 'PENDING_PREFERENCES';
+export const pending = () => ({
+  type: PENDING,
 });
 
-export const RECEIEVED = 'RECEIEVED_PREFERENCES';
+export const RECEIVED = 'RECEIVED_PREFERENCES';
 export const received = preferences => ({
-  type: RECEIEVED,
+  type: RECEIVED,
   preferences,
 });
 
-export const FETCH_FAILED = 'FAILED_PREFERENCES_FETCH';
+export const FAILED = 'FAILED_PREFERENCES';
 export const failed = error => ({
-  type: FETCH_FAILED,
+  type: FAILED,
   error,
 });
 
@@ -26,6 +26,8 @@ export const fetch = () => async (dispatch, getState) => {
   const { preferencesRequest } = state;
   if (preferencesRequest.pending || preferencesRequest.succeeded) { return; }
   if (!state.session.accessToken) { return; }
+
+  dispatch(pending());
 
   try {
     const preferences = await PreferencesEndpoint.get(apiOptionsFromState(state));
@@ -37,8 +39,21 @@ export const fetch = () => async (dispatch, getState) => {
 
 export const patch = changes => async (dispatch, getState) => {
   const state = getState();
-  const updatedPreferences = await PreferencesEndpoint.patch(apiOptionsFromState(state), changes);
-  dispatch(received(updatedPreferences));
+
+  // In response to a `PATCH`, the PreferencesEndpoint returns the full
+  // preferences model back to us, so we should use it to update state.
+  // This means a `fetch` action doesn't have to fire because we're effectively
+  // fetching already. However we don't block or return early because patch
+  // writes, potentially new info, each time.
+  dispatch(pending());
+
+  try {
+    const updatedPreferences = await PreferencesEndpoint.patch(
+      apiOptionsFromState(state), changes);
+    dispatch(received(updatedPreferences));
+  } catch (e) {
+    dispatch(failed(e));
+  }
 };
 
 export const IS_OVER_18 = 'IS_OVER_18';
@@ -49,17 +64,14 @@ export const isOver18 = () => ({
 export const setOver18 = () => async (dispatch, getState) => {
   const state = getState();
 
-  // if the user isn't logged in, dispatch the IS_OVER_18 action and let reducers
-  // update the state appropriately
+  // optimistically update over18. If the user is not logged in
+  // we don't have to do anything else.
+  dispatch(isOver18());
+
   if (!state.session.accessToken) {
-    dispatch(isOver18());
     return;
   }
 
-  try {
-    await dispatch(patch({ over_18: true }));
-  } catch (e) {
-    // fallback to setting over 18 manually if the api call fails for somereason
-    dispatch(isOver18());
-  }
+  // Otherwise we'll call patch so the real preferences object is updated
+  dispatch(patch({ over_18: true }));
 };
--- a/src/app/actions/preferences.js
+++ b/src/app/actions/preferences.js
@@ -15,7 +15,7 @@ export const received = preferences => ({
 });
 
 export const FETCH_FAILED = 'FAILED_PREFERENCES_FETCH';
-export const fetchFailed = error => ({
+export const failed = error => ({
   type: FETCH_FAILED,
   error,
 });
@@ -31,7 +31,7 @@ export const fetch = () => async (dispatch, getState) => {
     const preferences = await PreferencesEndpoint.get(apiOptionsFromState(state));
     dispatch(received(preferences));
   } catch (e) {
-    dispatch(fetchFailed(e));
+    dispatch(failed(e));
   }
 };
 
--- a/src/app/components/NSFWInterstitial/index.jsx
+++ b/src/app/components/NSFWInterstitial/index.jsx
@@ -24,7 +24,7 @@ export default function NSFWInterstitial() {
           action='/actions/setOver18'
           className='NSFWInterstitial__form'
         >
-          <button typoe='submit' className='NSFWInterstitial__button'>
+          <button type='submit' className='NSFWInterstitial__button'>
             CONTINUE
           </button>
         </Form>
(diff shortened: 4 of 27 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "test:watch": "npm run test -- -w"
   },
   "dependencies": {
-    "@r/api-client": "~3.14.3",
+    "@r/api-client": "git://github.com/schwers/snoode#48e8e107f58f1a4e7a90918b349043d4044c9120",
```
phil303: did you mean to check this in?

src/app/actions/preferences.js
```diff
@@ -0,0 +1,65 @@
...
+import { endpoints } from '@r/api-client';
+
+const { PreferencesEndpoint } = endpoints;
+
+export const FETCHING = 'FETCHING_PREFERENCES';
+export const fetching = () => ({
+  type: FETCHING,
+});
+
+export const RECEIEVED = 'RECEIEVED_PREFERENCES';
```
phil303: `RECEIVED`

src/app/reducers/preferencesRequest.js
```diff
@@ -0,0 +1,44 @@
...
+  switch (action.type) {
+    case loginActions.LOGGED_IN:
+    case loginActions.LOGGED_OUT: {
+      return DEFAULT;
+    }
+
+    case preferenceActions.FETCHING: {
+      return merge(state, {
+        pending: true,
+        failed: false,
```
phil303: Is this for handling retries? I think if that's the case, `FETCHING` might be better acting as the "reset" for state (minus the `pending` key). In other words, it'd also set `success` to `false`. Then your `RECEIVED` would just set `pending` and `succeeded` while your `FAILED` would set `pending` and `failed`.

src/app/reducers/preferencesRequest.js
```diff
@@ -0,0 +1,44 @@
...
+import * as loginActions from 'app/actions/login';
+import * as preferenceActions from 'app/actions/preferences';
+
+export const DEFAULT = {
+  succeeded: false,
+  pending: false,
+  failed: false,
+};
+
+export default function(state=DEFAULT, action={}) {
```
phil303: As a general sort of observation, this is a little different than how I've handled redux state in the past and I'm interested in both your and @nramadas's opinion here. I normally shy away from nested state like that of preferences in the `user` key. How I naively might've handled this would be to have a `preferences` reducer which also has knowledge of the fetch state.

Overall the state might look like:
```js
{
  user: {
    // stuff 
  },
  preferences: {
    pending: false,
    success: true,
    failed: false,
    over18: true,
    // ... other preferences
  },
}
```

All that being said, I actually kind of like this as well and as long as we were consistent about it - "if you have nested async state, this is how you make a reducer for that", I'd be happy either way.

src/app/router/handlers/handlerCommon.js
```diff
@@ -1,8 +1,9 @@
 import * as subscribedSubredditsActions from 'app/actions/subscribedSubreddits';
+import * as preferenceActions from 'app/actions/preferences';
 import * as userActions from 'app/actions/user';
 
-
 export const fetchUserBasedData = (dispatch) => {
   dispatch(subscribedSubredditsActions.fetchSubscribedSubreddits(true));
   dispatch(userActions.fetchMyUser());
+  dispatch(preferenceActions.fetch());
```
phil303: Some day....
687474703a2f2f657069637965617270726f6a6563742e66696c65732e776f726470726573732e636f6d2f323031312f30352f6261737469616e2d66616c636f722d77686f6f702d3439302e676966

src/app/components/NSFWInterstitial/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+      </p>
+      <div className='NSFWInterstitial__buttons'>
+        <Anchor href='/' className='NSFWInterstitial__button'>
+          NO THANK YOU
+        </Anchor>
+        <Form
+          action='/actions/setOver18'
+          className='NSFWInterstitial__form'
+        >
+          <button typoe='submit' className='NSFWInterstitial__button'>
```
phil303: spelled type wrong here

src/app/pages/PostsFromSubreddit.jsx
```diff
@@ -14,35 +16,68 @@ import isFakeSubreddit from 'lib/isFakeSubreddit';
...
-            rightLink= { subnavLink }
-            showWithoutUser={ true }
-          /> }
-        <PostsList postsListId={ postsListId } />
-      </div>
+    <div className={ className }>
+      { !forFakeSubreddit && <CommunityHeader subredditName={ subredditName } /> }
+      { renderSubnav &&
+        <SubNav
+          rightLink= { subnavLink }
```
phil303: nit extra space after equals sign

phil303:
:nail_care:

package.json
schwers: Yes, it depends on an apiclient patch, this is for testing purposes only. More details in the description (that patch needs reviewing as well if you have some time)

phil303: ack sorry i completely glossed over that line

schwers: No worries :)

src/app/reducers/preferencesRequest.js
schwers: The case I had in mind was if you land on a NSFW subreddit, and the preference request fails. You could hit the okay button and  the `patch` api call could work, which will call `RECEIVED`. In that example setting more of state explicitly seemed like the right thing to do. If you think its better to always go through a reset state I'd be fine with changing `patch` to call `fetching` as well, and maybe even rename `fetching` to `pending` to be more request type agnostic.

src/app/pages/PostsFromSubreddit.jsx
schwers: good catch, thanks

src/app/components/NSFWInterstitial/index.jsx
nramadas: in fact, it's actually pretty meta. 'type' was spelled 'typo', but 'typo' also has a type.

src/Server.js
```diff
@@ -52,6 +53,7 @@ export function startServer() {
       dispatchInitialTheme(ctx, dispatch);
       dispatchInitialCollapsedComments(ctx, dispatch);
       dispatchInitialCompact(ctx, dispatch);
+      dispatchInitialOver18(ctx, dispatch);
```
nramadas: i feel like we need to revisit `dispatchBeforeNavigation`. I suppose it works as intended right now, but considering the number of items that need to be dispatched before any thunk'd actions get called makes me feel like this isn't the best solution.

src/app/components/NSFWInterstitial/index.jsx
phil303: that's the most amazing thing i've ever read.

src/app/actions/preferences.js
nramadas: fun fact, the rule "i before e except after c" actually has more exceptions to it than situations where the rule applies.

src/app/actions/preferences.js
```diff
@@ -0,0 +1,65 @@
...
+export const failed = error => ({
+  type: FETCH_FAILED,
+  error,
+});
+
+export const fetch = () => async (dispatch, getState) => {
+  const state = getState();
+  // don't fetch if we're already doing so or if we're not logged in
+  const { preferencesRequest } = state;
+  if (preferencesRequest.pending || preferencesRequest.succeeded) { return; }
```
nramadas: how does this feel to you? i've wondered about the cleanest way to either block or take the latest response. Phil looked into `redux-saga`s for a bit, which actually seem kinda nice (they introduce observable like 'takeLatest' and 'takeEach' descriptors).

what you have doesn't seem terribly complex and might be enough for our needs. does it feel ok to you?

src/app/actions/preferences.js
schwers: "the more you know"

src/app/components/NSFWInterstitial/index.jsx
phil303: I believe submit is the default

phil303: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

src/app/components/NSFWInterstitial/styles.less
```diff
@@ -0,0 +1,48 @@
...
+  margin: 0 auto;
+
+  &__icon-wrapper {
+    margin: 0 auto;
+    height: 100px;
+    text-align: center;
+  }
+
+  &__icon {
+    font-size: 7em;
```
nramadas: have we decided on `em`s vs `px`s? we should figure out how exactly we want to size these things.

src/app/router/handlers/handlerCommon.js
schwers: 💯 A++ use of hover text

src/app/pages/PostsFromSubreddit.jsx
```diff
@@ -14,35 +16,68 @@ import isFakeSubreddit from 'lib/isFakeSubreddit';
...
-        { !forFakeSubreddit && <CommunityHeader subredditName={ subredditName } /> }
-        { renderSubnav &&
-          <SubNav
-            rightLink= { subnavLink }
-            showWithoutUser={ true }
-          /> }
-        <PostsList postsListId={ postsListId } />
-      </div>
+    <div className={ className }>
+      { !forFakeSubreddit && <CommunityHeader subredditName={ subredditName } /> }
```
nramadas: small syntax thing, are we okay with doing this vs ternaries?

src/Server.js
schwers: I sometimes wonder if it should be analogous to `Client::modifyData` -- it would be less idiomatic but would reduce the number of dispatches. Alternatively if might be interesting to pass ctx into reducers on init so they have a way of bootstrapping -- that also feels a little wrong though, but maybe it will spark some ideas...

src/app/reducers/preferencesRequest.js
phil303: Oh I see. Hmm, that's an interesting situation. I'd think in this particular case, we'd want to update the store optimistically anyway? In other words, set it to the users' preference in memory using `isOver18` and move forward in the ux while not waiting on the request. Since you're not waiting on the request's status, you don't really need to set anything in state.

src/app/pages/PostsFromSubreddit.jsx
schwers: Open to discussion, I've found myself slipping into it after using the ternaries for a while. Any preferences?

phil303: I have a slight preference for ternaries. Whenever I see this pattern (and I do it all the time), it feels slightly abusive.

src/app/components/NSFWInterstitial/styles.less
schwers: Honestly I was just porting over styles from the old interstitial and didn't give it much thought. I'll switch it to `px` for consistency for now. Philosophically I think we should consider using `em`s everywhere to be more responsive to a users preferred font-size (at least on iOS you can set fonts to be larger as an accessibility feature).

src/app/router/handlers/handlerCommon.js
phil303: I'm honestly surprised you found it. I had to look up how to do it :P

src/app/reducers/preferencesRequest.js
nramadas: yeah, so i too generally group the api status of data with the data itself, and put both in the reducer. my thinking being that if you maintain one reducer per "resource", it becomes a little easier to reason with all the actions that might change that resource. So, a reducer for 'user' listens to all the actions that affect the user, and manages all the api state that goes along with it.

i don't think it's crazy to have a reducer just for api state. i do worry a bit about "reducer fatigue" (a term i just made up). We have so many reducers right now, it might start to become a challenge understanding what reducers are in charge of what.

something else we could do instead is use combineReducer more aggressively. so imagine a folder structure like:
```
/reducers
 |-> index.js
 |-> /preferences
        |-> index.js
        |-> /api
              |-> index.js
        |-> /data
              |-> index.js
```
so, `index.js` in `/reducers/preferences` would use `combineReducer` to merge in `/reducers/preferences/api` and `reducers/peferences/data`. then, it becomes easier to understand where to look if we're messing with prefs, but we still maintain the separation of concerns for each "sub-reducer".

src/app/pages/PostsFromSubreddit.jsx
nramadas: yeah, i do it too (especially on personal projects), but it feels wrong. it could be we're being hard on ourselves and it's actually ok, or maybe our intuition is correct and we should steer away.

src/app/router/handlers/handlerCommon.js
nramadas: holy shit, i had no idea what the gif was supposed to be until schwers pointed out the hover text. now i'm loving that gif.

src/app/actions/preferences.js
schwers: I've enjoyed it. I like that it's fairly readable, and easy to extend or modify with your own request specific caching-logic if needed (e.g. invalidate if an hour passes or etc).

src/Server.js
nramadas: yeah. the thing i worry about with something like `modifyData` is that we'll need to know how to set up our state a certain way otherwise things break. So, let's say we change the reducer for `over18`, we'll need to remember to setup our `modifyData` call to match. the dispatches (at the time) felt like a good idea since our reducers _already_ know how to modify state, and it was nice to piggy back off of that.

once we're done with this sprint, lets take a day or two to think through it a bit.

src/app/pages/PostsFromSubreddit.jsx
schwers: From the sounds of it, we've all slipped into it after using ternaries. That seems like it be a sign that's it okay, but maybe only in this specific contest: rendering in jsx. Would you both be okay with keeping with it for now? There's a few other uses of it in the code base so I would want to either make remove all of them or review them and decide with more context which way to go.

src/app/reducers/preferencesRequest.js
schwers: So for changes, I'll move preferences out of user for now and when the sprint's over we should try the combine reducer or other approaches

schwers:
💇

src/app/actions/preferences.js
```diff
@@ -0,0 +1,77 @@
+import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+import { endpoints } from '@r/api-client';
+
+const { PreferencesEndpoint } = endpoints;
+
+export const PENDING = 'PEDNING_PREFERENCES';
```
phil303: we need to get you a spellchecker

src/app/actions/preferences.js
```diff
@@ -0,0 +1,77 @@
...
+  // optimistically update over18. If the user is not logged in
+  // we don't have to do anything else.
+  dispatch(isOver18());
+
+  if (!state.session.accessToken) {
+    return;
+  }
+
+  // Otherwise we'll call patch so the real preferences object is updated
+  await dispatch(patch({ over_18: true }));
```
phil303: I'm having trouble visualizing what this `await` does here

phil303: Also, why are we dispatching here? Why can't we just call `patch` directly or at least the contents of patch directly?

src/app/reducers/preferences.js
```diff
@@ -0,0 +1,30 @@
+import { models } from '@r/api-client';
+const { Preferences } = models;
+
+import * as loginActions from 'app/actions/login';
+import * as preferenceActions from 'app/actions/preferences';
+
+// state will be an instance of the Preferences object so we can call
```
phil303: Really don't like this. Until we come down hard one way or another (using smart models or not), can't we just jsonize this in `preferenceActions.received`?

src/app/actions/preferences.js
nramadas: i think the dispatching is to make it blocking on the server side. but, you don't need to await for that; thunk'd action automatically block on the server.

src/app/actions/preferences.js
schwers: alt-text

src/app/reducers/preferences.js
schwers: I agree, its not ideal. With every other model from api-client endpoints being a smart model I think its more important to aim for consistency. It also illustrates why we should consider moving away from the smart models.

src/app/actions/preferences.js
schwers: Whoops, the `await` is unnecessary if we're using `dispatch`. If this were to call the contents of `patch` directly we'd need it. Calling the contents of a thunk'd action seems a little odd to me, maybe because of lack of time spent in redux. Would the purpose of that be to just avoid another dispatch?

phil303:
:fish: (don't forget to revert the last commit)

schwers:
haha, will do, I'll keep waiting on the api-client patch review. Thanks!

phil303:
oh that's on me isnt it. on it.

schwers:
no worries, n.iranjan is checking as well.

PR #631, opened 2016-06-28
Use ProductionClient.css when in prod mode

ajacksified:
I kept wondering why people kept checking in broken features. This is why. Would it make sense to read a manifest from blueprints instead of hardcoding in this switch? We'd need to export a map of env to filename, and load that into `data`.

👓 @nramadas @schwers


schwers:
I think producing a manifest in build and reading it is a good idea. We're going to have to do that anyway for production building. Ideally I think we could avoid this template being aware of env and just pass a map of file-kind to path. :fish: for the production client fix for now

schwers:
This didn't end up getting merged in, but is now handled by https://github.com/reddit/reddit-mobile/pull/653.

PR #632, opened 2016-06-28
Posting 2X - not complete

phil303:
This has everything up to backend connections, tests, and polish in it. I'm making a PR now so it's a little easier to review/give feedback. I wish I had done this a little earlier since it's a pretty large patch for my first 2X one but what's done is done.

Here's the user interaction:

**Clicking write icon**
screen shot 2016-06-28 at 3 50 17 pm

**Clicking on one of the submission types**
screen shot 2016-06-28 at 3 50 23 pm

**Clicking community select showing recently visited subreddits** 
screen shot 2016-06-28 at 3 50 30 pm

**Searching for a sub / Autocomplete**
screen shot 2016-06-28 at 3 50 39 pm

**Selecting a sub**
screen shot 2016-06-28 at 3 50 46 pm

**Filling out all fields**
screen shot 2016-06-28 at 3 50 54 pm


src/app/components/AppTopNav/index.jsx
```diff
@@ -1,10 +1,15 @@
...
+
 import TopNav from 'app/components/TopNav';
+import PostSubmitModal from 'app/components/PostSubmitModal';
+import PostSubmitCommunityModal from 'app/components/PostSubmitCommunityModal';
 
 export const AppTopNav = () => (
   <UrlSwitch>
     <Case url='/login' exec={ () => null } />
+    <Case url='/submit' exec={ () => <PostSubmitModal /> } />
+    <Case url='/submit/to_community' exec={ () => <PostSubmitCommunityModal /> } />
```
phil303: should this all be in `AppMain`?

schwers: `AppMain` feels a little better, this was originally only supposed to control whether or no the TopNav displayed or not. This is an interesting usage though.

src/app/reducers/index.js
```diff
@@ -38,9 +42,11 @@ export default {
   loid,
   mail,
   messages,
+  posting,
   posts,
   postsLists,
```
phil303: just an observation, lots of things with the name "post" in them

src/app/components/PostSubmitModal/index.jsx
```diff
@@ -0,0 +1,165 @@
+import React from 'react';
+import every from 'lodash/every';
+import isEmpty from 'lodash/isEmpty';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+import { Anchor, Form } from '@r/platform/components';
+
+import cx from '../../../lib/classNames';
```
schwers: You can make this `import cx from 'lib/classNames'`

src/app/components/PostSubmitModal/index.jsx
```diff
@@ -0,0 +1,165 @@
...
+import Modal, { GREY_BUTTON, BLUE_BUTTON } from '../Modal';
+import * as postingActions from '../../actions/posting';
```
schwers: this can be `import * as postingActions from 'app/actions/posting'`

src/app/components/AppTopNav/index.jsx
schwers: You can use `Page` here as well to avoid using the thunk'd rendering. e.g.

src/app/actions/posting.js
```diff
@@ -0,0 +1,32 @@
...
+  data.resubmit = true;     // how to handle these?
+
+  // TODO: captcha?
+  // captcha - user's response to captcha challenge
+  // iden    - id of captcha
+
+  dispatch({ type: SUBMIT_PENDING });
+  const response = await PostsEndpoint.post(data);
+
+  if (response.errors) {
```
nramadas: just dropping a note that we should resolve error handling at some point

src/app/components/Modal/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+  );
+};
+
+Modal.propTypes = {
+  exitTo: T.string,
+  titleText: T.string,
+};
+
+Modal.defaultProps = {
+  type: '',
```
nramadas: should this be `exitTo`?

src/app/components/Modal/index.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { Anchor } from '@r/platform/components';
+
+import './styles.less';
+
+const T = React.PropTypes;
+
+const Modal = props => {
```
nramadas: :no_bell: 
```es6
const Modal = props => (
  <div className='Modal'>
  {/* ... */}
  </div>
);
```

src/app/components/PostSubmitCommunityModal/index.js
```diff
@@ -0,0 +1,124 @@
...
+import './styles.less';
+
+const T = React.PropTypes;
+
+const COMMUNITY_DEFAULT = 'Search for a community';
+const AUTOCOMPLETE_TITLE = 'Communities';
+const RECENT_COMMUNITY_TITLE = 'Recently visited';
+const SUBREDDIT_LIMIT = 7;
+
+class PostSubmitCommunityModal extends React.Component {
```
nramadas: couldn't this be a stateless component?

src/app/components/Modal/index.jsx
phil303: yep, ty

src/app/components/PostSubmitCommunityModal/index.js
```diff
@@ -0,0 +1,124 @@
...
+    title: T.string.isRequired,
+    subreddits: T.array.isRequired,
+    onSubredditClick: T.func.isRequired,
+    onSubredditInput: T.func.isRequired,
+    onExitModal: T.func.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    this.onSubredditInput = e => props.onSubredditInput(e.target.value);
```
nramadas: :no_bell: i don't think there is a ton of perf implications of just putting the callback inline with the template. not a big deal to declare it outside like this, but `<input onChange={ e => onSubredditInput(e.target.value) }/>` is probably ok

src/app/components/PostSubmitCommunityModal/index.js
phil303: It depends, when I have lists like for that of `renderSubredditRow`, I try to bind stuff ahead of time, which normally happens in the constructor. One callback as opposed to many and all that. I don't feel _that_ strongly, but I do think it's good practice.

src/app/actions/posting.js
schwers: Just another note that's weird about errors.

If you do something like:

```javascript
export const fetchPosts = params => async (dispatch, getState) {
  try {
    const posts = await postsApiCall(params);
    dispatch(fetched(posts));
  } catch (e) {
    dispatch(failed(e));
  }
}
```

The `catch` block catches exceptions in your rendering code, which seems confusing. It's either a good thing because we log errors that are based on api results we aren't expecting, or is bad because the code _looks_ like its just catching api errors.

package.json
```diff
@@ -18,12 +18,12 @@
     "test:watch": "npm run test -- -w"
   },
   "dependencies": {
-    "@r/api-client": "~3.14.6",
```
schwers: just a note, you'll want to bump this to ~3.16.0. Also we've been using semvar a little more conservatively -- we've had some api (code api, not api api) changes that have been breaking on minor releases, so we switched to using ~ instead of ^. What are your thoughts?

src/app/components/AppTopNav/index.jsx
phil303: As discussed, I think the overall logic for the router can/should be inverted. In other words, the router is above the main part of our app. Instead of generally displaying `AppTopNav` and on special urls, forcibly nulling it out, we have a `renderMain` type function that will render the top nav and any component passed into it.

We can make `UrlSwitch` more flexible with the concept of nested paths as @nramadas and I just discussed. This would allow us to do something like (pseudo-code):
```js
<UrlSwitch>
  <Page path="/" component={ App } />
    <Page path="r/:subredditName" component={ PostsFromSubredditPage } />
  <Page path="/login" component={ Login } />
</UrlSwitch>
```

src/app/actions/subredditAutocomplete.js
```diff
@@ -0,0 +1,52 @@
...
+let timeout;
+const FIRE_EVERY = 250;
+export const fetch = searchTerm => async (dispatch, getState) => {
+  clearTimeout(timeout);
+  timeout = setTimeout(async () => {
+    dispatch(autocompleteFetch(searchTerm));
+
+    const state = getState();
+    const apiOptions = apiOptionsFromState(state);
+    const { over18 } = state.accounts[state.user.name];
```
schwers: Now that I see this in the context of just landing my NSFW / preferences patch, this should probably use `state.preferences.over18`. the account isn't getting updated at present, partially because the accounts reducers doesn't know which account belongs to the logged in user. We could try to fix this by adding the user name to the `setOver18` action if we wanted to.

package.json
phil303: I totally agree.

src/app/components/PostSubmitCommunityModal/index.js
```diff
@@ -0,0 +1,124 @@
...
+        return { subredditName, iconUrl };
+      }),
+    };
+  }
+);
+
+const dispatcher = dispatch => ({
+  onSubredditClick: subredditName => {
+    dispatch(postingActions.selectSubreddit(subredditName));
+    dispatch(platformActions.setPage('/submit'));
```
schwers: 🔕 What was the rational on this versus going to a url like '`/submit?forSubreddit=${subreditName}`'

src/app/components/PostSubmitCommunityModal/styles.less
```diff
@@ -0,0 +1,81 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/themes/themeify';
+
+.PostSubmitCommunity {
+  &__search {
+    margin: 5px 18px 0;
+
+    .themeify({
+      border-bottom: 1px solid @theme-dark-line-color;
```
schwers: You can write `border-bottom: @theme-alt-border`

src/app/components/PostSubmitCommunityModal/index.js
```diff
@@ -0,0 +1,124 @@
...
+  renderSubredditRow({ subredditName, iconUrl }) {
+    const style = iconUrl ? { backgroundImage: `url(${iconUrl})` } : null;
+
+    return (
+      <div
+        className='PostSubmitCommunity__subreddits-row'
+        onClick={ () => this.props.onSubredditClick(subredditName) }
+        key={ subredditName }
+      >
+        <div className='PostSubmitCommunity__subreddits-icon'>
```
schwers: I think we should refactor rendering snoos into their own component either as a separate patch that's pretty soon for consistency. They have some subtler styles as that are theme dependent; in nightmode they have a black background and the keycolor is the border color. Some subreddits use the background color to change their icon as well (they have transparency in it). There's some good examples in the defaults in topnav and if you do a search through all of reddit and view communities 
image

src/app/components/PostSubmitModal/styles.less
```diff
@@ -0,0 +1,105 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/themes/themeify';
+
+.PostSubmitModal {
+  height: 40px;
+  font-size: 14px;
```
schwers: 🔕 This seems awfully close to the default

src/app/components/PostSubmitModal/styles.less
```diff
@@ -0,0 +1,105 @@
...
+      background-color: @theme-disabled-color;
+    });
+
+    &.ready {
+      background-color: @blue;
+    }
+  }
+
+  &__community {
+    margin: 7px 12px 0;
```
schwers: This seems like a 'css-smell' for code that should use `@grid-size`, from talking with T.avish in the past it sounds like most things should be using increments of 8px (@grid-size) with maybe a half-size (4px) added. looking through here all of the variables look pretty close to that so it might be worth trying it out and see how it feels.

src/app/components/TopNav/index.jsx
```diff
@@ -38,16 +35,17 @@ export const TopNav = (props) => {
 
   const settingsOpen = overlayMenu === SETTINGS_MENU;
   const communityMenuOpen = overlayMenu === COMMUNITY_MENU;
+  const submitPostOpen = overlayMenu === POST_SUBMIT;
 
-  let sideNavIcon = 'icon icon-menu icon-large';
-  if (settingsOpen) {
-    sideNavIcon += ' blue';
-  }
+  const sideNavIcon = cx('icon icon-menu icon-large', { blue: settingsOpen });
```
schwers: :upvote: for `cx`

src/app/components/PostSubmitCommunityModal/index.js
ajacksified: (should be `/r/${subredditName}/submit`, rather than a querystring, to match desktop routes.)

src/app/components/PostSubmitCommunityModal/index.js
phil303: Oh I didn't know that. Thanks Ryan. I'll tackle in a next patch if we're ok with that.

src/app/components/PostSubmitCommunityModal/index.js
phil303: Woops, this is a mistake since it breaks no-js land. I'll convert to the url.

src/app/components/PostSubmitModal/styles.less
phil303: Ah interesting. Will convert.

src/app/actions/subredditAutocomplete.js
phil303: @schwers There's a nasty race condition here that I'm not sure how to deal with. You can't get to this screen if you're not logged in but since preferences is a separate api call, we need to block until that is received. Is this a `waitForState` type of situation?

src/app/components/AppTopNav/index.jsx
schwers: 👍  I like the nested path idea, it looks nice and seems like it will help clean things up.

src/app/components/PostSubmitModal/styles.less
phil303: You're right... Looking at the mocks, it seems like I lifted the number from those. Personally I kinda like it at 14 but will bend to the whims of consistency.

src/app/actions/posting.js
nramadas: So, something i want to try in modmail when i get a chance is to have the "api" throw specific errors for api stuff. that way, the catch block would look something like:

```es6
catch (e) {
  if (e instanceof ValidationError) {
    dispatch(formErrors(e));
  } else if (e instanceof ServerError) {
    dispatch(genericError(e));
  } else {
    throw e;
  }
}
```

That way, we can still separate error logic from the success error, but non-api related errors won't get swallowed

src/app/actions/subredditAutocomplete.js
schwers: That seems right, you could `waitForState` on the preferences request `succeed` or `failure`. You could even dispatch 'fake' results (based off whatever subreddits we have in state) in `waitForState`'s error callback if you wanted to show something in the case of a slow request. How are you preventing getting to this screen if you're not logged in? The user object is fetched in handler common just like prefs, so the race-condition is technically here as well?

phil303: Yea the race condition wasn't specific to preferences since the check is just using the session validity, I just noticed it when you mentioned the change.

src/app/components/PostSubmitModal/styles.less
```diff
@@ -0,0 +1,105 @@
...
+  &__title {
+    margin: 0 12px;
+  }
+
+  &__content {
+    margin: 0 12px;
+  }
+
+  input:focus {
+    border-bottom: 1px solid @mint;
```
nramadas: i'm guessing the input has no border when not focused? won't this move everything under the input down by a pixel when it gets focused?

src/app/actions/subredditAutocomplete.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+let timeout;
+const FIRE_EVERY = 250;
+export const fetch = searchTerm => async (dispatch, getState, { waitForState }) => {
+  clearTimeout(timeout);
+  timeout = setTimeout(async () => {
+    dispatch(autocompleteFetch(searchTerm));
+
+    await waitForState(state => !isUndefined(state.preferences.over18));
```
phil303: @schwers @nramadas does this look kosher to you?

phil303:
:haircut: @schwers @nramadas

src/app/actions/subredditAutocomplete.js
nramadas: yeah, that seems ok to me

schwers: There's a minor edge case if the prefs call fails, hopefully won't happen enough to worry about. Maybe just leave a comment?

phil303:
:haircut: @schwers @nramadas

package.json
```diff
@@ -22,8 +22,8 @@
     "@r/build": "^0.6.0",
     "@r/flags": "^1.5.0",
     "@r/middleware": "~0.6.1",
```
schwers: should bump this to ~0.8.0 now

schwers:
:fish: part one

nramadas:
:tropical_fish:

PR #633, opened 2016-06-29
Add dropdowns for Posts

nramadas:

Adds dropdowns for just posts. It introduces `@r/widgets` to the codebase, so fun!

The report feature in our current dropdowns is in a painted door on purpose; our single inline form for reporting is inadequate and we've always wanted to properly build out a report feature. Since 2X has new  products for a lot of our features, I figured that would be something we sit down with design and crank out.

screens:
image


src/app/components/DropdownCover/index.jsx
```diff
@@ -0,0 +1,25 @@
+import './styles.less';
+import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+
+const T = React.PropTypes;
+
+export function DropdownCover(props) {
+  if (props.show) {
```
phil303: :no_bell: 
```js
export function DropdownCover({ show }) {
  return show ? <div className='DropdownCover'/> : null;
}
```

phil303:
@nramadas can you add some pics so we can see what this looks like?

package.json
```diff
@@ -37,6 +38,7 @@
     "koa-static": "3.0.0",
     "lodash": "4.12.0",
     "nodemon": "^1.9.1",
+    "raf": "^3.2.0",
```
phil303: are you using this?

src/app/components/DropdownCover/index.jsx
```diff
@@ -0,0 +1,25 @@
+import './styles.less';
+import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+
+const T = React.PropTypes;
+
+export function DropdownCover(props) {
```
phil303: What's this doing?

package.json
nramadas: it's a peer dependency of @r/widgets. i added it to suppress the install error

src/app/components/DropdownCover/index.jsx
nramadas: it adds a transparent-black overlay when the dropdown is opened that cover the screen behind the dropdown

package.json
phil303: got it, general question, we do peer dependencies to avoid bundling the same or close to the same package twice right? Does webpack's tree shaking help with this?

nramadas: i don't think it does. tree shaking is more about not pulling in an entire package. but i don't think tree shaking can dedupe bundled scripts within a package itself. For example, how would it tell the the `react` found in the source code of package A is the exact same react found in the source code of package B? I'm not sure it can.

phil303:
:fish: :fish:

package.json
schwers: For what its worth, I don't think tree shaking doesn't work on `commonjs` modules either. We could try to mimic a package bundler like [Rollup](https://github.com/rollup/rollup) that aims at being more aggressive about tree shaking by supplementing `package.json` ([more info](https://github.com/rollup/rollup#can-i-distribute-my-package-as-an-es6-module)) with a pointer to the ES6 source of a library. [Moment JS supports it](https://github.com/moment/moment/blob/develop/package.json#L26) for example. I think Rollup tries to convert commonjs bundles into ES6 as well, but I'm less sure if that's about tree shaking or not.

src/app/actions/posts.js
```diff
@@ -1,2 +1,50 @@
...
+  post,
+  type: TOGGLE_HIDE_RECEIVED,
+});
+
+export const toggleSavePost = post => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+  const method = post.saved ? 'del' : 'post';
+
+  try {
+    await endpoints.SavedEndpoint[method](apiOptions, { id: post.uuid });
```
schwers: What's your opinion on making this a mixin similar to [voting](https://github.com/reddit/node-api-client/blob/3X/models2/mixins/votable.js#L14)? Then Posts and Comments could each have a shared `.toggleSaved()` method.

src/app/actions/posts.js
```diff
@@ -1,2 +1,50 @@
...
+    console.error(e);
+  }
+};
+
+export const toggleHidePost = post => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+  const method = post.hidden ? 'del' : 'post';
+
+  try {
+    await endpoints.HiddenEndpoint[method](apiOptions, { id: post.uuid });
```
schwers: Same question about `saved` re a model method, could be `post.toggleHidden()`?

src/app/components/DropdownCover/index.jsx
```diff
@@ -0,0 +1,21 @@
+import './styles.less';
```
schwers: 🔕 I really really wish `.jsx` files got proper syntax highlighting.

src/app/components/Post/PostDropdown/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+    isSaved,
+    onToggleSave,
+    onToggleHide,
+  } = props;
+
+  return (
+    <div className='PostDropdown'>
+      { renderLinkRow('Permalink', 'link', permalink) }
+      { renderLinkRow(`More from r/${ subreddit }`, 'snoosilhouette', `/r/${ subreddit }`) }
+      { renderLinkRow(`${ author }'s profile`, 'user-account', `/u/${ author }`) }
```
schwers: 🔕 I've noticed you tend to put spaces in string interpolations, do you have a strong preference for it? Should we update the linter?

nramadas: i actually would prefer not to. i've been doing it because i've assumed the linter expects it. if it doesn't, i'll stop doing that.

src/app/components/Post/PostFooter/index.jsx
```diff
@@ -57,6 +59,17 @@ export default class PostFooter extends React.Component {
     this.mounted = true;
   }
 
+  componentWillReceiveProps(nextProps) {
+    // we might receive a new post model, but due to irregularities with our
+    // API, there is no guarantee that the new post model has all the same
+    // fields as the previous one. that is why we store 'saved' state in state.
+    // However, we still need a way of handling a new post model if it has
+    // saved state in it.
+    if (typeof nextProps.post.saved !== 'undefined') {
```
schwers: 🔕 I'm a little confused here, wouldn't the [action](https://github.com/nramadas/reddit-mobile/blob/9db5ea9abe2085a2a7af54cfe459041786c5b4a0/src/app/actions/posts.js#L28) handle using all of the old data, and manually toggling `saved`, thus leaving you with `saved` always being defined?

src/app/actions/posts.js
nramadas: i could look into it. is this largely to have those methods exist on the Model and avoid calling the endpoints directly?

src/app/components/Post/PostDropdown/index.jsx
phil303: I haven't run into a problem yet so I think the linter won't yell at you. I'd prefer no spaces here as well.

src/app/components/Post/PostFooter/index.jsx
nramadas: so, for posts, i was under the impression that endpoints that returned Post object had disparate fields. Because we can't really "merge" models themselves, when the old Post get's overwritten with the new Post model, they now have different fields.

What led me to believe that was this comment: https://github.com/reddit/reddit-mobile/blob/2X/src/app/components/Post/PostFooter/index.jsx#L42

So, if `saved` had to be stored in the component's state because that field might not always exist, i needed a way of overwriting it if a new model is in the props that _does_ define `saved`.

src/app/components/Post/index.jsx
```diff
@@ -168,13 +177,22 @@ const makeConnectedPostSelector = () => {
...
 const mapDispatchToProps = (dispatch, { postId }) => ({
   toggleExpanded: () => dispatch(postActions.toggleExpanded(postId)),
+  onToggleSavePost: post => dispatch(postActions.toggleSavePost(post)),
+  onToggleHidePost: post => dispatch(postActions.toggleHidePost(post)),
 });
 
+const mergeProps = (stateProps, dispatchProps) => ({
+  ...stateProps,
+  ...dispatchProps,
+  onToggleSavePost: () => dispatchProps.onToggleSavePost(stateProps.post),
```
schwers: This feels a little weird as lot of the other actions have been using ids, partially so you don't have do this merging like this. You'd have to look up the model in state though so there's a minor trade-off. What do you think of model versus id here?

src/app/components/PostsList/index.jsx
```diff
@@ -61,11 +61,12 @@ const renderPagination = (postRecords, nextUrl, prevUrl) => (
 
 const listSelector = createSelector(
   (state, props) => state.postsLists[props.postsListId],
+  state => state.posts,
   (_, props) => props.nextUrl,
   (_, props) => props.prevUrl,
-  (postsList, nextUrl, prevUrl) => ({
+  (postsList, posts, nextUrl, prevUrl) => ({
```
schwers: 🔕 Is it common in redux/reselect to use  memoized selectors? One subtle concern might be that if you upvote, or modify any posts really, this looks like it would re-render the entire list versus just the one connected post updating. Posts lists are small enough that this probably doens't matter, I'd worry more about complex comment trees though...

src/app/components/Post/index.jsx
nramadas: i don't have any strong feelings. i personally like using the actual model since the model objects are immutable, so it feels right passing them around from function to function. feels really "functional".

if you and phil prefer ids, i can switch it over to that

src/app/components/PostsList/index.jsx
nramadas: yeah, that's a good point. i guess the re-render would short-circuit at each post, since the props to that individual post would not change. however, the container's props would be different, so the container (in addition to the individual post that changed) would re-render

src/app/actions/posts.js
schwers: It also gives some minor code-reuse, we could do that here as well, so it would more so be a question of consistency. One 'feature' of smart models is it seemed like the api should provide some 'basic' functionality like this, especially if there are strange api inconsistencies we have to deal with.

src/app/components/Post/index.jsx
schwers: I have a slight preference for it purely for consistency, I'll leave it up to you and phil.

src/app/components/PostsList/index.jsx
nramadas: Something that's fun to do when testing this stuff is enabling 'pain flashing' in chrome dev tools and watching what exactly get's repainted. it's not perfect, since a react re-render !== repaint, but it gives some idea of what might be changing

schwers:
💅  just some small nits

nramadas:
:haircut:

src/app/components/Post/PostFooter/index.jsx
```diff
@@ -32,21 +35,12 @@ export default class PostFooter extends React.Component {
...
     // would force a re-render. We want this to make sure animations
     // are only rendered on the client _after_ the first render
     // so there are no bounces when you go between pages
-
-    // state.saved is requried because the api doesn't always
-    // update the state of the post properly. This is because
-    // the post doesn't always get cached, and the model update/mergeProps
-    // rely on the object that's been updated being in cache.
-    // store the saved state ourselves so that saved renders
-    // as expected on search pages and other non-listing pages
```
schwers: 🔕 I really like that we don't have to do this anymore with redux.

src/app/actions/posts.js
schwers: Just out of curiosity, what was your conclusion on this?

schwers:
:fish: still curious on the api model methods, but looks good otherwise.

PR #634, opened 2016-06-30
Add `no_ads` experiment

dwick:
👓 @prashtx 

ticket: https://reddit.atlassian.net/browse/ADS-567

(comscore/gtm are also disabled for no_ads group: https://reddit.atlassian.net/browse/ADS-579)

had to move the gtm code from the default layout to the bodylayout since it requires the user's `me.json` data.

using this existing feature config: https://github.com/reddit/reddit-private/blob/e918abff41544cfc0f67f29025869b52b1b87089/production.ini#L1229

forced variant:
screen shot 2016-06-30 at 9 10 50 am
screen shot 2016-06-30 at 9 11 32 am

normal:
screen shot 2016-06-30 at 9 11 17 am
screen shot 2016-06-30 at 9 11 04 am

### DIFF
--- a/src/featureflags.es6.js
+++ b/src/featureflags.es6.js
@@ -3,6 +3,7 @@ import Flags from '@r/flags';
 import constants from './constants';
 
 const {
+  NO_ADS,
   SMARTBANNER,
   VARIANT_RELEVANCY_TOP,
   VARIANT_RELEVANCY_ENGAGING,
@@ -10,6 +11,10 @@ const {
 } = constants.flags;
 
 const config = {
+  [NO_ADS]: {
+    url: 'experimentnoads',
+    variant: 'no_ads:treatment',
+  },
   [SMARTBANNER]: true,
   [VARIANT_RELEVANCY_TOP]: {
     and: {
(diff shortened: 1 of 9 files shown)


src/featureflags.es6.js
```diff
@@ -3,13 +3,20 @@ import Flags from '@r/flags';
...
   VARIANT_RELEVANCY_ENGAGING,
   VARIANT_RELEVANCY_RELATED,
 } = constants.flags;
 
 const config = {
+  [NO_ADS]: {
+    or: [{
+      url: 'experimentnoads',
+      variant: 'no_ads:treatment',
+    }],
```
prashtx: TLDR; the `or` is unnecessary here and uses a slightly hacky codepath on 1X (we've cleaned this up on 2x).

This will do what you want, but the `or` is unnecessary and I think technically misapplied. We interpret an object of rules as an OR, so `{url: 'foo', variant: 'bar'}` will match on either of those conditions. You're passing that as a single list entry to the `or` rule, which is the same as skipping the top-level `or` rule. We have an explicit `or` in case you need to match on two subreddits, for example, in which case you'd have two objects in the array as the `or` value.

prashtx:
Just a request to simplify the feature flag rules.
💅

dwick:
👓 @schwers 

💇 how's that?

schwers:
Looks good to me :fish:

PR #636, opened 2016-07-07
Add dropdown for Comments

nramadas:
Add a dropdown to the comments. Looks like this:

image

The edit comment toggle works partially; it tell the comment it's in edit mode, but the comment edit form itself is missing. I wanted this patch to focus on the dropdown, so i'll address the edit form in a different patch.


schwers:
Overall the patch looks good. My only concern is that the `PostDropdown/styles.less` file and this `CommentDropdown/styles.less` look almost the same (e.g. the biggest difference I see is classnames and some different icons). Dropdowns _seem_ like something we should be using some form of shared styles for consistency across the app and smaller css payloads. I'm curious to know what led to this approach and if you'd be opposed to sharing styles in some way? 💅

nramadas:
No, i'm not opposed to sharing. I would like to have a conversation on it though; my preference is to maybe create a common component out of those, but other people might prefer sharing styles. But either way, you're right, that bit could be DRY'd up a bit

nramadas:
:haircut:

src/app/components/Comment/CommentTools/index.jsx
```diff
@@ -2,104 +2,124 @@ import './styles.less';
...
 };
+
+const renderReply = ({ currentPage, id, commentReplying }) => {
+  const href = urlFromPage(currentPage, { queryParams: { commentReply: id } });
+
+  let className = 'CommentTools__reply icon icon-reply2';
+
+  if (commentReplying) {
+    className += ' CommentTools__reply__replying';
+  }
```
phil303: gonna keep pushing `classNames` :)

const classes = cx('CommentTools__reply icon icon-reply2', { 'CommentTools__reply__replying', commentReplying });

src/app/components/Comment/CommentDropdown/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+    username,
+    isSaved,
+    onEdit,
+    onDelete,
+    onToggleSave,
+  } = props;
+
+  return (
+    <Dropdown id={ id }>
+      { commentAuthor === username
```
phil303: I think it'd read nicer if you put this into a variable and `not` it when you want the opposite like on line 31.
```js
const isAuthor = commentAuthor === username
```

src/app/reducers/editingComment.js
```diff
@@ -0,0 +1,13 @@
+import * as commentActions from 'app/actions/comment';
+
+const DEFAULT = null;
+
+export default function editingComment(state=DEFAULT, action={}) {
+  switch (action.type) {
+    case commentActions.TOGGLE_EDIT_FORM: {
+      if (action.id) { return state === action.id ? null : action.id; }
+      return null;
```
phil303: :no_bell: should this and the ternary `return DEFAULT`?

phil303:
:nail_care: nothing major

src/app/components/Comment/CommentTools/index.jsx
schwers: @phil303 What's the difference between `cx` and https://github.com/JedWatson/classnames. That seems to be the one React's doc suggest. It looks like they even use `cx` as an example var name in a slightly different way than ours.

phil303: Probably almost nothing. That was the original inspiration since that used to be part of the react stdlib. I just didn't think we needed a whole module just for once simple function.

phil303:
:fish:

schwers:
:fish: might want to make the last few commit titles more specific though :)

PR #637, opened 2016-07-07
Support Ads in 2X

schwers:


schwers:
👓 @nramadas @phil303

schwers:
Depends on https://github.com/reddit/node-api-client/pull/136 if you're going to be testing locally.

src/app/actions/ads.js
```diff
@@ -0,0 +1,148 @@
...
+    dispatch(fetching(postsListId));
+
+    // We don't want to show ads if there are no posts to render, so we need
+    // to wait for the postsList to load.
+    await waitForState(state => {
+      const postList = state.postsLists[postsListId];
+      return postList && !postList.loading;
+    });
+
+    const loadedState = getState();
```
phil303: i don't believe you need to do this. I think the original reference to `state` is still good since it's only the values inside it that are immutable

src/app/middleware/adTracker.js
```diff
@@ -0,0 +1,27 @@
+import * as adActions from 'app/actions/ads';
+
+// This middleware should only be used in the client to track ad impressions.
+
+export default store => next => action => {
```
phil303: Couple things:
1. Why does this have to be middleware? Because it's having a side effect? It seems like this could just be in a thunk'd action?
2. If we decide this should be middleware, it seems awfully specific to postsLists, is there any way to make it more generic?

phil303: 3. Do we want to track the success/failure of pixel requests? I think most modern browsers have `onload` and `onerror` callbacks available for Image objects.

src/app/actions/ads.js
```diff
@@ -0,0 +1,148 @@
...
+});
+
+// Fetches an ad to render for a list of posts.
+//
+// We could look up page params from state, but having them passed in
+// is more explicit and safer if the page were to change while we await.
+export const fetchNewAdForPostsList = (postsListId, pageParams) =>
+  async (dispatch, getState, { waitForState }) => {
+    const state = getState();
+    const adRequest = state.ads[postsListId];
```
phil303: nit, should we just call this `ad`?

src/app/components/Ad/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+    const middleIsBelowTop = bottom > TOP_NAV_HEIGHT + height / 2;
+
+    if (middleIsAboveBottom && middleIsBelowTop) {
+      window.removeEventListener('scroll', this.scrollListener);
+      this.props.onAdMadeImpression();
+    }
+  }
+
+  render() {
+    return <Post { ...omit(this.props, 'adTracked', 'onAdMadeImpression') } />;
```
phil303: Do we not know what props are going into this or are there just a ton? Blacklisting some and passing the rest seems odd to me.

src/app/actions/ads.js
schwers: That seems counter-intuitive to me. I was under the impression the immutability principles of redux where that if state changes, its a new object. So if a substate of a reducer changes, doesn't the state atom, by definition, have to be a new reference?

phil303: I think just the values of individual keys change. The greater state object doesn't but test it just in case :)

phil303:
:nail_care: seems structurally sound. some small comments here and there

src/app/actions/ads.js
phil303: This would of course be different if we were using something like immutable.js but for vanilla redux, we're just using plain objects. I don't believe redux does anyone copying over of the individual key/vals as I could imagine that getting pretty expensive.

src/app/components/Ad/index.jsx
phil303: Looking at the postsList code I understand now that this is more or less a thin wrapper around a Post so it's not that odd I suppose. It still might be good to be explicit but :no_bell: for my part.

schwers: We don't really know what the props are going to be as this is essentially a pass through for <Post />. I can change it so the post props are separate so its more obvious what's happening.

src/app/actions/ads.js
phil303: @schwers welp, I'm totally wrong.
https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L135

schwers: I'm happy to test to verify. Yet this seems more fundamental than if we're using immutable.js or not -- the redux docs specify that all reducers must be pure functions. Redux's core doesn't need to worry about the copying as thats the responsibility of your reducer (just copy'ing the old reference is probably faster than a new allocation anyway). `getState` returns the last result of your store's reducer, and  our root reducer is `combineReducers`. Note that from its source, it creates a new object if any of your reducers return a new object, and assumes they'll only do that if state has changed. https://github.com/reactjs/redux/blob/master/src/combineReducers.js#L148

src/app/middleware/adTracker.js
schwers: Good points.
1. It could be a thunk'd action. I was thinking a little ahead to J.ack / P.rashan'ts event tracker patch for other events, which use middleware. So this seemed consistent. The more I think of it the more it should probably be thunk'd action.

schwers: (sorry didn't mean to make a new note)
2.  PostsLists ads are actually the only ads we have, I would be in favor of making it more generic later when we know what other kind of ads we'll have, if any.

schwers: 3. We may want to do that in the future, but not at the moment. I think the general plan is add more client side tracking once 2X is in but for now just try to get tracking parity.

src/app/actions/ads.js
schwers: We could, but if we were to write `ad.ad` it gets a little weird.

schwers: (Note that this example is being serious, `adRequest.ad` is the record that represents the post that's an ad)

phil303: should we call the reducer `adRequests` then?

src/app/middleware/adTracker.js
phil303: Regarding 2, if we go the middleware route, it seems like the action could pretty simply have a more generic key, like `adID` (something like that) and this sets us up for not having to do a refactor later to make it more generic then.

src/app/actions/ads.js
schwers: Funny enough I actually had it as `adRequests` initially but changed it. I'll change it back.

src/app/middleware/adTracker.js
schwers: We could use an `adId` in general now, regardless if we went the middleware route. The only reason its postsListId is I didn't want to generate a new id for the ads that the posts lists reducer would have to track, and the actions would have to thread another var through the fetching actions.

schwers:
💇  Some stuff is more explicit and the middleware is gone. The last few commits handle refactoring in an `adId`. This seemed better than using the `postsListId` as an `adId`, but please let me know if thats more confusing.

src/app/actions/ads.js
```diff
@@ -0,0 +1,180 @@
...
+//
+// We could look up page params from state, but having them passed in
+// is more explicit and safer if the page were to change while we await.
+export const fetchNewAdForPostsList = (postsListId, pageParams) =>
+  async (dispatch, getState, { waitForState }) => {
+    const state = getState();
+    const adRequest = state.adRequests[postsListId];
+    if (adRequest && adRequest.loading) { return; }
+
+    const adId = nextAdId();
```
phil303: you can leverage lodash's uniqueId here (basically does the same thing)

phil303:
:fish: one small comment but otherwise g2g

src/app/actions/ads.js
schwers: Oh thats pretty neat, I'll use that

PR #638, opened 2016-07-07
Adblock instrumentation

dwick:
👓 @schwers 

ticket: https://reddit.atlassian.net/browse/ADS-591

adblock disabled:
screen shot 2016-07-06 at 6 03 24 pm

blocking endpoint:
screen shot 2016-07-06 at 5 58 32 pm

blocking element:
screen shot 2016-07-06 at 6 22 40 pm
screen shot 2016-07-06 at 6 22 57 pm

### DIFF
--- a/src/lib/dom.es6.js
+++ b/src/lib/dom.es6.js
@@ -1,0 +1,18 @@
+export const isHidden = (el) => {
+  if (!el || el.hidden) {
+    return true;
+  }
+
+  const style = window.getComputedStyle(el);
+
+  if (style.display === 'none' ||
+      style.visibility === 'hidden') {
+    return true;
+  }
+
+  if (!el.parentElement) {
+    return false;
+  }
+
+  return isHidden(el.parentNode);
+}
--- a/src/lib/makeRequest.es6.js
+++ b/src/lib/makeRequest.es6.js
@@ -25,6 +25,22 @@ class Request extends superagent.Request {
 
     return p;
   }
+
+  // anything with a status of 0 is classified as a cors error by superagent.
+  // issue: https://github.com/visionmedia/superagent/issues/484
+  crossDomainError() {
+    const err = new Error('Request has been terminated. Possible causes: the network is offline, Origin is not allowed by Access-Control-Allow-Origin, the page is being unloaded, etc.');
+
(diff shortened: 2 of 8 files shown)


src/views/components/Ad.jsx
```diff
@@ -114,11 +116,17 @@ class Ad extends BaseComponent {
         loaded: true,
         ad: new models.Link(ad).toJSON(),
       });
-    }, () => {
+    }, function(err) {
```
schwers: Maybe stick with arrow syntax and do `(err) => {...` ? You shouldn't need the `.bind` that way.

src/lib/makeRequest.es6.js
```diff
@@ -25,6 +25,22 @@ class Request extends superagent.Request {
 
     return p;
   }
+
+  // anything with a status of 0 is classified as a cors error by superagent.
+  // issue: https://github.com/visionmedia/superagent/issues/484
+  crossDomainError() {
+    const err = new Error('Request has been terminated\nPossible causes: the network is offline, Origin is not allowed by Access-Control-Allow-Origin, the page is being unloaded, etc.');
```
schwers: Its probably worth double checking if these errors are making their way into the errorLog, and if they are just double check the \n won't mess with filtering the logs via `grep` or `grep -v`.

src/lib/dom.es6.js
```diff
@@ -0,0 +1,22 @@
+function isHidden(el) {
```
schwers: In some of the newer 2X code we've been writing:
```javascript
export const fnName = () => {
};
```

Then in your importing code you can do:
```javascript
import { fnName } from 'lib/someLibModule';
````

What's kinda neat about this is by exporting all the functions individually webpack can do treeshaking to remove the unneeded functions. This seems like a good place to use that convention, and it will make converting it into 2X styles a little easier.

schwers:
💅  Just some thing things, overall it looks good to go.

src/views/components/Ad.jsx
dwick: i had trouble with devtools with the arrow function. let's try again

dwick:
💇

schwers:
:fish:

PR #639, opened 2016-07-07
Login view cosmetic update

uzi:
This makes the login page look more along the
lines of the 1X login page, but a bare bones
version of it. The subcomponents are brought over
from there, and more will come in subsequent
updates.

Feel free to rip at this -- again, this part is just what
was there in 1X. Maybe we don't want these subcomponents 
where I put them.  Maybe we want them cleaned up more.

Just doing things in steps to make for smaller things to review.

### DIFF
--- a/src/app/components/Login/Input.jsx
+++ b/src/app/components/Login/Input.jsx
@@ -1,0 +1,47 @@
+import './Input.less';
+import React from 'react';
+
+import cx from 'lib/classNames';
+
+const T = React.PropTypes;
+
+function LoginInput(props) {
+  const { showTopBorder, name, type, placeholder, error, children } = props;
+  const inputClasses = cx('LoginInput__input', {
+    'error': !!error,
+    'show-top': showTopBorder,
+  });
+  const errorMessage = error ? <p className='LoginInput__error-text'>{ error }</p> : null;
+
+  return (
+    <div>
+      <div className='LoginInput'>
+        <label htmlFor={ name } className='hidden'/>
+        <input
+          id={ name }
+          name={ name }
+          type={ type }
+          className={ inputClasses }
+          placeholder={ placeholder }
+        />
+        { children }
+      </div>
+      { errorMessage }
+    </div>
+  );
+}
+
+LoginInput.propTypes = {
+  name: T.string.isRequired,
+  type: T.string,
+  placeholder: T.string.isRequired,
+  showTopBorder: T.bool,
+  error: T.string,
+};
+
+LoginInput.defaultProps = {
+  type: 'text',
+  showTopBorder: false,
+};
+
+export default LoginInput;
--- a/src/app/components/Login/Input.less
+++ b/src/app/components/Login/Input.less
@@ -1,0 +1,49 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/themes/themeify';
+
+.LoginInput {
+  padding-bottom: 1px;
+  // position relative for the optional nested button
+  // that is absolute positioned.
+  position: relative;
+
+  &__error-text {
+    color: @red;
+    padding: 3px 0  0 10px;
+  }
+
+  &__input {
+    border: none;
+    // for ios
+    -webkit-appearance: none;
+    border-radius: 0;
+
+    background-image: none;
+
+    .themeify({
+      border-bottom: @theme-border;
+    });
+
+    width: 100%;
+    background-color: transparent;
+    font-size: 1.2em;
+    padding: 10px;
+
+    &:focus {
+      outline: none;
+      border-bottom: 1px solid @blue;
+    }
+
+    &.show-top {
+      .themeify({
+        border-top: @theme-border;
+      });
+
+      background-image: none;
+    }
+
+    &.error {
+      border-bottom: 1px solid @state-danger-bg;
+    }
+  }
+}
--- a/src/app/components/Login/SquareButton.jsx
+++ b/src/app/components/Login/SquareButton.jsx
@@ -1,0 +1,50 @@
+import './SquareButton.less';
+import React from 'react';
+
+import cx from 'lib/classNames';
+
+const T = React.PropTypes;
+
+function SquareButton(props) {
+  const { enabled, modifier } = props;
+  const cls = cx('SquareButton', { 'm-disabled': !enabled, [modifier]: modifier.length });
+  const style = {
+    fontSize: props.fontSize,
+    marginTop: -(1.5 * props.fontSize) / 2,
+  };
+
+  function fireClick(e) {
+    if (props.enabled) { props.onClick(e); }
+  }
+
+  return (
+    <button
+      type={ props.type }
+      className={ cls }
+      onClick={ fireClick }
+    >
+      <div className='SquareButton__content' style={ style }>
+        { props.text }
+      </div>
+    </button>
+  );
+}
+
+SquareButton.propTypes = {
+  text: T.string.isRequired,
+  type: T.string,
+  enabled: T.bool,
+  modifier: T.string,
+  onClick: T.func,
+  fontSize: T.number,
+};
+
+SquareButton.defaultProps = {
+  type: 'button',
+  enabled: true,
+  modifier: '',
+  fontSize: 12,
+  onClick: () => {},
+};
+
+export default SquareButton;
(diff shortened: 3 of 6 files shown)


src/app/components/Login/Input.jsx
```diff
@@ -0,0 +1,51 @@
...
+import React from 'react';
+
+const T = React.PropTypes;
+
+function LoginInput(props) {
+  const { showTopBorder, name, type, placeholder,
+          onChange, error, children, value } = props;
+  const showTop = showTopBorder ? 'show-top' : '';
+
+  let errorClass = '';
```
phil303: this might be a good opportunity to use the `classNames` helper found in `app/lib`.
```js
import cx from 'app/lib/classNames';

// ...

const errorClass = cx({ 'error': !!error });
const errorMessage = !!error ? <p className='LoginInput__error-text'>{ error }</p> : null;
```

src/app/components/Login/Input.jsx
```diff
@@ -0,0 +1,51 @@
...
+    </div>
+  );
+}
+
+LoginInput.propTypes = {
+  name: T.string.isRequired,
+  type: T.string,
+  placeholder: T.string.isRequired,
+  showTopBorder: T.bool,
+  error:T.string,
```
phil303: nit spacing. Also, anything not required should probably have a default.

src/app/components/Login/Input.jsx
```diff
@@ -0,0 +1,51 @@
+import './Input.less';
+import React from 'react';
```
phil303: Generally I adhere to a python-esque import order (stdlib > third party > app specific). I'm not sure if we do something special since it's a less file but I've been grouping those with the app specific imports.

src/app/components/Login/SquareButton.less
```diff
@@ -0,0 +1,31 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/mixins/importAll';
```
phil303: general question for anyone - what's this do?

phil303:
@uzi can you also put some pics up so we can see the changes?

src/app/components/Login/Input.less
```diff
@@ -0,0 +1,53 @@
...
+  &__error-text {
+    color: @red;
+    padding: 3px 0  0 10px;
+  }
+
+  &__input {
+    border: none;
+    // for ios
+    -webkit-appearance: none;
+    border-top-right-radius: 0px;
```
phil303: does this need to be enumerated? Can't you just `border-radius: 0`?

https://github.com/reddit/reddit-mobile/pull/632/files#diff-55009b8917ee73d3c73f2b1f20bd6772R93

src/app/components/Login/SquareButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+    marginTop: -(1.5 * fontSize) / 2,
+  };
+
+  return <div className='SquareButton__content' style={ style }>{ content }</div>;
+}
+
+function SquareButton(props) {
+  let cls = 'SquareButton';
+  if (!props.enabled) { cls += ' m-disabled'; }
+  if (props.modifier.length) { cls += ` ${props.modifier}`; }
```
phil303: another opportunity for `classNames`:

```js
const { enabled, modifier } = props;
const classes = cx('SquareButton', { 'm-disabled': !enabled, [modifier]: modifier.length });
```

src/app/components/Login/SquareButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+    if (props.enabled) { props.onClick(e); }
+  }
+
+  return (
+    <button
+      type={ props.type }
+      className={ cls }
+      onClick={ fireClick }
+    >
+      { content(props.text, props.fontSize) }
```
phil303: :no_bell: I might argue that you don't need the `content` function (I generally like to keep my jsx as together as possible) but you could do something like this. Totally not a show-stopper though.
```js
<button
  type={ props.type }
  className={ cls }
  onClick={ fireClick }
>
  <div className='SquareButton__content' style={ makeStyle(props.fontSize) }>
    { props.text }
  </div>
</button>

const makeStyle = fontSize => ({ fontSize, marginTop: -(1.5 * fontSize) / 2 });
```

src/app/components/Login/styles.less
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  .SnooIcon {
+    margin: auto;
+    height: @size;
+    width: @size;
+    position: relative;
+  }
+
+  &__Form {
```
phil303: :no_bell: not sure if there's a style guide for this but these have generally been lowercased in the patches I've done so the class would be `Login__form`.

phil303:
:nail_care: also check out your build failures. There's some small clean up stuff in there.

src/app/components/Login/Input.less
nramadas: also, our build step is supposed to auto-enumerate

phil303: say whatttt? do we deliver different payloads for different browsers? I can imagine that bloating our payloads quite a bit if so.

src/app/components/Login/SquareButton.jsx
```diff
@@ -0,0 +1,52 @@
+import './SquareButton.less';
+import React from 'react';
+
+const T = React.PropTypes;
+
+function content(content, fontSize) {
```
nramadas: super-nit, but it's nice to have the component's function at the top. So putting `function SquareButton` at the top and putting the `content` template at the bottom.

src/app/components/Login/Input.less
nramadas: no, it just auto-prefixes if necessary. i'm assuming if something needs to be enumerated, it'll do that too (but i could be off in my assumption) https://github.com/postcss/autoprefixer.  it shouldn't add much bloat.

src/app/components/Login/Input.jsx
schwers: Most components have it imported first. The reasoning when I started on 2X was that this import is special and only really here for webpack's bundler -- it's not really used like the other imports are. I'm open to re-ordering it though. I think it should probably stay first in this patch for consistency though.

src/app/components/Login/Input.less
schwers: Not different payloads, but there's an auto-prefixer step built into our build system. It's currently disabled but we should be able to have it working soon. There wasn't a notion of different payloads but that would be a nice addition.

phil303: ok that makes more sense to me - although I'm curious how many mweb browsers actually need the prefix. I thought niranjan was talking about the enumeration of the shorthand properties.

phil303: side note - I'd love to get to the point where we ARE delivering different payloads to browsers.

src/app/components/Login/SquareButton.less
schwers: `(reference)` just imports the definitions and not the css rules. e.g. importing a file with mixins gets an empty class thats the name of the mixin, but if you use `(reference)` you only import the definition of the mixin.

phil303: oh, specifically, I meant what does `importAll` do. It seems like a bunch of stuff not really necessary in this file.

src/app/components/Login/Input.jsx
uzi: Cool, using this one

src/app/components/Login/Input.jsx
uzi: Is that needed?  I thought if it doesn't then it defaults to `undefined`.  Is it preferred to be explicit?

uzi:
:haircut: Ok, I think that takes care of the feedback let me know. Thanks!

phil303:
:fish: but can you or @schwers can you address this question? https://github.com/reddit/reddit-mobile/pull/639#discussion_r69986278. I have a feeling that although innocous, it's not really necessary to that file or `Login/styles.less`

src/app/components/Login/SquareButton.less
schwers: so we adopted `importAll` for basically index files in less. I might be missing something but I don't see what mixins this is using, do you know if it builds without this line at all?

schwers:
The build error is related to `extract-text-webpack-plugin`. Yesterday there was a change to webpack that removed a depreciated method `Chunk.entry` was removed. You should be able to bump node-build to `0.7.1` [patch](https://github.com/reddit/node-build/commit/68377a3e9d77f46c7dc62fba380f9325b4755cab)

PR #642, opened 2016-07-11
Fix typos in the nextcontent_mweb feature configs

prashtx:
If we build more experiments into mweb 1X, we should move to the new `@r/flags` version, which handles boolean rules better than the interim versions in `featureflags.es6.js`.


prashtx:
👓 @schwers

schwers:
this would be a nice place to add tests too in the future. :fish:

PR #643, opened 2016-07-11
Restrict nextcontent_mweb experiment to r/pokemongo

prashtx:
We're temporarily restricting this test to one subreddit, so that we can increase the bucket sizes and test that subreddit's traffic specifically.


schwers:
:fish:

PR #647, opened 2016-07-15
Post submit - backend/tests/captcha box

phil303:
:eyeglasses: @schwers

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,11 +18,11 @@
     "test:watch": "npm run test -- -w"
   },
   "dependencies": {
-    "@r/api-client": "~3.16.0",
-    "@r/build": "^0.7.1",
+    "@r/api-client": "~3.18.0",
+    "@r/build": "~0.7.2",
     "@r/flags": "^1.5.0",
     "@r/middleware": "~0.8.0",
-    "@r/platform": "~0.9.0",
+    "@r/platform": "~0.10.0",
     "@r/private": "~0.1.2",
     "@r/redux-state-archiver": "~0.0.5",
     "@r/widgets": "^0.1.2",
--- a/src/app/actions/posting.js
+++ b/src/app/actions/posting.js
@@ -1,31 +1,56 @@
-import { endpoints } from '@r/api-client';
+import url from 'url';
+
+import { endpoints, errors } from '@r/api-client';
+import * as platformActions from '@r/platform/actions';
+
+import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+
 
 const { PostsEndpoint } = endpoints;
+const { ValidationError, BadCaptchaError} = errors;
 
-export const SUBREDDIT_SELECT = 'POSTING__SUBREDDIT_SELECT';
 export const FIELD_UPDATE = 'POSTING__FIELD_UPDATE';
+export const CLOSE_CAPTCHA = 'POSTING__CLOSE_CAPTCHA';
 export const SUBMIT_PENDING = 'POSTING__SUBMIT_PENDING';
 export const SUBMIT_SUCCESS = 'POSTING__SUBMIT_SUCCESS';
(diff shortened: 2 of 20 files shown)


src/app/actions/posting.js
```diff
@@ -1,31 +1,56 @@
-import { endpoints } from '@r/api-client';
+import url from 'url';
```
phil303: Note: I ended up going with the url module. I haven't looked at impact on payload size but I took a look at the source and think it's a fairly small package. I don't like the idea of rolling our own parser because of the edge cases that might exist that I don't know about.

I'm totally open to discussing this though.

src/app/components/DropdownCover/index.jsx
```diff
@@ -14,8 +14,9 @@ DropdownCover.propTypes = {
 };
 
 const selector = createSelector(
-  state => state.widgets.tooltip.id,
-  tooltipId => ({ show: !!tooltipId }),
+  state => !!state.widgets.tooltip.id,
+  state => !!state.posting.captchaIden,
```
phil303: @nramadas, you might have interest in this. I'm basically "overloading" your DropdownCover class to also appear if there's a captcha box. I _think_ this should be fine but would love to get your thoughts.

nramadas: Yeah, that seems fine

src/app/components/PostSubmitModal/index.jsx
```diff
@@ -145,27 +160,77 @@ const mapStateToProps = createSelector(
...
+    readyToPost,
+    subreddit,
+  } = stateProps;
+
+  const {
+    onFieldUpdate,
+    onCloseCaptcha,
+    onGetNewCaptcha,
+    _onSubmit
+  } = dispatchProps;
```
nramadas: is this just to avoid having `_onSubmit` appear in the component's props? tbh, i don't know if it's a big deal. just doing a normal object clone+extend looks way cleaner and has less boilerplate, so it might be better to accept the `_onSubmit`.

phil303: Not exactly, although that's a side effect. It's more the general principle of documenting what's going into the props, although I suppose propTypes takes care of that. To me this is a clear signal, because my trust for propTypes being up to date is not.

src/app/components/CaptchaBox/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+    captchaImgUrl,
+    onCloseCaptcha,
+    onFieldUpdate,
+    onGetNewCaptcha,
+    onSubmit,
+  } = props;
+
+  return (
+    <div className='CaptchaBox'>
+      <div onClick={ onCloseCaptcha } className='icon icon-nav-close icon-large'></div>
```
schwers: :no_bell: No bell because the this isn't in eslint. Tehcnically the styleguide wants you to use a self-closing tag if there are no children here. https://github.com/reddit/styleguide/blob/master/javascript/react/README.md#tags (We missed getting that into the linter somehow, I'm totally okay with leaving it and doing another patch just for this rule)

package.json
```diff
@@ -18,8 +18,8 @@
     "test:watch": "npm run test -- -w"
   },
   "dependencies": {
-    "@r/api-client": "~3.16.0",
-    "@r/build": "^0.7.1",
+    "@r/api-client": "~3.18.0",
+    "@r/build": "~0.7.2",
```
schwers: :no_bell: Nice, thanks for fixing the extract bug

schwers:
:fish:

src/app/components/CaptchaBox/index.jsx
phil303: Oh interesting. I wonder how many of these I have in this and the previous patch... Let me do a quick scan

nramadas:
:tropical_fish:

PR #648, opened 2016-07-18
Redirect m. users to desktop if cookie exists

phil303:
Right now, mweb users who go to a www url will have their mweb redirect cookie respected. Mweb users who go to an m dot url will not. This patch adds the ability to respect that cookie on the mweb server.

:eyeglasses: @schwers

### DIFF
--- a/src/server/index.jsx
+++ b/src/server/index.jsx
@@ -9,6 +9,7 @@ import koaStatic from 'koa-static';
 import bodyParser from 'koa-bodyparser';
 import csrf from 'koa-csrf';
 import compress from 'koa-compress';
+import urlParser from 'url';
 
 import koasession from 'koa-session';
 import StatsdClient from 'statsd-client';
@@ -216,6 +217,7 @@ class Server {
       server.use(koaStatic(`${__dirname}/../../build`));
     }
 
+    server.use(this.checkForDesktopRedirect(app));
     server.use(this.checkToken(app));
     server.use(this.convertSession(app));
     server.use(this.setLOID(app));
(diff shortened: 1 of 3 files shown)


src/constants.es6.js
```diff
@@ -1,4 +1,5 @@
 export default {
+  DESKTOP_BASE: 'https://www.reddit.com',
```
schwers: I would use `config.reddit` from `/src/config.es6.js` here to match the redirection logic in `assets/js/clientEvents.es6.js`.

src/server/index.jsx
```diff
@@ -450,6 +452,24 @@ class Server {
...
+    return function * (next) {
+      // This cookie name is a little confusing since it's a vestige of a
+      // desktop experiment. This is saying, "don't redirect to mweb from
+      // desktop". From mweb's perspective, if this is true, we want to
+      // redirect to desktop.
+      const noMweb = this.cookies.get('mweb-no-redirect');
+      const url = urlParser.parse(this.request.href);
+      const subdomain = url.host.split('.')[0];
+      if (noMweb && subdomain === 'm') {
+        const desktopUrl = constants.DESKTOP_BASE + url.path;
```
schwers: I think the linter will want you to use interpolation here instead of concatenation

src/server/index.jsx
```diff
@@ -450,6 +452,24 @@ class Server {
     };
   }
 
+  checkForDesktopRedirect () {
+    return function * (next) {
+      // This cookie name is a little confusing since it's a vestige of a
+      // desktop experiment. This is saying, "don't redirect to mweb from
+      // desktop". From mweb's perspective, if this is true, we want to
+      // redirect to desktop.
```
umbrae: 🔕  I appreciate the comment but am sad we were so shortsighted in hindsight. :(

schwers:
💅 Just two little nits, overall it looks right

src/server/index.jsx
phil303: linter doesn't seem to care but i can change it to that if we do

phil303:
:haircut:

src/server/index.jsx
schwers: odd, seems like the `prefer-template` rule only doesn't care if both parts of the concatenation are variables but will complain if one is a string and the other is a var. seems a little odd on the rules part so I'll leave it up to you

schwers:
:fish:

src/server/index.jsx
phil303: cool i think it's fine

PR #650, opened 2016-07-20
2X error logging

schwers:
Depends on https://github.com/reddit/node-api-client/pull/139

The basic principle of this patch is that we want to have as much info for debugging as possible. Thus most of the work for errorLogging is done by the errorLogger middlware. It intercepts actions that are considered failures to log errors. This seems better than only using a try/catch or Promise.catch approach because the action names will give us more context and seems to set us up nicely for error-handling. It also attaches a `.catch` handler to any promise it sees dispatched as well as wrapping `next(action)` in a try/catch.

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -7,14 +7,46 @@ import isEmpty from 'lodash/isEmpty';
 
 import { isLocalStorageAvailable } from '@r/redux-state-archiver';
 
-import routes from 'app/router';
 import App from 'app';
+import config from 'config';
+import errorLog from 'lib/errorLog';
+import routes from 'app/router';
 import reducers from 'app/reducers';
+import reduxMiddleware from 'app/reduxMiddleware';
 import Session from 'app/models/Session';
 
+// register window.onError asap so we can catch errors in the client's init
+window.onerror = (message, url, line, column) => {
+  errorLog({
+    message,
+    url,
+    line,
+    column,
+    userAgent: window.navigator.userAgent,
+    requestUrl: window.navigator.userAgent,
+  }, {
+    hivemind: config.statsURL,
+  });
+};
+
+// This isn't supported in mobile browsers right now but it is in chrome.
+// Having it will give us better logging in debug (for promises that don't have
+// a .catch handler). Maybe mobile browsers will support it soon as well.
+window.onunhandledrejection = rejection => {
+  errorLog({
+    rejection,
+    userAgent: window.navigator.userAgent,
+    requestUrl: window.navigator.userAgent,
+  }, {
+    hivemind: config.statsURL,
+  });
+};
+
+// start the app now
 const client = Client({
   routes,
   reducers,
+  reduxMiddleware,
   modifyData: data => {
     // TODO if we start not using shell rendering in a serious way,
     // we'll need to unserialize all of the api models. This should
@@ -29,6 +61,7 @@ const client = Client({
     data.preferences = models.Preferences.fromJSON(data.preferences);
 
     data.collapsedComments = {};
+    data.meta.env = 'CLIENT';
 
     if (isLocalStorageAvailable()) {
       try {
--- a/src/app/actions/accounts.js
+++ b/src/app/actions/accounts.js
@@ -1,16 +1,29 @@
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
-import { endpoints } from '@r/api-client';
-
+import { endpoints, errors } from '@r/api-client';
 const { AccountsEndpoint } = endpoints;
+const { ResponseError } = errors;
 
 export const FETCHING_ACCOUNT = 'FETCHING_ACCOUNT';
-export const fetching = options => ({ type: FETCHING_ACCOUNT, ...options });
+// options is for the accounts endpoint, its properties are
+// name: T.string.isRequired,
+// loggedOut: T.bool, ( true if the user is loggedOut)
+export const fetching = options => ({
+  type: FETCHING_ACCOUNT,
+  ...options,
+});
 
 export const RECEIVED_ACCOUNT = 'RECEIVED_ACCOUNT';
 export const received = (options, apiResponse) => ({
   type: RECEIVED_ACCOUNT,
   ...options,
   apiResponse,
+});
+
+export const FAILED = 'FAILED_ACCOUNT';
+export const failed = (options, error) => ({
+  type: FAILED,
+  options,
+  error,
 });
 
 export const fetch = options => async (dispatch, getState) => {
@@ -22,6 +35,14 @@ export const fetch = options => async (dispatch, getState) => {
   dispatch(fetching(options));
   const query = { user: name, loggedOut };
 
-  const apiResponse = await AccountsEndpoint.get(apiOptionsFromState(state), query);
-  dispatch(received(options, apiResponse));
+  try {
+    const apiResponse = await AccountsEndpoint.get(apiOptionsFromState(state), query);
+    dispatch(received(options, apiResponse));
+  } catch (e) {
+    if (e instanceof ResponseError) {
+      dispatch(failed(options, e));
+    } else {
+      throw e;
+    }
+  }
 };
--- a/src/app/actions/commentsPage.js
+++ b/src/app/actions/commentsPage.js
@@ -1,35 +1,40 @@
+import { endpoints, errors } from '@r/api-client';
+const { CommentsEndpoint, PostsEndpoint } = endpoints;
+const { ResponseError } = errors;
 import { some } from 'lodash/collection';
 
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import getSubreddit from 'lib/getSubredditFromState';
 import { cleanObject } from 'lib/cleanObject';
 import features from 'app/featureFlags';
-import { endpoints } from '@r/api-client';
+
 import { paramsToCommentsPageId } from 'app/models/CommentsPage';
 import { paramsToPostsListsId } from 'app/models/PostsList';
-import { fetching as fetchingPosts, received as receivedPosts } from 'app/actions/postsList';
+import * as postListActions from 'app/actions/postsList';
 import { flags } from 'app/constants';
-
-const { CommentsEndpoint, PostsEndpoint } = endpoints;
 
 const {
   VARIANT_NEXTCONTENT_BOTTOM,
 } = flags;
 
 export const FETCHING_COMMENTS_PAGE = 'FETCHING_COMMENTS_PAGE';
-
-export const fetchingCommentsPage = (commentsPageId, commentsPageParams) => ({
+export const fetching = (commentsPageId, commentsPageParams) => ({
   type: FETCHING_COMMENTS_PAGE,
   commentsPageId,
   commentsPageParams,
 });
 
 export const RECEIVED_COMMENTS_PAGE = 'RECEIVED_COMMENTS_PAGE';
-
 export const received = (commentsPageId, apiResponse) => ({
   type: RECEIVED_COMMENTS_PAGE,
   commentsPageId,
   apiResponse,
+});
+
+export const FAILED = 'FAILED_COMMENTS_PAGE';
+export const failed = (commentsPageId, error) => ({
+  type: FAILED,
+  error,
 });
 
 export const VISITED_COMMENTS_PAGE = 'VISITED_COMMENTS_PAGE';
@@ -46,14 +51,22 @@ export const fetchCommentsPage = commentsPageParams => async (dispatch, getState
 
   if (commentsPage) { return; }
 
-  dispatch(fetchingCommentsPage(commentsPageId, commentsPageParams));
+  dispatch(fetching(commentsPageId, commentsPageParams));
 
   // note that the comments endpoint returns the post, so we don't have to also
-  // fetch that somewere else. it's in the api response so the post's reducer will
+  // fetch that somewhere else. it's in the api response so the post's reducer will
   // will automatically update the post slice of the store
-  const apiOptions = apiOptionsFromState(state);
-  const apiResponse = await CommentsEndpoint.get(apiOptions, commentsPageParams);
-  dispatch(received(commentsPageId, apiResponse));
+  try {
+    const apiOptions = apiOptionsFromState(state);
+    const apiResponse = await CommentsEndpoint.get(apiOptions, commentsPageParams);
+    dispatch(received(commentsPageId, apiResponse));
+  } catch (e) {
+    if (e instanceof ResponseError) {
+      dispatch(failed(commentsPageId, e));
+    } else {
+      throw e;
+    }
+  }
 };
 
 export function relevantContentPostsParams({ subredditName }) {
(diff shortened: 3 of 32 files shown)


schwers:
Whoops, hit enter a bit too soon. `lib/errorLog` has some updates to do more of the exception and rejection parsing for logs internally as opposed to being done adhoc like 1X

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+
+  return store => next => action => {
+    actionStack.push(action);
+
+    // Check for specific error actions, this gives insight into what kind of
+    // error happened. I.E. its easier to differentiate between a specific
+    // action or endpoint failing than seeing its equivalent stack trace.
+    const error = checkForSpecificErrors(action);
+    if (error) {
+      logErrorWithConfig({
```
schwers: I wasn't sure if it was better to split this out into separate functions if we want to add or remove different types of logging. I'd be happy to to turn this into a function that did the `baseErrorPayload` work so this looked like `logErrorWithConfig(error, store.getState(), actionStack.toString())`

schwers:
Worth mentioning something on actions. This patch relies on some new explicit error actions, so I did some style cleanup of action creators. At some point we talked about how a multiline style is more readable so I updated that and some other minor things for consistency. e.g.
```javascript
export const received = (id, results) => ({ type: TYPE, id, results })
```
becomes
```javascript
export const received = (id, results) => ({
  type: TYPE,
  id,
  results,
});
```

schwers:
👓 @phil303 @nramadas

src/Client.js
```diff
@@ -7,14 +7,43 @@ import isEmpty from 'lodash/isEmpty';
...
+    line,
+    column,
+    userAgent: window.navigator.userAgent,
+    requestUrl: window.navigator.userAgent,
+  }, {
+    hivemind: config.statsURL,
+  });
+};
+
+window.onunhandledrejection = rejection => {
```
phil303: Maybe leave a comment here about why we need this since it seems no mobile browsers actually support it but the latest desktop chrome does.
screen shot 2016-07-20 at 1 55 24 pm

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+// will be frequently generated enough so that a little ring-buffer is worth
+// it to reduce memory churn as actions pass through the store.
+// It also handles logging the state of the redux store from the perspective
+// of actions. This makes errors log'd out on the server have the most recent N
+// number of actions. For now this only tracks the action names to save space.
+// There are some actions we want to know about to know about to understand
+// a little more about what's actually happening, so this takes an optional
+// array of action types for actions we should keep around in their full state.
+class ActionStack {
+  constructor(size, typeWhiteList=[]) {
```
phil303: This name could probably be a little more informative. My first thought when seeing it was that only actions in the white list are being recorded, which isn't really the case. How about something like `expandedInfoList`

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+};
+
+// make a modulus such that if you go backwards past 0, you get that distance
+// away from the end of a ring buffer. e.g.
+//    ringModulus(-1, 10) => 9
+//    ringModulus(-2, 10) => 8
+//    ringModulus(-3, 10) => 7
+//    ringModulus(10, 10) => 0
+//    ringModulus(11, 10) => 1
+const ringModulus = (i, size) => {
```
phil303: Hey I learned another new thing about JS today. Python handles this natively:
```python
// python
In [1] -1 % 10
Out[1] 9

File: src/app/reduxMiddleware/errorLogger.js [hunk 5]
```diff
@@ -0,0 +1,190 @@
...
+// Use this to keep the list of most recent actions. This assumes actions
+// will be frequently generated enough so that a little ring-buffer is worth
+// it to reduce memory churn as actions pass through the store.
+// It also handles logging the state of the redux store from the perspective
+// of actions. This makes errors log'd out on the server have the most recent N
+// number of actions. For now this only tracks the action names to save space.
+// There are some actions we want to know about to know about to understand
+// a little more about what's actually happening, so this takes an optional
+// array of action types for actions we should keep around in their full state.
+class ActionStack {
```
phil303: So I think this can be more simply solved by using a implementing a stack of fixed length and popping off the last item when you get a new one. In other words, create an array of N length. When you get a new item, `unshift` onto the front and `pop` off the back. Then when it comes to calling `toArray`, you don't need to keep track of your current position or any of the logic you have in there. You just read from front to back. Wham bam thank you ma'am.

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+      const { error } = action;
+      if (error) {
+        return error;
+      }
+    }
+  }
+};
+
+const baseErrorPayload = (state, actionStack) => {
+  const { env, userAgent: metaUserAgent } = state.meta;
```
phil303: i think it's ok to just call it `userAgent` here.

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+    case wikiActions.FAILED: {
+      const { error } = action;
+      if (error) {
+        return error;
+      }
+    }
+  }
+};
+
+const baseErrorPayload = (state, actionStack) => {
```
phil303: what if you just passed `meta` into this, or better yet, just `env` and `userAgent`

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+
+    // async actions all have an associated Promises. To catch errors in these
+    // Promises we need to attach a .catch handler, otherwise they'll get
+    // turned into PromiseRejectionEvents which are pretty different than Errors
+    // (they don't even show up in window.onerror). This way we can catch
+    // any unhandled exceptions in async actions here.
+    // e.g. if we fetch some data and it parsed correctly, it could still
+    // cause an error in reducers or React's rendering. both of those would
+    // be caught here.
+    if (action instanceof Promise) {
```
phil303: Should this be an `else if`?

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+        // headPtr is less than the stack size.
+      }
+    }
+
+    return orderedActions;
+  }
+
+  // string representation of the action stack with most recent actions first.
+  toString() {
+    return `Redux Action Stack: ${this.toArray().map(x => JSON.stringify(x)).join(', ')}`;
```
phil303: Can just do:
```js
this.toArray().map(JSON.stringify)
```

src/config.js
```diff
@@ -0,0 +1,55 @@
+/*eslint max-len: 0*/
+
+// This file ports config from 1X mweb to 2X, for now global message has been
```
phil303: I feel like these 1X to 2X migration messages are only really valuable to the reviewers but in the short to medium will be obsoleted.

src/lib/errorLog.js
```diff
@@ -3,57 +3,114 @@ import makeRequest from './makeRequest';
...
+    parsedDetails = { ...parsedDetails, ...parseError(error) };
+  } else if (rejection) {
+    parsedDetails = { ...parsedDetails, ...parseRejection(rejection) };
+  }
 
-  if (/iPhone/i.test(agent) || /iPad/i.test(agent) || /iPod/i.test(agent)) {
-    if (/CriOS/i.test(agent)) {
-      return 'ios-chrome';
+  const formattedLog = formatLog(parsedDetails);
+  console.log(formattedLog);
```
phil303: Just checking, do you mean to `console.log` here?

src/lib/errorLog.js
```diff
@@ -3,57 +3,114 @@ import makeRequest from './makeRequest';
...
+
+  if (config.debugLevel === 'info') {
+    if (details.error && details.error.stack) {
```
phil303: why not just one condition?
```js
if (config.debugLevel === 'info' && details.error && details.error.stack)
```

src/lib/errorLog.js
```diff
@@ -3,57 +3,114 @@ import makeRequest from './makeRequest';
 
 const { ResponseError } = errors;
 
-function simpleUA(agent) {
-  if (/server/i.test(agent)) { return 'server'; }
 
-  // Googlebot does silly things like tell us it's iPhone, check first
-  // see https://googlewebmastercentral.blogspot.com/2014/01/a-new-googlebot-user-agent-for-crawling.html
-  if (/Googlebot/i.test(agent)) { return 'googlebot-js-client'; }
+const isAPIFailure = details => (details.error && details.error instanceof ResponseError);
```
phil303: :no_bell: if `details` is always going to exist, can just do
```js
const isAPIFailure = details => details.error instanceof ResponseError;
```

src/lib/errorLog.js
```diff
@@ -3,57 +3,114 @@ import makeRequest from './makeRequest';
...
+
+  return {
+    message: `Error: ${error.message}`,
+  };
+};
+
+const parseRejection = rejection => {
+  // rejection should be an instanceof PromiseRejectionEvent
+  if (rejection.reason && rejection.reason.stack) {
+    if (rejection.reason.stack) {
```
phil303: what's going on with the double if block here?

phil303:
:nail_care: Also, you have to rebase.

src/app/actions/accounts.js
```diff
@@ -1,10 +1,13 @@
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
-import { endpoints } from '@r/api-client';
-
+import { endpoints, errors } from '@r/api-client';
 const { AccountsEndpoint } = endpoints;
+const { ResponseError } = errors;
 
 export const FETCHING_ACCOUNT = 'FETCHING_ACCOUNT';
-export const fetching = options => ({ type: FETCHING_ACCOUNT, ...options });
+export const fetching = options => ({
```
nramadas: are the keys in options always the same? if so, can we add a comment to document what those keys are?

src/app/actions/commentsPage.js
```diff
@@ -1,37 +1,46 @@
...
-import { endpoints } from '@r/api-client';
+
 import { paramsToCommentsPageId } from 'app/models/CommentsPage';
 import { paramsToPostsListsId } from 'app/models/PostsList';
-import { fetching as fetchingPosts, received as receivedPosts } from 'app/actions/postsList';
+import {
+  fetching as fetchingPosts,
+  received as receivedPosts,
+  failed as failedPosts,
+} from 'app/actions/postsList';
```
nramadas: :no_bell: no love for `import *`?

src/app/reduxMiddleware/errorLogger.js
schwers: whoops, yeah refactor bug

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -0,0 +1,190 @@
...
+      const action = this.actions[i];
+      if (action) {
+        orderedActions.push(action);
+      } else {
+        break; // if we hit a null, we're at the end of the array and
+        // headPtr is less than the stack size.
+      }
+    }
+
+    return orderedActions;
```
nramadas: this seems a little complicated. couldn't you just do:
```es6
toArray() {
  return this.actions.slice(this.headPtr).concat(this.action.slice(0, this.headPtr));
}
```
maybe i'm missing an edge case, but it looks like all you want to do is break the right and convert it into an array, which i believe the above does.

src/app/reduxMiddleware/errorLogger.js
nramadas: so, i actually think a ring is a little easier than a queue, and (while it doesn't matter at our scale) much more performant. with a queue, you have to keep track of its size as it grows, and only unshift after it hits a certain size. not hard to do, but slightly more error prone than a simple modulo

phil303: ? Just initialize an array of fixed length (every element starts `undefined`), which is what he's already doing. Why start with the complicated solution??

nramadas: ugh, yeah, you're right. i had a brain fart there. but still, i think the Ring is nicer without adding much complexity. consider:
```es6
class RingQueue {
  constructor(size) {
    this.queue = new Array(size);
    this.head = 0;
    this.size = size;
  }

  push(val) {
    this.queue[this.head % this.size] = val;
    this.head++;
  }

  values() {
    return this.queue.slice(this.head)
      .concat(this.queue.slice(0, this.head))
      .filter(x => !!x);
  }
}

class NormalQueue {
  constructor(size) {
    this.queue = new Array(size);
  }

  push(val) {
    this.queue.unshift(val);
    this.queue.pop();
  }

  values() {
    return this.queue.slice(this.head)
      .concat(this.queue.slice(0, this.head))
      .filter(x => !!x);
  }
}
```
The ring is not any more complex than the other queue, but is a lot faster. :shrug: i'm not married to a ring queue however, so i wouldn't mind seeing a normal queue.

phil303: Your `values` method is wrong but I get your drift. I'm not sure about "a lot faster" though. I'm not even sure about faster. I'm just not seeing the payoff but I'm not as against it as I initially was.

phil303: Long story short, either of these is a win Schwers. Niranjan, as usual, is the king of cleaning up code.

src/app/reduxMiddleware/errorLogger.js
schwers: I think you need a reverse in there but yeah this would work. Honestly I kept it this way because it seemed more in the spirit of "keeping it performant". But I'm okay with updateing for clarity.

src/lib/errorLog.js
schwers: another refactor bug

src/app/reduxMiddleware/errorLogger.js
schwers: I need the platform state to make `requestUrl` as well, so passing all of state seems better because 

```javascript
logErrorWithConfig({
  ...baseErrorPayload(store.getState(), actionStack),
   error,
});
``` (repeated 3 times)

Would have to look this instead:

```javascript
const state = store.getState();
const { env, userAgent } = state.meta;
const { currentPage } = state.platform;
logErrorWithConfig({
   ...baseErrorPayload(env, userAgent, currentPage, actionStack),
   error,
});
```

Not sure if you saw my other comment but one question I had was if you thought 
```javascript
logErrorWithConfig({
  ...baseErrorPayload(store.getState(), actionStack),
   error,
});
```

Should just be turned into something like:
```javascript
logErrorWithConfig(error, store.getState(), actionStack);
```

This would allow `logErrorWithConfig` to pull out the parts of state `baseErrorPayload` needs. `baseErrorPayload` could probably even go away at that point because its pretty straightforward when we have the parts extracted. Thoughts?

src/config.js
schwers: My thought was they might be useful in debugging at some point, but yeah they'll probably just lose value in the near future. removing.

src/lib/errorLog.js
schwers: Yup, this makes it easier to see the logs on the client and if you're ssh'd into a specific app server

schwers:
💇

src/app/reduxMiddleware/errorLogger.js
phil303: Ok fair point. Just an FYI, you could just do:
```js
const { env, userAgent, platform: { currentPage } } = store.getState();
```
But you're right, doing that over and over would be annoying. I think your combining `baseErrorPayload` with `logErrorWithConfig` is a nice solution though, since it seems like we don't need `baseErrorPayload` isolated yet.

phil303:
one last comment regarding `logErrorWithConfig`. Otherwise :fish:!

nramadas:
:tropical_fish: :tropical_drink:

PR #651, opened 2016-07-21
Error pages

phil303:
They look like this:

screen shot 2016-07-21 at 10 04 46 am


phil303:
:eyeglasses: @schwers or @nramadas

nramadas:
so, this takes care of bad urls within the normal client side flow, right? what happens if the user hits the node server with a bad url?

src/app/components/ErrorPage/index.jsx
```diff
@@ -0,0 +1,27 @@
...
+const GO_BACK = 'Go Back!';
+const GO_TO_FRONTPAGE = 'To the front page!';
+
+export default (props) => {
+  // NOTE: the second condition is a side-effect of a bug in our platform
+  // code, which is setting the url's referrer to itself. Ideally, only the
+  // first condition is needed.
+  const toFrontpage = props.referrer === '' || props.referrer === props.url;
+
+  const href = toFrontpage ? props.referrer : '/';
```
nramadas: shouldn't this be `toFrontpage ? '/' : props.referrer`?

phil303: yes! good find. I inverted the logic on the boolean (used to be called `goBack`) and I forgot to update this.

schwers:
Just as note status codes aren't mutually exclusive from the html ;)

Should this be renamed 404Page versus ErrorPage?

nramadas:
lgtm :fish:

phil303:
@schwers that thought had crossed my mind and originally I thought this might handle other page errors but I'm actually not sure what those would be since resources are fetched with ajax. I'd be fine with renaming.

phil303:
@nramadas @schwers regarding returning a 404. That's an excellent point that I completely forgot to address. Is there a way to return 404 status codes with platform?

nramadas:
Well, the other error pages would be when the server 500s, right?

phil303:
:haircut:

schwers:
Should `/handlers/Status404PageHandler.js` be `/handlers/Status404Page.js` with our new naming conventions?

phil303:
@schwers yes. Good catch.

schwers:
🐠  part two, the 🍣

PR #652, opened 2016-07-23
Add UTM data to semantic URLs. (CHAN-71)

wting:
Confirmed locally:
```shell
╭─@eikyo ~ ‹python-2.7.12›
╰─ curl -s -A "wting-${RANDOM}" http://localhost:4444/r/explainlikeimfive/comments/4ukacd/eli5_why_college_is_so_high_priced_in_the_us/ | python -c 'import sys; from BeautifulSoup import BeautifulSoup; print(BeautifulSoup(sys.stdin.read()).findAll("script", type="application/ld+json")[0].contents[0])' | jq '{url, sharedContent}'
{
  "url": "http://localhost:4444/r/explainlikeimfive/comments/4ukacd/eli5_why_college_is_so_high_priced_in_the_us/?utm_source=search&utm_medium=structured_data",
  "sharedContent": {
    "url": "http://localhost:4444/r/explainlikeimfive/comments/4ukacd/eli5_why_college_is_so_high_priced_in_the_us/?utm_source=search&utm_medium=structured_data"
  }
}
```

👓  @prashtx 

cc: @drunken-economist @schwers

### DIFF
--- a/src/views/components/GoogleCarouselMetadata.jsx
+++ b/src/views/components/GoogleCarouselMetadata.jsx
@@ -4,6 +4,31 @@ import has from 'lodash/object/has';
 
 import mobilify from '../../lib/mobilify';
 import propTypes from '../../propTypes';
+import url from 'url';
+
+export function isRedditDomain (hostname) {
+  return hostname.endsWith('reddit.com');
+}
+
+/*
+ * Adds UTM parameters to track inbound links from search engines.
+ */
+export function addUtmTracking (url_string) {
+  if (!url_string) {
+    return url_string;
+  }
+
+  const urlObject = url.parse(url_string, true);
+  if (!isRedditDomain(urlObject.hostname)) {
+    return url_string;
+  }
+
+  urlObject.query.utm_source = 'search';
+  urlObject.query.utm_medium = 'structured_data';
+  // node.url re-uses search string instead of query object if it exists
+  delete urlObject.search;
+  return url.format(urlObject);
+}
 
 function formatComments (comments, origin) {
   return comments.filter(c => c && c.author).map(function(c) {
@@ -47,7 +72,7 @@ function GoogleCarouselMetadata (props) {
   const baseObject = {
     '@context': 'http://schema.org',
     '@type': 'DiscussionForumPosting',
-    url: origin + props.url,
+    url: addUtmTracking(origin + props.url),
     headline: listing.title,
     datePublished: published.format(),
     author: {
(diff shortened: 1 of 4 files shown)


test/lib/extractErrorMsg.es6.js
```diff
@@ -9,26 +9,26 @@ const GENERIC_MESSAGE = 'There was a problem';
...
 
 describe('lib: extractErrorMsg', () => {
-  it('returns error messages nested in a compolete, 2-deep array', () => {
+  it('returns error messages nested in a complete, 2-deep array', () => {
     expect(extractErrorMsg({errors: [['error type', TEST_MESSAGE]]}))
       .to.equal(TEST_MESSAGE);
   });
-  
+
   it('returns a any message found in error.message', () => {
```
drunken-economist: typo, not a huge deal though 'returns a any message'

drunken-economist:
🐟

prashtx:
In cases where `sharedContent.url` points to a Reddit page, I think we should also add the UTM parameters. I think you can piggyback on the existing conditions in the handling of `sharedContent`.

💅

wting:
- cc'ed @schwers since @ajacksified is on PTO.
- added support for `sharedContent.url`
- added relevant tests

💇

wting:
- fix linter complaints

💇

src/views/components/GoogleCarouselMetadata.jsx
```diff
@@ -4,6 +4,27 @@ import has from 'lodash/object/has';
...
+export function isRedditDomain (hostname) {
+  return hostname.endsWith('reddit.com');
+}
+
+/*
+ * Adds UTM parameters to track inbound links from search engines.
+ */
+export function addUtmTracking (url_string) {
+  const urlObject = url.parse(url_string, true);
+  if (!url_string || !isRedditDomain(urlObject.hostname)) {
```
prashtx: Is it possible for `url_string` to be `null`/`undefined`, or are you guarding against empty string? `url.parse` will throw an exception if it receives null/undefined, so if those are possibilities, we should test before that call.

An empty string will lead to a null value for `urlObject.hostname`, so if that's plausible, then it's fair to test here.

prashtx:
1 small question about null/undefined/empty handling.

💅

src/views/components/GoogleCarouselMetadata.jsx
wting: Good point, the original intent was to guard against an empty string.

- `baseObject.url` will always be populated.
- `baseObject.sharedContent.url` will exist or be an empty string due to [`mobilify`](https://github.com/reddit/reddit-mobile/blob/8767a9747f4a5e03091b0f9729db1a298131a214/src/lib/mobilify.es6.js#L3).

It doesn't hurt to write defensive code, so I'll add null checking anyway.

wting:
- add `null` handling and related test

💇

prashtx:
slick.

🐟

wting:
Merged into 32fbd83d.

PR #654, opened 2016-07-26
Fix adblock tracking from throwing exceptions

schwers:
Fixes `/login` and `/register` pages from rendering as blank when you go there via an in-app navigation. 

The adblock tracking code is throwing an exception because the adblock test isn't rendered on that page, which breaks rendering. I modified the tracking to not test for adblock when we know the `div#adblock-test` won't be on the page, which would guarantee a false-positive. I changed `testAdblock` to not throw exceptions when the div isn't found.

👓  @dwick

### DIFF
--- a/assets/js/trackingEvents.es6.js
+++ b/assets/js/trackingEvents.es6.js
@@ -90,6 +90,8 @@ function trackingEvents(app) {
 
   function testAdblock() {
     const el = document.getElementById('adblock-test');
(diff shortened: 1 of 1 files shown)


schwers:
👓  @nramadas or @phil303

phil303:
:fish:

nramadas:
double :fish: :fish:

PR #656, opened 2016-07-27
Add Makefile.

wting:
👓 : @schwers 

This file is to formalize some of the build processes for working with this
repo, as well as providing a self-documenting entry point for new devs working
on this codebase.

Devs can hit `make <tab>` to see which targets they can build, while defaulting to `dev-server`.

Longer term, the goal is to standardize some of these targets cross repositories. This provides a consistent interface for devs while allowing each repo to customize what `make test` means.

Tested via `make clean && make`, and cloning to a new location and running `make`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,36 @@
+.PHONY: dev-server
+dev-server: dependencies dev-tools
+	nodemon -e jsx,js,less,css index.js
+
+.PHONY: dev-autobuild
+dev-autobuild: dependencies dev-tools
+	npm run watch
+
+.PHONY: dev-tools
+dev-tools:
+	@command -v mocha &>/dev/null || npm install -g mocha
+	@command -v nodemon &>/dev/null || npm install -g nodemon
+
+.PHONY: lint
+lint:
+	npm run lint
+
+.PHONY: test
+test: dev-tools
+	mocha
+
+.PHONY: clean
+clean:
+	rm -fr ./build ./node_modules
+
+.PHONY: dependencies
+dependencies: node_modules lib/snooboots/LICENSE build
+
+build:
+	npm run build
+
+lib/snooboots/LICENSE:
+	git submodule update --init
+
+node_modules:
+	npm install


phil303:
@wting While I think this is neat, what you're describing is the intention of `package.json`. This feels like a very thin wrapper, and as far as I can tell, everything you're describing can be done via that.

wting:
@phil303 True, this is just a thin wrapper of `package.json`. `Makefile` isn't supposed to encompass or replace that.

Some extra functionality it can provide is like automatically installing git hooks, or running linters as part of tests, or providing a `make clean` target to reset the repo.

The goal is to provide a self-documenting, discoverable interface that's consistent between codebases. This helps devs switching between repos. Off the top of your head: What's the build process for a Python 2 Pylons server? A Python 3 Pyramid server? Puppet? How about running tests for each codebase?

`Makefile` is the greatest common denominator and acts as a thin wrapper for language specific build tools.

Longer term, this makes containerization easier. By defining a consistent interface (`make clean && make test && make build`) it lets each codebase decide how they want to perform each step regardless of the language or tools.

phil303:
William and I chatted about this in person. At a high level, I'm not inherently against this as I was. The cross-platform compatibility for devs is very attractive and it will also answer the question of how to install git hooks for devs.

wting:
@phil303 and I discussed in person. Our biggest concern is now that we have two sources of truth between `Makefile` and `package.json`, and `Makefile` is non-standard for Node developers.

OTOH `Makefile` gives us the ability of forcing hooks to be installed as part of running tests so we can catch linting errors earlier instead of depending on developer's manual intervention.

We decided to give this a shot, and if it doesn't work we'll just revert the commit.

Makefile
```diff
@@ -0,0 +1,36 @@
+.PHONY: dev-server
+dev-server: dependencies dev-tools
+	nodemon -e jsx,js,less,css index.js
```
phil303: :no_bell: We should have these run their npm counterparts directly so there's no chance for code drift between the two.

spladug:
Note: if you're going to be "standardiz[ing] some of these targets cross repositories" that should probably not be done one off without input from other teams.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+.PHONY: dev-autobuild
+dev-autobuild: dependencies dev-tools
+	npm run watch
+
+.PHONY: dev-tools
+dev-tools:
+	@command -v mocha &>/dev/null || npm install -g mocha
```
schwers: I'm a little confused as to why these checks are necessary. It looks like `dev-tools` is always being run in conjunction with `dependencies`. npm's `package.json` already has a notion of `devDependencies` which includes `mocha`. Could we add `nodemon` to our `package.json`'s `devDependencies` section and rely on `dependencies` to run `npm install`? I also feel like a global install here is a bad idea.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+.PHONY: test
+test: dev-tools
+	mocha
+
+.PHONY: clean
+clean:
+	rm -fr ./build ./node_modules
+
+.PHONY: dependencies
+dependencies: node_modules lib/snooboots/LICENSE build
```
schwers: We have some git hooks in the `hooks/` directory, should we install them here?

schwers:
I agree with Neil on this, if we to standardize we should be getting input from other teams. Plus its worth noting that we already have infrastructure for running npm based projects (at least that is to say, we run mweb and mweb staging via npm commands as far as I know).  I left some questions on the specifics of the makefile.

My concerns are mostly higher level questionsabout the benefits of a makefile. I feel like a lot of the examples you gave can already be accomplished with npm scripts: 

- Consider that you can point npm scripts at a shell script that does something like `rm -rf ./build ./node_modules` to effectively make `npm run clean` by adding `"clean": ". ./my-clean-script.sh"`. But what's worth noting is npm is technically runnable on windows too, so something more universal would be to include the npm package `rimraf` and add `"clean": "rimraf ./build ./node_modules"` to `package.json`
- We could use npm script hooks like `preinstall` to install git hooks and update submodules
- We can combine scripts to do tests and linting together by adding `"lintAndTest": "npm run lint && npm run test"

To me it seems like these should be things we have in `package.json` already, and the makefile could be an even thinner wrapper. Could we talk through this some more after lunch?

wting:
@spladug, @schwers: Hey, thanks for the feedback! I should have ran this by y'all first before submitting the PR. Initially I was looking at this PR as testing the waters w/o a full commitment.

>To me it seems like these should be things we have in package.json already, and the makefile could be an even thinner wrapper. Could we talk through this some more after lunch?

@schwers: I sat down with @gtaylor today and it turns out a lot of this is made redundant as part of his work. I'd love to still grab lunch with you regardless.

Sorry about the delayed response. I dropped the ball due to an overly-aggressive gmail filter from auto-subscribing 60+ repos on joining reddit.

PR #657, opened 2016-07-27
Add a /health endpoint

uzi:
👓 @schwers

### DIFF
--- a/src/server/meta/index.js
+++ b/src/server/meta/index.js
@@ -54,6 +54,10 @@ export default (router, apiOptions) => {
     ctx.type = 'application/json';
   });
 
+  router.get('/health', (ctx) => {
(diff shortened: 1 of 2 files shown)


uzi:
💇

src/server/meta/index.js
```diff
@@ -54,6 +54,10 @@ export default (router, apiOptions) => {
     ctx.type = 'application/json';
   });
 
+  router.get('/health', (ctx) => {
```
schwers: 🔕 I'd say "you don't need parens here" but it looks like the rest of the file has it, good cleanup to do later

schwers:
:fish:

src/server/meta/index.js
uzi: 👍

PR #658, opened 2016-07-28
Use STATIC_BASE for building fonts, make that part of build step

schwers:
Depends on https://github.com/reddit/node-build/pull/20

Fixes font building in production. This change makes it so intermediate font files are ignored in the repo (sideffect: diffs will be cleaner when we add icons, yey). Fonts are now built when you run `npm run watch` and `npm run build` and respect the `STATIC_BASE` envirorment vairable

👓 @nramadas


nramadas:
:fish:

PR #659, opened 2016-07-29
remove svg compatibility check

phil303:
I'm not sure what mweb browers don't support SVG. This check was causing an error in the console which was a bit maddening. Removing it cleans up that warning and a bunch of code.

## Before
screen shot 2016-07-28 at 5 14 27 pm

## After
screen shot 2016-07-28 at 5 14 50 pm
Sweet relief.

### DIFF
--- a/src/app/components/SVG/index.jsx
+++ b/src/app/components/SVG/index.jsx
@@ -1,59 +1,39 @@
 import React from 'react';
 
+const T = React.PropTypes;
+
 const _NS = 'http://www.w3.org/2000/svg';
+const ICON_SIZE = 20;
 
-function SVG (props) {
-  const width = props.width || SVG.ICON_SIZE;
-  const height = props.height || SVG.ICON_SIZE;
-
-  if (SVG.ENABLED) {
-    const move = props.move;
-    const out = props.out;
-
-    return (
-      <svg
-        className={ props.className }
-        version='1.1'
-        xmlns={ _NS }
-        x='0px'
-        y='0px'
-        width={ `${width}px` }
-        height={ `${height}px` }
-        viewBox={ `0 0 ${width} ${height}` }
-        onMouseMove={ move }
-        onMouseLeave={ out }
-      >
-        { props.children }
-      </svg>
-    );
-  }
-
-  const fallbackIcon = props.fallbackIcon;
-  if (fallbackIcon) {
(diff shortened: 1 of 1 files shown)


src/app/components/SVG/index.jsx
```diff
@@ -1,59 +1,26 @@
...
+    </svg>
+  );
 }
 
-SVG.ENABLED = typeof document !== 'undefined' &&
-              (!!document.createElementNS && !!document.createElementNS(_NS, 'svg').createSVGRect);
-
-SVG.ICON_SIZE = 20;
-
 export default SVG;
```
nramadas: :no_bell: `export default function SVG(props) {`?

src/app/components/SVG/index.jsx
```diff
@@ -1,59 +1,26 @@
...
-  }
-
-  return <span className='placeholder'/>;
+const ICON_SIZE = 20;
+
+function SVG(props) {
+  const { move, out, width=ICON_SIZE, height=ICON_SIZE } = props;
+  return (
+    <svg
+      className={ props.className }
```
nramadas: we should just destructure `className` out alongside `move`, `out`, etc. same with `children` below.

nramadas:
:nail_care:

src/app/components/SVG/index.jsx
phil303: ugh, i hate it when you're right

phil303:
:haircut:

nramadas:
:fish:

src/app/components/SVG/index.jsx
```diff
@@ -1,59 +1,24 @@
...
-  return <span className='placeholder'/>;
+const ICON_SIZE = 20;
+
+export default function SVG(props) {
+  const { move, out, width=ICON_SIZE, height=ICON_SIZE } = props;
+  return (
+    <svg
+      className={ props.className }
+      version='1.1'
+      xmlns={ _NS }
```
schwers: 🔕  Fun fact, xmlns isn't actually supported. But it sounds like it should be fixed _soon_.  More info https://github.com/facebook/react/issues/1657

schwers:
:fish:

ajacksified:
android 4.2, which sees nothing if there isn't a fallback, and can't revert to desktop because they can't see the menu button.

PR #660, opened 2016-07-29
Login view error handling

uzi:

It's still limited in the errors that it handles, but more or less works similar to the 1X implementation. Still missing besides that are the signup page and the recovery option.

Oh, and I'm missing a test in the reducer... I'll go do that now.

### DIFF
--- a/src/app/components/Login/Input.jsx
+++ b/src/app/components/Login/Input.jsx
@@ -6,7 +6,7 @@ import cx from 'lib/classNames';
 const T = React.PropTypes;
 
 function LoginInput(props) {
-  const { showTopBorder, name, type, placeholder, error, children } = props;
+  const { showTopBorder, name, type, placeholder, error, onChange, value, children } = props;
   const inputClasses = cx('LoginInput__input', {
     'error': !!error,
     'show-top': showTopBorder,
@@ -18,11 +18,13 @@ function LoginInput(props) {
       <div className='LoginInput'>
         <label htmlFor={ name } className='hidden'/>
         <input
-          id={ name }
-          name={ name }
-          type={ type }
           className={ inputClasses }
+          id={ name }
+          onChange={ onChange }
+          name={ name }
           placeholder={ placeholder }
+          type={ type }
+          value={ value }
         />
         { children }
       </div>
@@ -32,16 +34,18 @@ function LoginInput(props) {
 }
 
 LoginInput.propTypes = {
+  error: T.string,
   name: T.string.isRequired,
-  type: T.string,
+  onChange: T.func,
   placeholder: T.string.isRequired,
   showTopBorder: T.bool,
-  error: T.string,
+  type: T.string,
+  value: T.string.isRequired,
 };
 
 LoginInput.defaultProps = {
-  type: 'text',
   showTopBorder: false,
+  type: 'text',
 };
 
 export default LoginInput;
--- a/src/app/components/Login/index.jsx
+++ b/src/app/components/Login/index.jsx
@@ -1,15 +1,88 @@
 import './styles.less';
 
 import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
 import { METHODS } from '@r/platform/router';
 import { Form } from '@r/platform/components';
+
+import * as sessionActions from 'app/actions/session';
 
 import SnooIcon from 'app/components/SnooIcon';
 import LoginInput from './Input';
 import SquareButton from './SquareButton';
 
-export default class Login extends React.Component {
+class Login extends React.Component {
+  constructor (props) {
+    super(props);
+
+    this.state = {
+      isPasswordField: true,
+      password: '',
+      username: '',
+    };
+
+    this.clearPassword = this.clearField.bind(this, 'password');
+    this.clearUsername = this.clearField.bind(this, 'username');
+    this.updatePassword = this.updateField.bind(this, 'password');
+    this.updateUsername = this.updateField.bind(this, 'username');
+    this.toggleEye = this.toggleEye.bind(this);
+  }
+
+  toggleEye() {
+    const { isPasswordField } = this.state;
+    this.setState({ isPasswordField: !isPasswordField });
+  }
+
+  clearField(fieldName, e) {
+    const { resetSessionError } = this.props;
+    e.preventDefault();
+    this.setState({ [fieldName]: '' }, resetSessionError);
+  }
+
+  updateField(fieldName, e) {
+    e.preventDefault();
+    this.setState({ [fieldName]: e.target.value });
+  }
+
+  renderClear(methodName) {
+    return (
+      <button
+        type='button'
+        className='Login__input-action-btn'
+        onClick={ this[methodName] }
+      >
+        <span className='icon icon-x red' />
+      </button>
+    );
+  }
+
+  renderEye() {
+    const { isPasswordField } = this.state;
+    const blue = isPasswordField ? '' : 'blue';
+
+    return (
+      <button
+        type='button'
+        className='Login__input-action-btn'
+        onClick={ this.toggleEye }
+      >
+        <span className={ `icon icon-eye ${blue}` } />
+      </button>
+    );
+  }
+
   render() {
+    const { session } = this.props;
+    const { isPasswordField, password, username } = this.state;
+    const passwordFieldType = isPasswordField ? 'password' : 'text';
+    const errorType = session ? session.error : null;
+
+    const error = { username: '', password: '' };
+    if (errorType) {
+      error.password = 'Sorry, that’s not the right password';
+    }
+
     return (
       <div className='Login'>
         <SnooIcon />
@@ -23,13 +96,31 @@ export default class Login extends React.Component {
             type='text'
             placeholder='Username'
             showTopBorder={ true }
-          />
+            error={ error.username }
+            onChange={ this.updateUsername }
+            value={ username }
+          >
+            {
+              error.username
+              ? this.renderClear('clearUsername')
+              : null
+            }
+          </LoginInput>
           <LoginInput
-              name='password'
-              type='password'
-              placeholder='Password'
-              showTopBorder={ false }
-          />
+            name='password'
+            type={ passwordFieldType }
+            placeholder='Password'
+            showTopBorder={ false }
+            error={ error.password }
+            onChange={ this.updatePassword }
+            value={ password }
+          >
+            {
+              error.password
+              ? this.renderClear('clearPassword')
+              : this.renderEye()
+            }
+          </LoginInput >
           <div className='Login__submit'>
             <SquareButton text='LOG IN' type='submit'/>
           </div>
@@ -38,3 +129,16 @@ export default class Login extends React.Component {
     );
   }
 }
+
+const mapStateToProps = createSelector(
+  state => state.session,
+  session => ({ session }),
+);
+
+const mapDispatchToProps = (dispatch) => ({
+  resetSessionError: () => {
+    dispatch(sessionActions.sessionError(null));
+  },
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(Login);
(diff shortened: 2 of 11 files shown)


src/app/components/Login/index.jsx
```diff
@@ -1,15 +1,78 @@
...
+    };
+
+    this.clearPassword = this.clearField.bind(this, '_password');
+    this.clearUsername = this.clearField.bind(this, '_username');
+    this.toggleEye = this.toggleEye.bind(this);
+  }
+
+  toggleEye() {
+    const { passwordFieldType } = this.state;
+    const newFieldType = passwordFieldType !== 'password' ? 'password' : 'text';
```
phil303: I think the logic becomes simpler to follow if you make this a boolean.
For instance this becomes
```
toggleEye() {
  this.setState({ isPasswordField: !this.state.isPasswordField });
}
```
And down below, your renderEye method becomes:
```
renderEye() {
  const blue = this.state.isPasswordField ? '' : 'blue';
  // etc
}

The only place that suffers is the `type` attr and I think that's ok for the extra clarity.

File: src/app/components/Login/index.jsx [hunk 2]
```diff
@@ -1,15 +1,78 @@
...
+        type='button'
+        className='Login__input-action-btn'
+        onClick={ this[methodName] }
+      >
+        <span className='icon icon-x red' />
+      </button>
+    );
+  }
+
+  renderEye(blue) {
```
phil303: There doesn't seem to be a need for this to be a parameter since this is dependent on `this.state`.

src/app/components/Login/index.jsx
```diff
@@ -38,3 +117,16 @@ export default class Login extends React.Component {
     );
   }
 }
+
+const mapStateToProps = createSelector(
+  state => state.session,
+  (session) => ({ session }),
```
phil303: nit, i think generally you shouldn't be wrapping single args in parens.

src/server/session/loginproxy.js
```diff
@@ -13,8 +13,14 @@ export default (router, apiOptions) => {
       writeSessionToResponse(ctx, data);
     } catch (e) {
       console.log(e);
-      console.log(e.stack);
-      ctx.throw(401, 'Incorrect username or password');
+      let errormsg = 'UNKNOWN_ERROR';
+
+      // A little unfortunate looking, but this is how the response looks
+      if (Array.isArray(e) && !!e[0] && Array.isArray(e[0]) && !!e[0][0]) {
```
phil303: yikes

Unfortunate is the right word. I think you can remove one condition here, which is the second one since `undefined`/`null` are not Arrays. So it can look like this:
```
(Array.isArray(e) && Array.isArray(e[0]) && !!e[0][0])
```

There might be a better way to do this altogether but I'm not seeing it.

src/server/session/loginproxy.js
```diff
@@ -13,8 +13,14 @@ export default (router, apiOptions) => {
...
+        errormsg = e[0][0];
+      }
+      // TODO: figure out how to get this to return json content type
+      ctx.throw(400, JSON.stringify({error: errormsg}));
```
phil303: should this be a 401?

src/app/reducers/session.js
```diff
@@ -8,6 +8,13 @@ export default function(state=DEFAULT, action={}) {
       const { session } = action.payload;
       return session;
     }
+    case sessionActions.SESSION_ERROR: {
+      // In the case of an error, we return an object with just that.
+      // These allows three states: 1) logged in, 2) logged out with no error
+      // and 3) logged out with an error (which can be cleared by sending null)
+      const { error } = action.payload;
+      return { error };
```
phil303: :no_bell: You can just one-line this:
```
return { error: action.payload };
```

src/app/components/Login/index.jsx
uzi: Sure -- I'll change that -- it was a carryover from the 1X implementation.

src/app/components/Login/index.jsx
uzi: Fair.

src/app/components/Login/index.jsx
uzi: Will change.

src/server/session/loginproxy.js
uzi: Cool, I'll verify that and simplify.

src/app/components/Login/index.jsx
```diff
@@ -23,13 +86,29 @@ export default class Login extends React.Component {
             type='text'
             placeholder='Username'
             showTopBorder={ true }
-          />
+            error={ error.username }
+            inputRef={ (c) => { this._username = c; } }
+          >
+            {
+              error.username
+              ? this.renderClear('clearUsername')
```
phil303: why not just pass the callback directly?

src/server/session/loginproxy.js
uzi: Ahh, yes -- good catch, I didn't notice that I changed that.

src/app/reducers/session.js
uzi: Yyyyeah... sure, why not.  Thanks!

src/app/components/Login/index.jsx
```diff
@@ -1,15 +1,78 @@
...
+        <span className={ `icon icon-eye ${blue}` } />
+      </button>
+    );
+  }
+
   render() {
+    const { session } = this.props;
+    const { passwordFieldType } = this.state;
+    const blue = passwordFieldType === 'text' ? 'blue' : '';
+    const errorType = (session) ? session.error : null;
```
phil303: why wrap session in parens?

src/app/components/Login/index.jsx
schwers: There doesn't need to be a parameter, but I might argue it's a little cleaner this way. I personally like being to look at a function and know what it needs without reading state, it feels more functional or top-down.

Whatever you end up deciding to do here: `blue` could technically be any string, it might be more readable to have it be a `boolean` parameter instead and this function could control the classname (alternatively maybe rename `blue` to `modifierClass`?)

src/app/components/Login/index.jsx
```diff
@@ -23,13 +86,29 @@ export default class Login extends React.Component {
             type='text'
             placeholder='Username'
             showTopBorder={ true }
-          />
+            error={ error.username }
+            inputRef={ (c) => { this._username = c; } }
```
schwers: small nit, you don't need the parens around c here

src/app/less/icon-font.less
```diff
@@ -18,6 +18,10 @@
     color: @blue;
   }
 
+  &.red {
+    color: @red;
```
schwers: Could you add a hover / touch state version of `red` as well? Further down in this file there's a section for that. Not that this particularly needs it but just a nice touch if you're in here.

src/app/components/Login/index.jsx
phil303: :no_bell: The way you're doing this feels "not very react-y" to me in that it's very imperative. And yes, it's probably fine for this one small component but I'd like to not set the precedent if possible. The react way to do this would be to maintain the input state (even inside the component is fine) and update the state with `onChange` events. When your user hits clear, you clear the fields in the state.

Please see these [docs](https://facebook.github.io/react/docs/forms.html) for an example of how to do this. :no_bell: because I don't want this to block you but I do feel semi-strongly that this is not a great way to do this.

src/app/components/Login/index.jsx
phil303: Normally I'd agree with you on principle but in practice I found this to be a very hard thing to keep up. Sometimes your sub-renders take so many damn arguments, you end up switching it over just to use state. I left that comment because then it provides consistency with that not-so-edgy edge case. But I'll downgrade it to :no_bell: because I can't whole-heartedly disagree with your point.

phil303:
:nail_care:

uzi:
💇

src/app/components/Login/index.jsx
```diff
@@ -1,15 +1,94 @@
...
+    newState[fieldName] = '';
+    this.setState(newState, resetSessionError);
+  }
+
+  updateField(fieldName, e) {
+    e.preventDefault();
+
+    let newState = {};
+    newState[fieldName] = e.target.value;
+    this.setState(newState);
```
schwers: I think this could be a little cleaner, you could write:
```javascript 
this.setState({ [fieldName]: e.target.value })
```
As a side note, because newState is never reassigned you can use `const` here. Mutating a property is technically not considered to be mutating newState as far as let and const are concerned

src/app/components/Login/index.jsx
```diff
@@ -1,15 +1,94 @@
...
+    this.setState({ isPasswordField: !isPasswordField });
+  }
+
+  clearField(fieldName, e) {
+    const { resetSessionError } = this.props;
+    e.preventDefault();
+
+    let newState = {};
+    newState[fieldName] = '';
+    this.setState(newState, resetSessionError);
```
schwers: Same comment as below, re style. Seems like this could be
```javascript
this.setState({ [fieldName]: '' });
```

src/app/components/Login/styles.less
```diff
@@ -33,4 +33,16 @@
...
   }
+
+  &__input-action-btn {
+    width: 40px;
+    height: 100%;
+    position: absolute;
+    right: 0;
+    font-size: 1.8em;
+
+    [class^="icon-"] {
```
schwers: 🔕 You can write `.icon {` here, the new icon font css is a little easier to work with.

src/app/components/Login/Input.jsx
```diff
@@ -32,16 +34,18 @@ function LoginInput(props) {
 }
 
 LoginInput.propTypes = {
+  error: T.string,
```
schwers: 🔕  Alphabetization? 💯

src/app/reducers/session.js
schwers: I second the one liner in this case :)

src/app/components/Login/index.jsx
```diff
@@ -1,15 +1,94 @@
...
+  }
+
+  renderEye() {
+    const { isPasswordField } = this.state;
+    const blue = isPasswordField ? '' : 'blue';
+
+    return (
+      <button
+        type='button'
+        className='Login__input-action-btn'
```
schwers: Really tiny nit, can you make this button have no outline on focus or active? There's a little orange box that shows up around it.

schwers:
Just some small style things, looks good otherwise 💅

src/app/components/Login/index.jsx
schwers: might me
```less
 &:focus {
    outline: none;
 }
```

src/app/reducers/session.js
uzi: Talked with @schwers and keeping as is... I'd have to do `return { error: action.payload.error };` and that seems less clean.  I could also just return action.payload, but it felt safer to make a new object.  That's why it's staying as is.

src/app/components/Login/Input.jsx
uzi: Oh, uhh, yeah... there was a eslint rule we had at Planet that would squawk about that.  I don't know if it's a practice we want in general, but I went with it here.

src/app/components/Login/styles.less
uzi: Fixing. Thanks!

src/app/components/Login/index.jsx
uzi: GAH!  I keep forgetting about this trick.  Yep, I'll go this route.

src/app/components/Login/index.jsx
uzi: Thanks, trying it.

uzi:
💇

src/app/components/Login/Input.jsx
schwers: Cool, I'm a fan of it for propTypes and imports

schwers:
🌊  🐚  🐟  🐠  🐳  🐋  🐬  🐙  🍣  🍥

PR #662, opened 2016-07-29
Update readme.md

schwers:


phil303:
@schwers can we kill this?

PR #664, opened 2016-08-03
Expose canonical URLs

prashtx:
Fix an issue preventing the existing canonical-url implementation from kicking in.

Set actual canonical URLs for paths where the mobile-web and desktop URLs don't match.

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -131,6 +131,8 @@ export function buildProps(ctx, app) {
     showEUCookieMessage: ctx.showEUCookieMessage,
     showGlobalMessage: ctx.showGlobalMessage,
     country: ctx.country,
+    // Default canonical path is the same as the m.reddit.com path
+    canonicalPath: ctx.url,
   };
 
   ctx.props.apiOptions = buildAPIOptions(ctx);
@@ -354,6 +356,12 @@ function routes(app) {
     const sort = this.query.sort || SORTS.HOT;
     const time = listingTime(this.query, sort);
 
+    if (this.url.endsWith('/')) {
+      props.canonicalPath = this.url;
+    } else {
+      props.canonicalPath = `${this.url}/`;
+    }
+
     this.preServerRender = function indexPagePreRender() {
       // If we're on a next/prev for an invalid thing_id, so that no results are
       // sent back, we have a stale cache. Redirect back to the first page of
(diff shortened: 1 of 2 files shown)


prashtx:
👓 @schwers

src/routes.jsx
```diff
@@ -473,7 +475,19 @@ function routes(app) {
...
-    props.data.set('listing', app.api.links.get(listingOpts));
+    const listingPromise = app.api.links.get(listingOpts);
+    props.data.set('listing', listingPromise);
+
+    // For the canonical path, we need the subreddit, listing ID, and listing title.
+    if (ctx.params.subreddit && ctx.params.listingTitle && ctx.params.listingId) {
+      const { subreddit, listingTitle, listingId } = ctx.params;
+      props.canonicalPath = `/r/${subreddit}/comments/${listingId}/${listingTitle}/`;
+    } else {
+      listingPromise.then(listing => {
```
schwers: Should this be added to `props.data`? `horse-react/server` will call `Promise.all` on `props.data`[[1]](https://github.com/reddit/node-horse-react/blob/master/src/server.es6.js#L63) before rendering. I'm worried about a race-condition when `listingPromise` resolves last for some reason and `cannonicalPath` isn't set before rendering.

src/routes.jsx
```diff
@@ -752,10 +773,12 @@ function routes(app) {
   }
 
   router.get('saved', '/u/:user/saved', function * () {
+    this.props.canonicalPath = `/user/${this.params.user}/saved/`;
```
schwers: Do we know if crawlers visit these pages with some authentication? Saved and Hidden seem like pages we don't want crawled in general.

schwers:
💅  Patch looks good overall, I just have a minor question and an implementation note.

src/routes.jsx
prashtx: We actually block these via `robots.txt`, but it bugged me that we had canonical tags that were not actually canonical...

prashtx:
💇

src/routes.jsx
schwers: Ah I see, okay.

schwers:
:fish:

prashtx:
@schwers: caught an issue with lack of trailing `/` in listings canonicals.

💇

schwers:
:fish:

PR #665, opened 2016-08-03
Sort and Time selectors. Doesn't include comment tree updates.

schwers:
(edit: widgets patch merged, no longer dependent)

This patch adds sorting and time filtering selectors for lists of posts and comments. Comments don't completely work, but are included to get an MVP of this feature into 2X. Comments will need some api work, because the comment tree is stored ad-hoc in each individual comments' `.replies` field. e.g. Go to a comments page sorted by best (a), change sort to top (b), and then back to best (c). You may notice on page (c) that some comment replies aren't ordered the same as page (a) even though they have the same exact sort.

### DIFF
--- a/src/app/components/Dropdown/styles.less
+++ b/src/app/components/Dropdown/styles.less
@@ -3,10 +3,16 @@
 
 .Dropdown {
   width: 600px;
-
+  
   .themeify({
     border: @theme-border;
   });
+
+  &__arrow {
+    .themeify({
+      border-bottom-color: @theme-body-color !important; // tool tip uses inline-styles
+    });
+  }
 }
 
 .DropdownRow,
@@ -36,18 +42,44 @@
     width: 50px;
     text-align: center;
 
+    &.m-selected {
+      color: @blue;
+    }
+
+    &.icon-check-circled.m-selected {
+      color: @lime;
+    }
+
     &.icon-link {
       font-size: 36px;
     }
 
-    &.icon-snoosilhouette,
-    &.icon-user-account,
-    &.icon-save,
-    &.icon-hide,
-    &.icon-flag,
+    &.icon-check-circled,
     &.icon-delete_remove,
-    &.icon-posts {
+    &.icon-flag,
+    &.icon-hide,
+    &.icon-history,
+    &.icon-new,
+    &.icon-posts,
+    &.icon-save,
+    &.icon-snoosilhouette,
+    &.icon-user-account {
       font-size: 22px;
+    }
+
+    &.icon-bar-chart,
+    &.icon-hot,
+    &.icon-op {
+      font-size: 21px;
+    }
+
+    &.icon-karma {
+      font-size: 28px;
+    }
+
+    &.icon-circle {
+      font-size: 20px;
+      color: @pale-grey;
     }
 
     &.icon-save {
--- a/src/app/components/PostAndCommentList/index.jsx
+++ b/src/app/components/PostAndCommentList/index.jsx
@@ -56,7 +56,7 @@ const PostsCommentsAndPagination = props => {
   return (
     <div>
       { records.map(renderRecord) }
-      { records.length && <PaginationButtons records={ records } /> }
+      { records.length > 0 && <PaginationButtons records={ records } /> }
     </div>
   );
 };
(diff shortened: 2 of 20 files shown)


schwers:
👓  @nramadas

src/app/components/Dropdown/styles.less
```diff
@@ -2,11 +2,19 @@
 @import (reference) '~app/less/themes/themeify';
 
 .Dropdown {
-  width: 600px;
+  width: 300px; // fallback incase this device doesn't support calc
+  width: calc(~'100%' - 2 * @grid-size); // effectively `margin: 0 @grid-size`
```
nramadas: :no_bell: i'm surprised you need this. `<Tooltip/>`s are supposed to automatically margin themselves an amount equal to the buffer they get as props. If this isn't working, i'll take a minute to try to fix it.

src/app/components/Dropdown/styles.less
```diff
@@ -36,20 +44,46 @@
...
+    &.icon-save,
+    &.icon-snoosilhouette,
+    &.icon-user-account {
       font-size: 22px;
     }
 
+    &.icon-bar-chart,
+    &.icon-hot,
+    &.icon-op {
+      font-size: 21px;
```
nramadas: :no_bell: one day, our icons will get normalized. one day...

src/app/components/PostAndCommentList/index.jsx
```diff
@@ -56,7 +56,7 @@ const PostsCommentsAndPagination = props => {
   return (
     <div>
       { records.map(renderRecord) }
-      { records.length && <PaginationButtons records={ records } /> }
+      { records.length > 0 && <PaginationButtons records={ records } /> }
```
nramadas: :no_bell: what was the bug here?

src/app/components/Dropdown/styles.less
schwers: Interesting, fwiw, I didn't try removing width completely from this, seeing `width: 600px` made me assume it was needed.

src/app/components/PostAndCommentList/index.jsx
schwers: The `&&`will short-circuit the first value if its false. in this case it was 0 which react interprets as the string '0'.

src/app/components/Dropdown/styles.less
schwers: Okay, so I tried removing the `width:...` and the tooltip doesn't really render properly, assuming we want to still do the full-width style. This is because the content of the dropdown isn't that wide. With the width set to 600px like before, `<Tooltip />` uses `getPositions` to try and come up with a rough alignment and then `positionTooltip` does a rough pass at verification afterwards when it sets style properties on the tool tip element. In this case the dropdown is first being centered by `getPositions`, but `positionTooltip` ensures the left is greater than the given offset. In this case the centered left is about -300, so its left is updated to 8 but the check to put the right side at offset (if the right side is past offset) doesn't use the updated `left` property [[tooltip.js]](https://github.com/reddit/node-widgets/blob/master/lib/modules/tooltip.js#L131). Even if `right` was properly set by `<Tooltip />` the hardcoded `width` will force the content to go past `right: 8px` (I tried making the content have a larger width as well). So I think we need to keep the width here, unless you see another way to size the content appropriately assuming the dropdown's `right` is set correctly.

src/app/components/PostAndCommentList/index.jsx
nramadas: heh, that's kinda funny actually

src/app/components/Dropdown/styles.less
schwers: From IRL, I missed that tooltip tries to override width to 'auto' as well. By fixing the issue with tooltip setting `right` this works fine with the old `width: 600px`.

src/app/router/handlers/PostsFromSubreddit.js
```diff
@@ -19,7 +20,7 @@ export default class PostsFromSubreddit extends BaseHandler {
       multi,
       multiUser,
       sort,
-      t,
+      t: listingTime(queryParams, sort),
```
nramadas: could we name this `time` instead of `t`?

nramadas:
:fish:

PR #666, opened 2016-08-04
Prevent client from re-trying failed server-side requests

schwers:
I tested this locally by simulating an timeout in loading `/hot.json` from the frontpage. You can try this by linking `node-api-client` and then in `node-api-client/old-src/endpoints/v1.es6.js` on line 160 in `returnGETPromise` [[src]](https://github.com/reddit/node-api-client/blob/master/old-src/endpoints/v1.es6.js#L160) add the following:

```javascript
if (options.uri === 'https://www.reddit.com/hot.json') {
  err = {
    status: 503,
    message: 'fake error for testing retrying',
  };
}
```

This simulates a 503 of the listing endpoint hitting `hot.json` Note: this assumes you're logged out, you can edit the `options.uri` comparison

👓  @nramadas

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -1,4 +1,5 @@
 import 'babel-polyfill';
+import React from 'react-dom';
 
 import { sendTimings, onError, loadShimThenInitialize, onLoad, render } from './clientLib';
 
@@ -19,7 +20,7 @@ const App = mixin(ClientReactApp);
 import config from '../../src/config';
 import cookies from 'cookies-js';
 import setLoggedOutCookies from '../../src/lib/loid';
-import routes from '../../src/routes';
+import routes, { buildProps } from '../../src/routes';
 
 import trackingEvents from './trackingEvents';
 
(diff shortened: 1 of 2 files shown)


assets/js/client.es6.js
```diff
@@ -128,7 +129,41 @@ function initialize(bindLinks) {
     setEvents(app, hasHistAndBindLinks, render, $body);
 
     sendTimings(beginRender, props.adsEnabled);
-  });
+  };
+
+  const serverStatus = bootstrap.ctx ? bootstrap.ctx.status : 500;
+
+  if (Math.floor(serverStatus / 100) === 5) {
```
nramadas: would it be simpler to just say `((serverStatus >= 500) && (serverStatus < 600))`?

src/app-mixin.jsx
```diff
@@ -127,15 +132,15 @@ function mixin (App) {
...
-      const title = errorMsgMap[statusCode] || errorMsgMap.default;
+      const title = ERROR_MSG_MAP[statusCode] || ERROR_MSG_MAP.default;
 
       if (!isNaN(parseInt(statusCode))) {
         ctx.status = parseInt(statusCode);
       } else {
         ctx.status = 503;
       }
 
-      Object.assign({}, ctx.props || {}, {
```
nramadas: wait, what was this even doing before?

nramadas:
:nail_care:

src/app-mixin.jsx
schwers: Exactly.

schwers:
💇

nramadas:
:fish:

PR #667, opened 2016-08-06
Use true black for night theme

ajacksified:
Use a real black for the background color on most pages. It's better for battery life, is _actually_ darker in dark situations, like at night time, and I think it still looks good. Should probably be discussed with designers, but here's a PR just in case. Alternatively, we could detect android/ios and provide android only with the black, and gray everywhere else, but I don't feel strongly about keeping the gray vOv

Here's a couple screenshots, black first and then the current gray:

screen shot 2016-08-06 at 12 38 42 pm

screen shot 2016-08-06 at 12 40 00 pm

screen shot 2016-08-06 at 12 38 56 pm

screen shot 2016-08-06 at 12 39 54 pm

tldr: it may be very slightly less aesthetically pleasing, but it's more actually useful, and it saves battery on amoled screens.

### DIFF
--- a/src/app/components/cssCommons/PostAndCommentList.less
+++ b/src/app/components/cssCommons/PostAndCommentList.less
@@ -3,6 +3,6 @@
 
 .PostAndCommentList {
   .themeify({
-    background-color: @theme-line-color;
+    background-color: @theme-body-color;
   });
 }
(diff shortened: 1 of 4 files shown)


ajacksified:
👓 @nramadas and @schwers ?

schwers:
I'm a fan of this approach, I feel like true blacks on iOS are only a matter of time. I can follow up with the designers to see what they think. Thanks for making this!

schwers:
talked with Derick on this, he's a-ok with it. We should land this, I'll find out timeline

ajacksified:
Rebased, still works.

schwers:
:fish:

PR #668, opened 2016-08-09
Prevent client retrying 4XX statuses too

schwers:
From the logs, we seem to get a surprising number of 401s and 404s. In both cases we're autoretrying on the client which defeats the purpose of https://github.com/reddit/reddit-mobile/pull/666, and in the 404 case we often will get the same 404 response back on the client anyway. 

👓  @nramadas


nramadas:
:fish:

PR #671, opened 2016-08-10
Use a better atob/btoa. Fix malformed session handling

schwers:
We installed the 'atob' package to avoid sending the node Buffer
module to the client. But we weren't actually using it. Worse still,
something is generating malformed session cookies, potentially atob.
To remedy this we install a new, better, smaller, and more tested
base64 enconding libarary that is compatible with atob/btoa.

When you have a malformed token cookie, the server was broken. This
adds better error handling to prevent that from happening.

Lastly our stack traces weren't always including helpful debug info.
This should fix that.


schwers:
👓  @nramadas  I'm happy to port the new btoa/atob replacement to node-private and make it a peer dep if this looks good. Just to reduce payload size a little bit.

nramadas:
so, i'm okay with changing libraries, but I would like to see some way of verifying that this actually solves the issue. at least, before we switch node-private (and modmail) over to it.

schwers:
So after further review of our possible atob/btoa patches, I think the one we had might actually be the best. It's the smallest and has been updated most recently (~1 week). 

In short, most of what this patch does now is add better error handling for malformed cookies and improves some of our error parsing.

💇

src/lib/errorLog.js
```diff
@@ -43,14 +43,16 @@ export default function (details, errorEndpoints, config={}) {
...
-      url: parts.slice(0, len - 2).join(':'),
-      line: parts[len - 2],
-      column: parts[len - 1],
-      message: `Error: ${error.message}`,
-    };
+    const parts = sourceInfoFromStack(error.stack);
+    if (parts) {
+      const len = parts.length;
+      return {
+        url: parts.slice(0, len - 2).join(':'),
```
nramadas: it would probably be cleaner to have `sourceInfoFromStack` build most of this dictionary directly. something like:

```es6
const parsedError = parseErrorStack(stack);
if (parsedError) {
   return { ...parsedError, message: `Error: ${error.message}` };
}
```

schwers:
Closing this as the cookie work and error tracking is getting split up into two separate PRs

PR #672, opened 2016-08-11
Add register page

jamesmaa:
👓 @schwers @nramadas 

What should I do about QA? I hand tested it manually and it seems to work

Here's the sister PR: https://github.com/reddit/node-private/pull/1


jamesmaa:
General look of the page: 
image

src/app/components/Login/index.jsx
```diff
@@ -86,6 +87,13 @@ class Login extends React.Component {
     return (
       <div className='Login'>
         <SnooIcon />
+        <div className='Login__register-link'>
+          <p>
+            <Anchor href='/register'>
+              { 'New user? Sign up!' }
```
nramadas: nit, but i don't think you need the curly brackets.
```es6
<Anchor href='/register'>New user? Sign up!</Anchor>
```

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+import SnooIcon from 'app/components/SnooIcon';
+import SquareButton from 'app/components/LoginRegistrationForm/SquareButton';
+import LoginInput from 'app/components/LoginRegistrationForm/Input';
+import 'app/components/LoginRegistrationForm/styles.less';
+
+const DEFAULT_ERROR = () => {
+  return (
+    <span>
+      { 'Sorry, something went wrong. Please try again later or report the bug to ' }
```
nramadas: I don't think you need the curlies here either

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+  'USERNAME_TAKEN_DEL': 'Sorry the account associated with that username is deleted',
+};
+
+const TERMS = (
+  <a
+    href='/help/useragreement'
+    className='text-link'
+    target='_blank'
+  >
+    'Terms '
```
nramadas: nit, but the `'`s are unnecessary

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+      { 'Sorry, something went wrong. Please try again later or report the bug to ' }
+      <Anchor href='/r/bugs' className='register__bug-link'>
+        { '/r/bugs' }
+      </Anchor>
+    </span>
+  );
+};
+
+const EMAIL_ERRORS = {
+  'BAD_EMAIL': 'Sorry, the email you entered is invalid',
```
nramadas: you can leave out the quotes in the key:
```es6
const EMAIL_ERRORS = {
  BAD_EMAIL: 'Sorry, the email you entered is invalid',
  // etc
};
```

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+  <a
+    href='/help/contentpolicy/'
+    className='text-link'
+    target='_blank'
+  >
+    'Content Policy'
+  </a>
+);
+
+const renderTerms = () => {
```
nramadas: :no_bell: you could make this a constant like you did with PRIVACY, CONTENT, etc

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+        error.password = PASS_ERRORS[errorType];
+      } else if (errorType in USER_ERRORS) {
+        error.username = USER_ERRORS[errorType];
+      } else {
+        error.default = DEFAULT_ERROR();
+      }
+    }
+
+    return (
+    <div className='Register'>
```
nramadas: nit: indentation

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+        </Form>
+        { renderTerms() }
+    </div>
+    );
+  }
+}
+
+const mapStateToProps = createSelector(
+  state => state.session,
+  session => ({ session }),
```
nramadas: it seems like `session` is only really important for the errors that it will pass through eventually. instead having the component take in a session object, you can have it take in errors instead. something like this:

```es6
  state => state.session,
  session => {
    const error = { username: '', password: '', email: '', default: ''};
    const errorType = session ? state.session.error : null;

    if (errorType in EMAIL_ERRORS) {
      error.email = EMAIL_ERRORS[errorType];
    } else if (errorType in PASS_ERRORS) {
      error.password = PASS_ERRORS[errorType];
    } else if (errorType in USER_ERRORS) {
      error.username = USER_ERRORS[errorType];
    } else {
      error.default = DEFAULT_ERROR();
    }

    return { error };
  }
);
```

Now, your component doesn't need to worry about session at all. instead, it just takes in an error object.

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+);
+
+const mapDispatchToProps = (dispatch) => ({
+  resetSessionError: () => {
```
nramadas: if we make the change above and don't pass session into the component, we don't need the component to understand "resetting" a session. instead, the callback could look like this:
```es6
const mapDispatchToProps = dispatch => ({
  resetErrors: () => dispatch(sessionActions.sessionError(null)),
});
```

src/server/session/registerproxy.js
```diff
@@ -0,0 +1,27 @@
...
+
+    try {
+      let newsletter_subscribe = newsletter.length === 1 ? true : false;
+      const data = await PrivateAPI.register(apiOptions, username, password, email, newsletter_subscribe);
+      // writeSessionToResponse will set the cookies
+      writeSessionToResponse(ctx, data);
+    } catch (e) {
+      let errormsg = 'UNKNOWN_ERROR';
+
+      // A little unfortunate looking, but this is how the response looks
```
nramadas: well, we have control over this, don't we? couldn't we make `PrivateAPI` return errors in a different format?

src/server/session/registerproxy.js
```diff
@@ -0,0 +1,27 @@
+import writeSessionToResponse from './writeSessionToResponse';
+import { PrivateAPI } from '@r/private';
+
+export default (router, apiOptions) => {
+  router.post('/registerproxy', async (ctx/*, next*/) => {
+    const { username, password, email, newsletter} = ctx.request.body;
+
+    try {
+      let newsletter_subscribe = newsletter.length === 1 ? true : false;
```
nramadas: a few nits:
0. `const` instead of `let`.
0. camelCase instead of snake_case.
0. the ternary is unnecessary 

```es6
const newsletterSubscribe = !!newsletter.length;
```

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,237 @@
...
+          <LoginInput
+            name='username'
+            type='text'
+            placeholder='Choose a unique username'
+            showTopBorder={ true }
+            error={ error.username }
+            onChange={ this.updateUsername }
+            value={ username }
+          >
+            {
```
nramadas: small nit, ternarys like this look a bit better written out when the clauses have an extra indent:

```es6
{ 
  error.username
    ? this.renderClear('clearUsername')
    : null
}
```

or, for a bit more compactness
```es6
{ error.username
  ? this.renderClear('clearUsername')
  : null }
```

nramadas:
:nail_care:

jamesmaa:
Really appreciate the thorough review! 
Update:
- Responded to comments
- Moved error parsing to `node-private`
💇

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,236 @@
...
+  <a
+    href='/help/contentpolicy/'
+    className='text-link'
+    target='_blank'
+  >
+    'Content Policy'
+  </a>
+);
+
+const RENDER_TERMS = () => {
```
nramadas: I think i may have worded my previous comment incorrectly :)

What i meant was, instead of this being a function, it could just be a value like the other constants.
```es6
const RENDER_TERMS = (
  <div className='Register__terms'>
    By signing up, you agree to our { TERMS }
    and that you have read our { PRIVACY }
    and { CONTENT }.
  </div>
);
```

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,236 @@
...
+          >
+            <input 
+              type='checkbox'
+              name='newsletter'
+              defaultChecked
+            />
+            Subscribe to newsletter
+          </label>
+          {
+              error.default
```
nramadas: i think the whole ternary is indented one extra

src/app/components/Register/Index.jsx
```diff
@@ -0,0 +1,236 @@
...
+
+    if (errorType in EMAIL_ERRORS) {
+      error.email = EMAIL_ERRORS[errorType];
+    } else if (errorType in PASS_ERRORS) {
+      error.password = PASS_ERRORS[errorType];
+    } else if (errorType in USER_ERRORS) {
+      error.username = USER_ERRORS[errorType];
+    } else if (errorType) {
+      // Not a token type error message
+      error.default = renderErrorMsg(errorType);
```
nramadas: it feels a bit weird have the redux selector return react components when mapping state to props. it would be better for it to set a string and have the template convert the string into components.

@phil303 what are your thoughts?

src/app/models/Register.js
```diff
@@ -0,0 +1,13 @@
+
+import superagent from 'superagent';
+
+export const registerUser = (username, password, email, newsletter) =>
+  new Promise((resolve, reject) => {
+    superagent
+      .post('/registerproxy')
+      .send({ username, password, email, newsletter})
```
nramadas: nit: missing a space after newsletter

src/app/components/Register/Index.jsx
phil303: yea i agree. It's more verbose and almost philosophical defining these borders but operating on the principle of least surprise, this feels like a better approach. It keeps all the html output in the render function.

phil303: Additional note, you can just make the default the `errorType` string, then call `renderErrorMsg` above if the string is non-empty.

jamesmaa:
💇 Already done :)

nramadas:
lgtm. if you can make travis happy, then i'll fish :nail_care:

jamesmaa:
How do I make travis happy? 😢 
`Module not found: Error: Can't resolve 'app/components/Register' in '/home/travis/build/reddit/reddit-mobile/src/app/pages'`

nramadas:
:fish: (let's debug travis later)

PR #673, opened 2016-08-11
Handle comment threads that are too deep

uzi:
When a comment thread goes deeper than the maximum
levels to render a tree, it was displaying
"More comments (0)" and the action was failing.

This makes it so the behavior is more in line with
the desktop site in which it will load the comment
tree from the given point after clicking a
"Continue this thread" message.

This is how desktop handles it:
screen shot 2016-08-10 at 5 44 40 pm

This is how it's currently handled:
screen shot 2016-08-10 at 5 44 22 pm

This is if you click on it right now:
screen shot 2016-08-10 at 5 43 46 pm

This is after the change:
screen shot 2016-08-10 at 5 45 01 pm

👓 @schwers @phil303

### DIFF
--- a/assets/less/components/comment/CommentContinueThisThread.less
+++ b/assets/less/components/comment/CommentContinueThisThread.less
@@ -1,0 +1,26 @@
+.CommentContinueThisThread {
+  &__caron,
+  &__msg,
+  &__dots {
+    display: inline-block;
+    vertical-align: middle;
+  }
+  
+  &__caron {
+    margin-right: @grid-size;
+    transform: rotate(-90deg);
+    
+    &[class*=" icon-"] {
+      color: @blue;
+    }
+  }
+  
+  &__msg {
+    color: @blue;
+  }
+  
+  &__dots {
+    margin-right: @half-grid-size;
+    color: @blue;
+  }
+}
--- a/src/views/components/comment/Comment.jsx
+++ b/src/views/components/comment/Comment.jsx
@@ -11,6 +11,7 @@ import CommentHeader from './CommentHeader';
 import CommentTools from './CommentTools';
 import CommentReplyForm from './CommentReplyForm';
 import CommentSeeMore from './CommentSeeMore';
+import CommentContinueThisThread from './CommentContinueThisThread';
 import CommentEditForm from './CommentEditForm';
 
 const T = React.PropTypes;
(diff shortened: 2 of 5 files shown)


uzi:
Oh, and the thread I used to test -- https://m.reddit.com/r/Showerthoughts/comments/4x37o2/marvel_movies_make_me_care_about_characters_i/

src/views/components/comment/Comment.jsx
```diff
@@ -643,13 +644,28 @@ export default class Comment extends BaseComponent {
       );
     }
 
+    const dots = Math.max(nestingLevel - 6, 0);
+
+    if (reply.count === 0) {
+      const { comment, permalinkBase } = this.props;
+      const dest = permalinkBase + comment.id;
```
schwers: I have a minor preference for a template string here.
```javascript
const dest = `${permalinkBase}${comment.id}`;
```
It just makes it a little more obvious that the result is a string

phil303: I wonder if this should be in our style guide and [linter](http://eslint.org/docs/rules/prefer-template). I personally don't feel too strongly either way but i've seen this kind of comment more than once since I've been here, which to me is strong indicator that we should automate it.

src/views/components/comment/CommentContinueThisThread.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import fill from 'lodash/array/fill';
+
+const T = React.PropTypes;
+
+function getLoadingMsg() {
+  return 'Continue this thread';
```
phil303: I think this should just be a constant.

src/views/components/comment/CommentContinueThisThread.jsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+function renderDots(count) {
+  const content = fill(Array(count), '•').join(' ');
```
phil303: neat trick

schwers:
Could you add the `Comment__continuethisthread` class to line 49/50 in `assets/less/components/comment/Comment.less`? It would keep the spacing, and borders consistent with the 'More comment' divs. It looks like this if you add it:

image

assets/less/components/comment/CommentContinueThisThread.less
```diff
@@ -0,0 +1,26 @@
+.CommentContinueThisThread {
```
schwers: Super minor, it seems like me might want to share these styles with `CommentSeeMore.less`, but I don't feel strongly about it.

phil303:
:fish: code looks good to me but I'll let Ryan have the final fish.

schwers:
Just a couple of minor nits, overall the patch looks good and this is a great improvement. Want to add it to 2X as well? ;) 💅

src/views/components/comment/Comment.jsx
uzi: Fair... though I figure it's something you're generally told once... maybe twice... then it sticks. Fixing.

src/views/components/comment/CommentContinueThisThread.jsx
uzi: Can't take credit -- copied from another file.

src/views/components/comment/CommentContinueThisThread.jsx
uzi: Sure, no problem.

uzi:
💇

schwers:
:fish: 🐠  🍹  🍣

PR #674, opened 2016-08-11
Fix hiding ads based on user preferences

uzi:
Turns out that data.preferences is the place to look.

It also doesn't appear that props.prefs is used, so removing that.

👓 @schwers @phil303

### DIFF
--- a/src/views/pages/index.jsx
+++ b/src/views/pages/index.jsx
@@ -28,9 +28,6 @@ class IndexPage extends BasePage {
     multi: React.PropTypes.string,
     multiUser: React.PropTypes.string,
     page: React.PropTypes.number,
-    prefs: React.PropTypes.shape({
-      hide_ads: React.PropTypes.bool.isRequired,
-    }),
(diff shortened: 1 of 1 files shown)


src/views/pages/index.jsx
```diff
@@ -162,8 +159,7 @@ class IndexPage extends BasePage {
     }
 
     let showAds = props.adsEnabled && !!props.config.adsPath;
-
-    if (props.prefs && props.prefs.hide_ads === true) {
+    if (data.preferences && data.preferences.hide_ads === true) {
```
phil303: you don't need the `=== true`.

phil303:
:nail_care:

src/views/pages/index.jsx
uzi: I blame whoever git blame blames. Sure ... I'll change it.

uzi:
💇

schwers:
:fish:

PR #675, opened 2016-08-11
Add GoogleCarouselMetadata to comments 

schwers:
👓  @uzi or @nramadas 

For verification of channels related stuff:

👓  @prashtx or @wting 

I made some changes related to the structured data fields, see https://reddit.atlassian.net/browse/CHAN-124 for explanation

### DIFF
--- a/src/app/components/GoogleCarouselMetadata/index.jsx
+++ b/src/app/components/GoogleCarouselMetadata/index.jsx
@@ -1,0 +1,166 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+import omit from 'lodash/omit';
+import url from 'url';
+import { models } from '@r/api-client';
+
+import config from 'config';
+import { cleanObject } from 'lib/cleanObject';
+import safeStringify from 'lib/safeStringify';
+
+const T = React.PropTypes;
+const { CommentModel, PostModel } = models;
+const { reddit } = config;
+
+const PARSED_REDDIT_URL = url.parse(reddit);
+const { host: REDDIT_HOST, protocol: REDDIT_PROTOCOL }= PARSED_REDDIT_URL;
+
+export const GoogleCarouselMetadata = (props) => {
+  const { post, comments, pageUrl } = props;
+  // post, comments, and pageUrl all specify `isRequired` in their propTypes.
+  // This makes it easier in debug to tell when we aren't passing in
+  // data to this component. We still keep the explicit checks here
+  // to ensure this component never errors (which could lead to GoogleBot seeing
+  //  seeing a false 5XX error code, negatively affecting SEO).
+  if (!post || !comments || !comments.length) {
+    return null;
+  }
+
+  // TODO: this needs to set nonce on the script tag for csp-reporting
+  return (
+    <div dangerouslySetInnerHTML={ {
+      __html: `
+        <script type='application/ld+json'>
+        ${safeStringify(commentsPageMetaData(pageUrl, post, comments))}
+        </script>
+      `,
+    } } />
+  );
+};
+
+GoogleCarouselMetadata.propTypes = {
+  post: T.instanceOf(PostModel).isRequired,
+  comments: T.arrayOf(T.instanceOf(CommentModel)).isRequired,
+  pageUrl: T.string.isRequired,
+};
+
+const selector = createSelector(
+  (state, props) => state.posts[props.postId],
+  (state, props) => props.commentRecords.map(r => state.comments[r.uuid]),
+  (post, comments) => ({ post, comments }),
+);
+
+export default connect(selector)(GoogleCarouselMetadata);
+
+const commentsPageMetaData = (pageUrl, post, comments) => ({
+  '@context': 'http://schema.org',
+  '@type': 'DiscussionForumPosting',
+  url: addUtmTracking(desktopify(pageUrl)), // we rely on desktopify to ensure
+  // the page url is prefixed with the cannonical reddit url, provided by config
+  headline: post.title,
+  datePublished: ISODate(post.createdUTC),
+  author: {
+    '@type': 'Person',
+    name: post.author,
+    url: `${reddit}/u/${post.author}/`,
+  },
+
+  // Post content info
+  ...formatPostMediaContent(post),
+
+  // comments
+  commentCount: post.numComments,
+
+  comment: {
+    '@list': formatComments(comments),
+  },
+});
+
+const formatPostMediaContent = post => {
+  if (post.selfTextHTML) {
+    return {
+      articleBody: post.expandedContent,
+    };
+  } else if (post.media && post.media.oembed) {
+    const embed = post.media.oembed;
+
+    return {
+      url: desktopify(embed.url || post.cleanUrl),
+      thumbnailURL: firstPreviewUrl(post) || embed.thumbnail_url,
+    };
+  }
+
+  return cleanObject({
+    url: addUtmTracking(desktopify(post.cleanUrl)),
+    thumbnailURL: firstPreviewUrl(post),
+  });
+};
+
+export const firstPreviewUrl = post => {
+  if (post.preview && Array.isArray(post.preview.images) && post.preview.images.length) {
+    return post.preview.images[0].source.url;
+  }
+};
+
+const formatComments = comments => (
+  comments.filter(c => c && c.author).map(c => ({
+    '@type': 'Comment',
+    text: c.bodyHTML,
+    datePublished: ISODate(c.createdUTC),
+    author: {
+      '@type': 'Person',
+      name: c.author,
+      url: `${reddit}/u/${c.author}/`,
+    },
+    commentCount: c.replies.length,
+    aggregateRating: c.score,
+  }))
+);
+
+const ISODate = utcSeconds => (new Date(utcSeconds * 1000)).toISOString();
+
+export const isRedditDomain = hostName => hostName.endsWith('reddit.com');
+
+export const addUtmTracking = urlString => {
+  if (!urlString) { return urlString; }
+
+  const urlObject = url.parse(urlString, true);
+  if (!isRedditDomain(urlObject.hostname || '')) { return urlString; }
+
+  return url.format({
+    // url.format re-uses the search string instead of query object if it exists
+    ...omit(urlObject, 'search'),
+    query: { // we're appending utm params to the query. to do so we need to
+      // splat the query object from the parsed url and add the utm params
+      ...urlObject.query,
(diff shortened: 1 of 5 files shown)


src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,148 @@
...
+        <script type='application/ld+json'>
+        ${safeStringify(commentsPageMetaData(pageUrl, post, comments))}
+        </script>
+      `,
+    } } />
+  );
+};
+
+GoogleCarouselMetadata.propTypes = {
+  post: T.instanceOf(PostModel).isRequired,
```
nramadas: is the `isRequired` check really true? the template seems to assume they might not always exist

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,148 @@
...
+const formatPostMediaContent = post => {
+  if (post.selfTextHTML) {
+    return {
+      articleBody: post.expandedContent,
+    };
+  } else if (post.media && post.media.oembed) {
+    const embed = post.media.oembed;
+
+    return {
+      url: desktopify(embed.url, post.url),
```
prashtx: It looks like `desktopify` just has one parameter. Did you meant to call it with just `embed.url`?

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,148 @@
...
+      thumbnailURL: firstPreivewUrl(post) || embed.thumbnail_url,
```
prashtx: Could you fix the spelling of "Preview", so it's easier to grep for this?

nramadas: typo, `firstPreviewUrl`

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,148 @@
...
+    author: {
+      '@type': 'Person',
+      name: c.author,
+      url: `${reddit}/u/${c.author}`,
+    },
+    interactionCount: [
+      {
+        '@type': 'UserInteraction',
+        userInteractionType: 'http://schema.org/UserComments',
+        value: c.replies.length,
```
prashtx: It looks like this should move to a `commentCount` field on the `Comment` object: http://schema.org/Comment

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,148 @@
...
+      },
+      {
+        '@type': 'UserInteraction',
+        userInteractionType: 'https://schema.org/UserLikes',
+        value: c.score,
```
prashtx: This should probably be an `aggregateRating` field on the `Comment` object: http://schema.org/Comment

prashtx:
I'm not sure if it causes any problems that we include the self text HTML rather than the raw markdown. I think that's the 1X behavior as well, so I'm fine with the working assumption that it's OK (and I'll try to confirm).

A couple of small things and a couple of schema tweaks, otherwise looks awesome!

💅

src/app/components/GoogleCarouselMetadata/index.jsx
schwers: oh this was supposed to be e`embed.url || post.ur`, thanks for catching this!

src/app/components/GoogleCarouselMetadata/index.jsx
schwers: It is true that the component will work just fine without it. But by making it a `propType` the component will complain to us in the future if we break it. I actually debated making the component always render in Dev (as opposed to right now where it only renders for Googlebot and other crawlers). I'm a little uneasy of removing the safety checks just because it seems important that this component doesn't fail, causing Googlebot to get an error. Do you or @prashtx  have a strong opinion about the safety checks or if it should always render in dev-mode?

prashtx: I think we should keep it server-only, since that's an important aspect of the real-life functionality. With that constraint, I think we typically wouldn't see it in dev-mode, since we'll mostly just get a shell from the server, right? If that's true, then I support the `propType` check and think always-render-in-dev-mode is not necessary.

schwers:
💇

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,138 @@
...
+    ...{ query: {
+      utm_source: 'search',
+      utm_medium: 'structured_data',
+    } },
+  });
+};
+
+export const desktopify = url => {
+  if (url) {
+    return url.replace(/https?:\/\/(?:np\.)?(?:m\.)?reddit.com/g, 'www');
```
wting: Can we add a test or two for this function? I'm not sure if I'm understanding if this is desired behavior:

```js
> 'http://m.reddit.com/r/nba/comments/abc123/'.replace(/https?:\/\/(?:np\.)?(?:m\.)?reddit\.com/g, 'www');
'www/r/nba/comments/abc123/'
```

You can also simplify the np and m grouping with a `|`:

```js
https?:\/\/(?:np|m)?\.reddit.com
```

Is the `/g` flag necessary? For example:

```js
> 'http://m.reddit.com/r/nba/comments/abc123?redirect=http://np.redditacom.fakereddit.net/malicious_link'.replace(/https?:\/\/(?:np\.)?(?:m\.)?reddit.com/g, 'www');
'www/r/nba/comments/abc123?redirect=www.fakereddit.net/malicious_link'
```

Is it more robust if we use a URL parser to replace the subdomain?

🔕 nit: `reddit\.com`

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,138 @@
...
+export const addUtmTracking = urlString => {
+  if (!urlString) { return urlString; }
+
+  const urlObject = url.parse(urlString, true);
+  if (!isRedditDomain(urlObject.hostname)) { return urlString; }
+
+  return url.format({
+    // url.format re-uses the search string instead of query object if it exists
+    ...omit(urlObject, 'search'),
+    ...{ query: {
```
nramadas: do we need to splat? wouldn't this work?
```es6
return url.format({
   ...omit(urlObject, 'search'),
   query: {
      utm_source: 'search',
      utm_medium: 'structured_data';
   },
});
```

prashtx:
metadata and spelling changes look good to me. the other reviewers have the code-level stuff well covered.

src/lib/safeStringify.js
```diff
@@ -0,0 +1,6 @@
+export default obj => (
+  JSON.stringify(obj)
```
wting: :no_bell: It doesn't look like it'll be a problem with the current `commentsPageMetaData` implementation, but this will break for circular references. A contrived example:

```js
commentsPageMetadata .. ({
  comment: {
    'count': commentsPageMetadata.commentCount,
    '@list': ..
  },
})
```

src/app/components/GoogleCarouselMetadata/index.jsx
nramadas: Just leaving a comment to say that we discussed in person, and we decided to add a few comments explaining the `return null` in the template.

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,138 @@
...
+  (state, props) => props.commentRecords.map(r => state.comments[r.uuid]),
+  (post, comments) => ({ post, comments }),
+);
+
+export default connect(selector)(GoogleCarouselMetadata);
+
+const commentsPageMetaData = (pageUrl, post, comments) => ({
+  '@context': 'http://schema.org',
+  '@type': 'DiscussionForumPosting',
+  url: addUtmTracking(`${reddit}${pageUrl}`), // page url has a leading slash
```
wting: Can we add an assertion here for this invariant (preferably) or a test that checks `GoogleCarouselMetadata.pageURL` always starts with a leading slash?

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,138 @@
...
+  headline: post.title,
+  datePublished: ISODate(post.createdUTC),
+  author: {
+    '@type': 'Person',
+    name: post.author,
+    url: `${reddit}/u/${post.author}`,
```
prashtx: I think we need to add a trailing slash to ensure that we're linking to the "standard" version of the URL and avoid redirects in the future.

src/app/components/GoogleCarouselMetadata/index.jsx
schwers: This won't technically work, as we want to append to any query params already in the url. My first version wasn't correct either. I updated the relevant bits to be: 
```javascript
query: { // we're appending utm params to the query. to do so we need to
  // splat the query object from the parsed url and add the utm params
  ...urlObject.query,
  utm_source: 'search',
  utm_medium: 'structured_data',
},
```

src/app/components/GoogleCarouselMetadata/index.jsx
```diff
@@ -0,0 +1,138 @@
...
+  const urlObject = url.parse(urlString, true);
+  if (!isRedditDomain(urlObject.hostname)) { return urlString; }
+
+  return url.format({
+    // url.format re-uses the search string instead of query object if it exists
+    ...omit(urlObject, 'search'),
+    ...{ query: {
+      utm_source: 'search',
+      utm_medium: 'structured_data',
+    } },
```
schwers: Good catch, this was not appending the params but I updated it do so, it also reads a little cleaner now.

src/app/components/GoogleCarouselMetadata/index.jsx
schwers: Another good catch. I updated this to rely on url.parse, and it checks the host of the result. It now turns relative urls in canonical desktop urls too, so some of the edge-cases where we'd call `desktopify(post.cleanUrl)` are now fixed.

src/app/components/GoogleCarouselMetadata/index.jsx
schwers: I updated this to rely on `desktopify` (has tests now), so this should always work even if the url is relative without a leading slash, or is an absolute url.

schwers:
💇

wting:
`desktopify` and `addUtmTracking` look good to me!

:fish:

nramadas:
lgtm :tropical_fish:

PR #676, opened 2016-08-12
2 x bug cleanup

schwers:
Depends on fixes to voting mixing in api-client, see https://github.com/reddit/node-api-client/pull/142


schwers:
👓  @uzi or @nramadas or @phil303

phil303:
from a code point of view, this looks good. It's a little overwhelming to go through and make sure you've fixed everyone of these bugs but I think I can safely assume you've already tested them

:fish: alish

PR #678, opened 2016-08-12
Remove permalinkBase as it's not used.

uzi:
The comment model has the permalink as a
cleanPermalink attribute, so use that rather
than constructing it.

👓 @schwers @phil303


phil303:
:fish:

PR #679, opened 2016-08-12
Add the amphtml link element for self posts

prashtx:
For self post pages rendered on the server, we need to add a link
element pointing to the equivalent AMP page, which is how we make the
AMP page visible to the world. Without a configured value, we will not
include an element, so this can be turned on/off via the AMP environment
variable.

### DIFF
--- a/src/config.es6.js
+++ b/src/config.es6.js
@@ -60,6 +60,7 @@ function config() {
     statsURL: process.env.STATS_URL || 'https://stats.redditmedia.com/',
     mediaDomain: process.env.MEDIA_DOMAIN || 'www.redditmedia.com',
     adsPath: process.env.ADS_PATH || '/api/request_promo.json',
+    amp: process.env.AMP,
     manifest: {},
 
     trackerKey: process.env.TRACKER_KEY,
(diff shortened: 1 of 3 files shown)


prashtx:
👓  @uzi (I'm waiting on confirmation that the config change process for mweb is not too onerous, but the code is otherwise ready for review.)

uzi:
:fish:

jamesmaa:
Hey I wanted to bring this to your attention. I've been getting a bug when I've been going to the comments page (e.g. http://localhost:4444/comments/50h2t4/this_bin_that_lets_you_vote_for_things_with/)

I see a TypeError in my chrome browser:
image
I'm referring to the `Cannot read property 'property' of undefined`. 
It was coming from 
https://github.com/reddit/reddit-mobile/blob/master/src/views/pages/BasePage.jsx#L87.

I added a debugger statement to see what was causing the issue:
image

Something about `props.data.set('ampURL', ampURL)` is causing this problem.
If I comment out that line and the one that sets canonical `props.data.set('canonical', canonical)`, then the errors go away. 

Could you try to reproduce the bug and confirm whether it is a problem?

jamesmaa:
I don't know why this is happening, which is why I'm seeing if you could help

prashtx:
James and I spoke separately and understood the cause of this issue (the `canonical` props.data promise resolves to `undefined`, since we use it just for coordination).

PR #680, opened 2016-08-16
Http error support

phil303:
Adds support for 404s and 403s specifically and anything else not a 200 with a more generic error page.

There's a counterpart api-client patch [here](https://github.com/reddit/node-api-client/pull/143) to support this patch as well.

:eyeglasses: @nramadas


src/app/components/ErrorPage/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+const T = React.PropTypes;
+
+const GO_BACK = 'Go Back!';
+const GO_TO_FRONTPAGE = 'To the front page!';
+const ERROR_404 = 'Page Not Found';
+const ERROR_403 = 'Sorry, you don\'t have access to this.';
+const ERROR_GENERIC = 'Something went wrong.';
+
+const ERROR_MAP = {
+  '403': ERROR_403,
```
nramadas: nit, you don't need the quotes

phil303: yea i know. i actually did that on purpose to be explicit about the implicit casting JS does.

src/app/components/ErrorPage/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+const ErrorPage = props => {
+  // NOTE: the second condition is a side-effect of a bug in our platform
+  // code, which is setting the url's referrer to itself. Ideally, only the
+  // first condition is needed.
+  const toFrontpage = props.referrer === '' || props.referrer === props.url;
+  const href = toFrontpage ? '/' : props.referrer;
+  const anchorContent = toFrontpage ? GO_TO_FRONTPAGE : GO_BACK;
+
+  const errorContent = ERROR_MAP[String(props.status)] || ERROR_GENERIC;
```
nramadas: small nit, but you don't need to cast props.status as a string.
```es6
a = {
   1: 'foo',
   '2': 'bar',
};

a[1] === 'foo'; // true
a['1'] === 'foo'; // true
a[2] === 'bar'; // true
a['2'] === 'bar'; // true
```

...javascript

nramadas: also, if you definitely want to convert to a string, would `${props.status}` be any cleaner?

phil303: heh yep. same answer as above. Was purposefully trying to remove the ambiguity there.

src/app/models/CommentsPage.js
```diff
@@ -13,7 +13,8 @@ export const newCommentsPage = (id, params) => ({
   id,
   params,
   postId: params.id,
-  loading: true,
   loadingMoreChildren: {},
   results: [],
+  loading: true,
```
nramadas: i think this was above `loadingMoreChildren` due to alphabetization

phil303: Look at the original, since when does `l` come after `p`? :trollface: 

I grouped them by functional similarity but I don't feel terrrrrribly strongly either.

src/app/router/handlers/CommentsPage.js
```diff
@@ -50,6 +52,12 @@ export default class CommentsPage extends BaseHandler {
 
     dispatch(commentsPageActions.fetchRelevantContent());
     dispatch(commentsPageActions.visitedCommentsPage(this.urlParams.postId));
+
+    const commentsPageId = paramsToCommentsPageId(commentsPageParams);
+    await waitForState(state => {
+      const commentsPage = state.commentsPages[commentsPageId];
+      return commentsPage && !!commentsPage.responseCode;
+    }, state => dispatch(setStatus(state.commentsPages[commentsPageId].responseCode)));
```
nramadas: 🔕  since `waitForState` blocks with the `await` anyway, passing in a success handler, it might look cleaner to put that bit after the `wait`.

```es6
  await waitForState(state => {
    const commentsPage = state.commentsPages[commentsPageId];
    return commentsPage && !!commentsPage.responseCode;
  });

  const newState = getState();
  dispatch(setStatus(newState.commentsPages[commentsPageId].responseCode));
```

src/app/router/handlers/PostsFromSubreddit.js
```diff
@@ -42,5 +44,10 @@ export default class PostsFromSubreddit extends BaseHandler {
     const { subredditName } = subredditPostsParams;
     dispatch(subredditActions.fetchSubreddit(subredditName));
     fetchUserBasedData(dispatch);
+
+    await waitForState(state => {
+      const postsList = state.postsLists[postsListId];
+      return postsList && !!postsList.responseCode;
+    }, state => dispatch(setStatus(state.postsLists[postsListId].responseCode)));
```
nramadas: 🔕 same `waitForState` potential here

src/app/models/CommentsPage.js
nramadas: :derp: i just assumed it was alphabetical because s.chwers alphabetizes a lot of these kinds of keys

nramadas:
small nits, but :fish:

PR #683, opened 2016-08-17
Switch mobile banner click tracking from tune to branch

uzi:


uzi:
Note that this is a WIP ... it sets the url correctly, but clicking it doesn't seem to have the desired effect.

src/lib/initBanner.es6.js
```diff
@@ -1,25 +1,54 @@
...
   if (!shouldShowBanner()) { return; }
 
   // create a container for our banner and add it to the body
   const bannerContainer = document.createElement('div');
   bannerContainer.className = 'SmartBannerContainer tween';
   document.body.appendChild(bannerContainer);
 
+  let url = clickUrl;
+  if (!url) {
+    branch.init('key_live_hoc05HaCXaME10UMwyj3filpqzfu2Ue6', function(err, data) {
```
phil303: this should be a constant

src/lib/initBanner.es6.js
```diff
@@ -1,25 +1,54 @@
...
+      // callback to handle err or data
+      window.referring_link = data.referring_link;
```
phil303: not sure what these are doing or being used for? Can you give me a quick rundown and possibly add a comment.

phil303:
Just did a quick pass at it since I was reviewing your other patch. Apologies if I jumped the gun.

src/lib/initBanner.es6.js
nramadas: 🔕  small nit, but i always prefer es6 style anonymous function to es5 style. especially when they are callbacks

src/lib/initBanner.es6.js
```diff
@@ -1,25 +1,54 @@
...
+export default function initBanner(clickUrl) {
   if (!shouldShowBanner()) { return; }
 
   // create a container for our banner and add it to the body
   const bannerContainer = document.createElement('div');
   bannerContainer.className = 'SmartBannerContainer tween';
   document.body.appendChild(bannerContainer);
 
+  let url = clickUrl;
+  if (!url) {
```
nramadas: when would we have a `clickUrl`? if we're using branch to generate the urls, this should not exist at all, right?

src/lib/initBanner.es6.js
```diff
@@ -1,25 +1,54 @@
...
+        // branch dashboard and allow you sort/parse data. Optional/not required.
+        // tags: [ 'tag1', 'tag2' ],
+        data: {
+          // Pass in data you want to appear and pipe in the app,
+          // including user token or anything else!
+          '$og_image_url': window.location.href,
+          '$deeplink_path': window.location.href,
+        },
+      },
+      function(err, link) {
```
nramadas: 🔕  same with this callback

src/lib/initBanner.es6.js
```diff
@@ -1,25 +1,54 @@
...
+        // set 'link' to the element ID behind your existing banner.
+        window.document.getElementsByClassName('SmartBanner__Button')[0].href = link;
```
nramadas: instead of looking up the element and changing the button's url here, why not just delay the render until this point? so, this callback would look something like:

```es6
// render the banner
ReactDOM.render(
  <SmartBanner
    url={ url }
    { /* etc */ }
  />,
  bannerContainer
);

// remaining stuff, like the banner animation
```

src/lib/smartBannerState.es6.js
```diff
@@ -4,7 +4,6 @@ import constants from '../constants';
 
 const BASE_VAL = {
   show: false,
-  impressionUrl: '',
   clickUrl: '',
```
nramadas: i don't we're using this anymore either. we can get rid of it, i think

nramadas:
:nail_care:

phil303:
@uzi can we kill this?

PR #685, opened 2016-08-17
Cosmetic fixes for Register page

jamesmaa:
Made cosmetic fixes to the Register page.

- Change error tag from <span> to <p>, a block level html tag
- Add `DEFAULT_ERRORS` for handling `errorType` = 'UNKNOWN_ERROR'
- Change default error text to red

👓 : @nramadas

### DIFF
--- a/src/app/components/LoginRegistrationForm/styles.less
+++ b/src/app/components/LoginRegistrationForm/styles.less
@@ -30,9 +30,17 @@
   }
 
   &__checkbox {
-    position: relative;
-    top: 5px;
+    margin-top: 15px;
+  }
+
+  &__checkbox-label {
     vertical-align: top;
+    margin-left: 4px;
+  }
+
+  &__error-text {
+    color: @red;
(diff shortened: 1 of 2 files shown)


nramadas:
@jamesmaa quick favor, could you a few screenshots?

jamesmaa:
Yep!
image
image
image

nramadas:
Thanks! While we're in there, can we add two more cosmetic changes?

0. There seems to be an extra line between "Choose a unique username" and "Choose a unique password". Let's get rid of one of the lines
0. Let's add some spacing between the checkbox and "Subscribe to newsletter". Maybe 4px or so?
0. Also with the checkbox, it doesn't seem to be vertically aligned with the text. Can we try adjusting the line-height, or giving both a `diplay: inline-block; vertical-align: middle;` treatment?

jamesmaa:
`top` seems more vertically aligned than `middle`
image
image

nramadas:
:fish:

PR #687, opened 2016-08-19
Test lib

jamesmaa:
Wrote some tests for `2x/src/lib`

Here is the current status:
image

I'm having trouble writing some of the tests. @nramadas @phil303 do you guys have experience writing tests with the following constraints? 
- Dealing with browser capabilities (e.g. localStorage, cookies)
- Mocking objects/requests

I skipped some tests because they seemed pretty short/obvious

### DIFF
--- a/test/lib/RingStack.test.js
+++ b/test/lib/RingStack.test.js
@@ -1,0 +1,42 @@
+import chai from 'chai';
+import sinonChai from 'sinon-chai';
+import RingStack from '../../src/lib/RingStack';
+
+const expect = chai.expect;
+
+chai.use(sinonChai);
+
+// Not read-only. Could set private variables
+
+describe('lib: RingStack', () => {
+  const stack = new RingStack(3);
+  it('insert elements', () => {
+    stack.push(1);
+    expect(stack.values()).to.eql([1]);
+    stack.push(2);
+    expect(stack.values()).to.eql([2, 1]);
+    stack.push(3);
+    expect(stack.values()).to.eql([3, 2, 1]);
+  });
+
+  it('overwrite elements in a particular order', () => {
+    stack.push(4);
+    expect(stack.values()).to.eql([4, 3, 2]);
+    stack.push(5);
+    expect(stack.values()).to.eql([5, 4, 3]);
+  });
+
+  it('is not read only. could set private variables', () => {
+    stack.size = 4;
+    expect(stack.size).to.equal(4);
+    stack.push(6);
+    expect(stack.values()).to.eql([6, 5, 4]);
+    expect(stack.head).to.equal(3);
+    stack.push(7);
+    expect(stack.values()).to.eql([7, 6, 5, 4]);
+    expect(stack.head).to.equal(0);
+    stack.push(8);
+    expect(stack.values()).to.eql([8, 7, 6, 5]);
+    expect(stack.head).to.equal(1);
+  });
+});
--- a/test/lib/classNames.test.js
+++ b/test/lib/classNames.test.js
@@ -1,0 +1,36 @@
+import chai from 'chai';
+import sinonChai from 'sinon-chai';
+import classNames from '../../src/lib/classNames';
+
+const expect = chai.expect;
+
+chai.use(sinonChai);
+
+describe('lib: classNames', () => {
+  it('is a function', () => {
+    expect(classNames).to.be.a('function');
+  });
+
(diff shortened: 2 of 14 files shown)


test/lib/classNames.test.js
```diff
@@ -0,0 +1,37 @@
...
+    it('is a function', () => {
+        expect(classNames).to.be.a('function');
+    });
+
+    it('creates correct classname for string types', () => {
+        expect(classNames('hello', 'world', 'baby')).to.equal('hello world baby');
+        expect(classNames('hello')).to.equal('hello');
+    });
+
+    it('allows duplicate css classes', () => {
```
phil303: :no_bell: arguably that's an oversight on my part and not a feature :)

test/lib/classNames.test.js
```diff
@@ -0,0 +1,37 @@
...
+        expect(classNames(['hello', 'world', 'baby'])).to.equal('hello world baby');
+    });
+
+    it('creates correct classname for object types', () => {
+        expect(classNames({ baz: false, bob: true })).to.equal('bob');
+    });
+
+    it('does not support nested dictionaries. Only allows truthy values', () => {
+        const d = { baz: true };
+        d.bob = d;
```
phil303: this test makes me a weeeee bit nervous. If the day comes we decide for it to recurse then this is just going to overflow the stack. Maybe just make this another object?

test/lib/cleanObject.test.js
```diff
@@ -0,0 +1,20 @@
...
+const expect = chai.expect;
+
+chai.use(sinonChai);
+
+describe('lib: cleanObject', () => {
+    it('is a function', () => {
+        expect(cleanObject).to.be.a('function');
+    });
+
+    const obj = {foo: null, bar: undefined, baz: true};
```
phil303: style nit, a space around the curlies

test/lib/cleanObject.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+    it('removes null and undefined values', () => {
+        expect(cleanObject(obj)).to.eql({baz: true});
```
phil303: same here

test/lib/RingStack.test.js
```diff
@@ -0,0 +1,43 @@
...
+        expect(stack.head).to.equal(3);
+        stack.push(7);
+        expect(stack.values()).to.eql([7, 6, 5, 4]);
+        expect(stack.head).to.equal(0);
+        stack.push(8);
+        expect(stack.values()).to.eql([8, 7, 6, 5]);
+        expect(stack.head).to.equal(1);
+
+    })
+});
```
phil303: style nit, if you're using sublime text, there's a preference called something like "ensure newline at EOF", mind turning that on?

phil303:
This is awesome work! Thanks for taking it on. There are certain ways to mimic _most_ of the dom in node, namely the library [jsdom](https://github.com/tmpvar/jsdom) and its compatriots. But let's chat/evaluate whether we need to use it here. We might be able to get away with not pulling in another library.

:nail_care:

jamesmaa:
Cool thanks for the comments, I found the Sublime preference for `ensure_newline_at_eof_on_save`

With some of the tests, I wrote tests that froze the behavior of the functions. e.g. for classNames, I knew that the function would allow duplicate class names even though it was not an intended function. I figured that it would be better to use these tests in an expository manner to document how functions behave, so devs in the future could change the function/tests if they wish to alter the behavior. 

I'm not sure if that's the purpose of these tests, since the libs don't have documentation themselves.

phil303:
I very much got that vibe and I'm totally for it (which is why I no_bell'd that one test). When you say "the libs don't have documentation themselves", I'm curious what you'd like to see in that domain. If you've heard of [jsdocs](https://en.wikipedia.org/wiki/JSDoc), do you think that'd be helpful here? Or are you looking for more thorough, more prose-like documentation on a wiki of some sort?

jamesmaa:
I'm leaning more towards "simple -> better" so I think as long 

I was working with the assumption that if you don't have documentation, your code becomes your documentation. So as long I don't have to thoroughly read the code to understand intended behavior, I'm fine with either. jsdocs seems like the easier solution, and I don't think we even need to be that thorough since most of the functions I'm testing are pretty simple anyways.

phil303:
:fish:! make sure to rebase

PR #688, opened 2016-08-19
Fix a couple of routes

uzi:
There are some routes that would be helpful to fix in 1X, and this takes care of that.

Adds `/comments/:listingId/:listingTitle/:commentId` and fixes a bug with `/comments/:listingId/:listingTitle?`

👓 @phil303 @nramadas


phil303:
the code itself looks fine. i don't know much about this codebase though so take what you will from that :fish:

PR #690, opened 2016-08-19
Hide GoldInfo if user has no gold

uzi:
👓 @phil303 @nramadas 

This is how it is:

screen shot 2016-08-19 at 11 06 34 am

This is how the patch makes it:

screen shot 2016-08-19 at 11 06 43 am

An alternative would be a variant of what we do if you view someone else's profile:

screen shot 2016-08-19 at 11 06 54 am

Instead of "give foo gold", it could be a "Buy gold"... which, granted isn't implemented yet.

### DIFF
--- a/src/app/components/UserProfileSummary/index.jsx
+++ b/src/app/components/UserProfileSummary/index.jsx
@@ -30,9 +30,7 @@ export const UserProfileSummary = props => {
           <UserProfileBadgeIcon iconName='cake' color='mint' />
         </UserProfileBadge>
       </UserProfileRow>
-      <UserProfileRow>
-        <GoldInfo { ...props } />
-      </UserProfileRow>
+      <GoldInfo { ...props } />
     </div>
   );
 };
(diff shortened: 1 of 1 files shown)


src/app/components/UserProfileSummary/index.jsx
```diff
@@ -46,13 +44,17 @@ const GoldInfo = props => {
   const { user, isMyUser } = props;
 
   if (isMyUser) {
+    if (!user.goldExpiration) { return null; }
```
phil303: instead of putting this check in `GoldInfo`, it'd be more obvious what's happening if you put this a level above. So it'd look more like this in `UserProfileSummary`:
```js
export const UserProfileSummary = props => {
   // ....
      </UserProfileBadge>
     </UserProfileRow>
     { goldExpiration ? <GoldInfo goldExpiration={ goldExpiration /> : null }
  </div>
}

```

phil303:
:nail_care:

src/app/components/UserProfileSummary/index.jsx
uzi: Oh, you're right -- fixing, thanks

uzi: Oh wait... the problem is that we're also displaying GoldInfo if it's not isMyUser ... so it's not as easy to do this check there.

phil303:
:fish:

PR #691, opened 2016-08-22
Add visitedPosts and recentClicks to CookieSync

jamesmaa:
👓 @dwick 

Questions:
Does the cookie key matter?

### DIFF
--- a/src/app/side-effect-components/CookieSync.js
+++ b/src/app/side-effect-components/CookieSync.js
@@ -1,16 +1,39 @@
+import uniq from 'lodash/uniq';
 import { makeCookieArchiver } from '@r/redux-state-archiver';
+import { RECENT_CLICKS_LENGTH } from 'app/constants';
+
+const recentClicksSelector = state => {
+  const recentClicks = state.visitedPosts.concat(Object.keys(state.expandedPosts));
+  return uniq(recentClicks).slice(0, RECENT_CLICKS_LENGTH);
+};
+
+const recentSubredditsSelector = state => {
+  return state.recentSubreddits
+    .filter(sub => state.subreddits[sub.toLowerCase()])
+    .map(sub => state.subreddits[sub.toLowerCase()].name);
+};
+
+const combiner = (theme, compact, over18, recentSubreddits, recentClicks, username, euCookieNotice) => {
+  const res = {
+    theme,
+    compact,
+    over18,
+    EUCookieNotice: euCookieNotice,
+  };
+
+  res[`${username}_recent_srs`] = recentSubreddits.join(',');
+  res[`${username}_recentclicks2`] = recentClicks.map(post => `t3_${post}`).join(',');
+
+  return res;
+};
 
 export default makeCookieArchiver(
   state => state.theme,
   state => state.compact,
   state => state.preferences.over18,
-  state => state.recentSubreddits,
+  recentSubredditsSelector,
+  recentClicksSelector,
+  state => !state.user.loggedOut && state.user.name ? state.user.name : '',
   state => state.euCookieNotice.numberOfTimesShown,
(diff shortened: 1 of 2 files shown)


dwick:
The keys do matter, they need to be the same as r2. The should be `recentclicks2` and `recent_srs`. Also, `recent_srs` should be comma separated thing fullnames (`t5_abc`).

💅

jamesmaa:
image
image

💇

src/app/side-effect-components/CookieSync.js
```diff
@@ -1,14 +1,37 @@
...
+};
+
+const combiner = (theme, compact, over18, recentSubreddits, recentClicks, username) => {
+  const res = {
+    theme,
+    compact,
+    over18,
+  };
+
+  res[`${username}_recent_srs`] = recentSubreddits.join(',');
```
dwick: does this handle `undefined` or do we end up with `"undefined_recent_srs"`?

jamesmaa: Added a check to make sure username is not undefined

src/app/side-effect-components/CookieSync.js
```diff
@@ -32,6 +32,6 @@ export default makeCookieArchiver(
   state => state.preferences.over18,
   recentSubredditsSelector,
   recentClicksSelector,
-  state => state.user.loggedOut ? '' : state.user.name,
+  state => !state.user.loggedOut && state.user.name ? state.user.name : '',
```
dwick: 🔕 missed the `state.user.loggedOut` check, not sure this is actually necessary. could also just make it `state.user.name` which is what we're actually after.

dwick:
🐟

src/app/side-effect-components/CookieSync.js
jamesmaa: I think we need it because `state.user.name` == `me` when logged out
image

src/app/side-effect-components/CookieSync.js
```diff
@@ -1,14 +1,37 @@
...
+import { RECENT_CLICKS_LENGTH } from 'app/constants';
+
+const recentClicksSelector = state => {
+  const recentClicks = state.visitedPosts.concat(Object.keys(state.expandedPosts));
+  return take(uniq(recentClicks), RECENT_CLICKS_LENGTH);
+};
+
+const recentSubredditsSelector = state => {
+  return state.recentSubreddits
+              .filter(sub => state.subreddits[sub])
```
phil303: I think you're going to have a problem here. The `recentSubreddits` state stores the subreddit names as their canonical forms, aka they have upper and lowercases. The `subreddits` state stores things in just their lowercased form, which means you're going to have lookup misses here.

src/app/side-effect-components/CookieSync.js
```diff
@@ -1,14 +1,37 @@
+import { take, uniq } from 'lodash/array';
 import { makeCookieArchiver } from '@r/redux-state-archiver';
+import { RECENT_CLICKS_LENGTH } from 'app/constants';
+
+const recentClicksSelector = state => {
+  const recentClicks = state.visitedPosts.concat(Object.keys(state.expandedPosts));
+  return take(uniq(recentClicks), RECENT_CLICKS_LENGTH);
```
phil303: nit. You can actually get away with not using `take`.
```js
return uniq(recentClicks).slice(0, RECENT_CLICKS_LENGTH);
```

src/app/side-effect-components/CookieSync.js
jamesmaa: nice catch. Do you know why there is a discrepancy between the two states?

jamesmaa:
💇

dwick:
🐟 lgtm

PR #693, opened 2016-08-22
Add Docker image building via Drone.

gtaylor:
This pull request has a `.drone.yml` for automatically building Docker images, and a `Dockerfile.staging` to determine the nature of the Docker image that gets built. These will be used with the upcoming Kubernetes-based staging environment.

Here's the run-down of a typical build:

* Developer pushes to a branch on the canonical reddit/reddit-mobile repo.
* Github sends a webhook request to Drone.
* Drone clones the source, restores `node_modules` from a local cache, then runs through a full build/lint/test cycle.
* If the build passes, a Dockerfile is built. We more or less `cp` the stuff that we tested in the previous step, which becomes our staging image.
* If all goes well, we push the resulting image to Docker Hub under [reddit/reddit-mobile](https://hub.docker.com/r/reddit/reddit-mobile/tags/). The tag will be the branch name.

:eyeglasses: @nramadas @phil303 @schwers 

Of particular emphasis for review are the env vars in `Dockerfile.staging`. I'm not sure what we want to default to. Also, some of these appear to be build-time vars, while others seem to be runtime vars. We'll want to move the former to `.drone.yml` and the latter to the Kubernetes ConfigMaps. I'd appreciate guidance on where the stuff currently in `Dockerfile.staging` should end up.


Dockerfile.staging
```diff
@@ -0,0 +1,17 @@
+# reddit-mobile Dockerfile for deploying on staging
+
+FROM reddit/reddit-nodejs:latest
+
+# TODO: Do these make sense for staging?
+ENV MINIFY_ASSETS "false"
```
phil303: yea i think we want this to be true right? staging should be as close to prod as possible.

Dockerfile.staging
```diff
@@ -0,0 +1,17 @@
...
+ENV OAUTH_CLIENT_ID XXXXXXXXX
+ENV SECRET_OAUTH_CLIENT_ID XXXXXXXXX
+ENV OAUTH_SECRET XXXXXXXXX
+
+ADD . /src/
+WORKDIR /src
+
+EXPOSE 4444
+# start the server
+CMD npm run server
```
phil303: should this be `npm start`?

Dockerfile.staging
```diff
@@ -0,0 +1,17 @@
+# reddit-mobile Dockerfile for deploying on staging
+
+FROM reddit/reddit-nodejs:latest
+
+# TODO: Do these make sense for staging?
+ENV MINIFY_ASSETS "false"
+ENV GOOGLE_ANALYTICS_ID UA-XXXXXX-1
+ENV OAUTH_CLIENT_ID XXXXXXXXX
+ENV SECRET_OAUTH_CLIENT_ID XXXXXXXXX
```
nramadas: Are these replaced with real client ids within the container, or will they always remain as 'X's? I wonder if this will prevent us from being able to authenticate with oauth against production.

Dockerfile.staging
gtaylor: I'll defer to you all on that.

Something else I was wondering: Do some or all of these need to be set before `npm run build`, or are they runtime variables?

Dockerfile.staging
gtaylor: @phil303 Pulled that from the readme.md. Seems to start something that serves 2X. I'm not sure if it's the same as `npm start`, though.

Dockerfile.staging
phil303: Actually, that's a good point. This is a build related config but I've never seen it before. @schwers Do you know what this is?

Dockerfile.staging
gtaylor: In staging, we can pass in whatever env vars we'd like. I should have clarified in the PR, but these were meant to be placeholder values until I had a chance to harass someone.

I guess we need to divide this up into two buckets:

* Env vars that we need for build time
* Env vars that we _can_ pass in at runtime. If at all possible, any sensitive tokens should go here.

I can clear out anything not matching the above.

.drone.yml
```diff
@@ -0,0 +1,30 @@
+cache:
+  mount:
+    - node_modules
+
+build:
+  image: reddit/reddit-nodejs:latest
+  pull: true
+  commands:
+    - npm install
+    - npm run build
```
nramadas: we don't need to build before linting and testing, so it might be better to move this to execute after lint and test do. We can also save some time by not building if the fails to lint or pass tests.

so, the order would look like:
```
- npm install
- npm run lint
- npm run test
- npm build
```

Dockerfile.staging
nramadas: ok cool, that makes sense to me. yeah, we don't need any of those env variables to build, we only need them at runtime

.drone.yml
gtaylor: Pushed in 00b7cfc.

Dockerfile.staging
gtaylor: @nramadas removed them all in db44e03.

Dockerfile.staging
schwers: `MINIFY_ASSETS` is a legacy option from 1X, if this is running `npm run build` minification will happen anyway

nramadas:
this LGTM :fish:

phil303:
:fish:sauce

schwers:
:fish: 🍣

uzi:
:ramen:

gtaylor:
Rebased and merged in 23e66cd6f2b9f4e77e69794a74c53c55c3576bde.

PR #694, opened 2016-08-23
Resolve data promises with non-undefined

prashtx:
We assume that promise values in `props.data` will resolve to objects,
so we should avoid resolving them with `undefined`, even if we we're in
a situation where the value has become irrelevant.


prashtx:
👓  @uzi @schwers

schwers:
:fish:

PR #695, opened 2016-08-23
For the soft launch, keep the robots away

nramadas:
:eyeglasses: @schwers || @phil303


prashtx:
This is risky, because bots will see this and stop crawling m.reddit.com entirely. We'll essentially lose all mweb crawls during the period of the soft launch.

nramadas:
This is not the right approach. Closing.

PR #696, opened 2016-08-24
Add ReCaptcha to account registration

uzi:
Add a new ReCaptcha component that follows the API at
https://developers.google.com/recaptcha/docs/display
and make use of it in account registration.

### DIFF
--- a/src/views/components/ReCaptcha.jsx
+++ b/src/views/components/ReCaptcha.jsx
@@ -1,0 +1,63 @@
+import React from 'react';
+
+import BaseComponent from './BaseComponent';
+
+const T = React.PropTypes;
+
+class ReCaptcha extends BaseComponent {
+  static propTypes = {
+    elementId: T.string,
+    sitekey: T.string.isRequired,
+    theme: T.oneOf(['light', 'dark']),
+    type: T.oneOf(['image', 'audio']),
+    size: T.oneOf(['normal', 'compact']),
+    tabindex: T.number,
+    onSuccess: T.func,
+    onExpiration: T.func,
+  };
+
+  static defaultProps = {
+    elementId: 'g-recaptcha',
+    theme: 'light',
+    type: 'image',
+    size: 'normal',
+    tabindex: 0,
+  };
+
+  componentDidMount() {
+    const {
+      elementId,
+      sitekey,
+      theme,
+      type,
+      size,
+      tabindex,
+      onSuccess,
+      onExpiration,
+    } = this.props;
+
+    window.grecaptcha.render(elementId, {
+      sitekey,
+      theme,
+      type,
+      size,
+      tabindex,
+      callback: onSuccess,
+      'expired-callback': onExpiration,
+    });
+  }
+
+  shouldComponentUpdate() {
+    return false;
+  }
+
+  render() {
+    const { elementId } = this.props;
+
+    return (
+      <div id={ elementId } />
+    );
+  }
+}
+
+export default ReCaptcha;
--- a/src/views/pages/login.jsx
+++ b/src/views/pages/login.jsx
@@ -13,6 +13,7 @@ import BasePage from './BasePage';
 import SnooIconHeader from '../components/snooiconheader';
 import Modal from '../components/Modal';
 import ForgotPassword from '../components/forgotpassword';
+import ReCaptcha from '../components/ReCaptcha';
 
 import MinimalInput from '../components/formElements/minimalinput';
 import SquareButton from '../components/formElements/SquareButton';
@@ -24,6 +25,10 @@ const ERROR_MESSAGES = {
   DEFAULT: 'Sorry, Something has gone wrong and we\'re not sure what',
   NEWSLETTER_NO_EMAIL: 'Sorry, we need an email to send you the newsletter',
 };
+
+const CAPTCHA_ERRORS = [
+  'BAD_CAPTCHA',
+];
 
 const EMAIL_ERRORS = [
   'BAD_EMAIL',
@@ -50,6 +55,7 @@ const ERROR_TYPES = [
   ...EMAIL_ERRORS,
   ...USER_ERRORS,
   ...PASS_ERRORS,
+  ...CAPTCHA_ERRORS,
 ];
 
 const PASS_FIELD_TYPES = {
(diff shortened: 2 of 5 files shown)


uzi:
Note that folks are welcome to code review this, but I'm not going to eyeglass anyone until it's truly ready.  AFAIK, the only think to do here is to take care of the `sitekey` values, which may vary between localhost development, staging and production.  That will be figured before this lands when all stakeholders have a chance to look at this.

src/views/pages/login.jsx
```diff
@@ -475,6 +493,13 @@ class LoginPage extends BasePage {
                                                     error,
                                                     errorMessage,
                                                     newsletter) : null }
+          { registerMode ? <ReCaptcha
+                             sitekey='XXX SITEKEY GOES HERE XXX'
+                             callback={ this.ReCaptchaCallback }
+                             theme={ captchaTheme }
+                           />
+                         : null }
+          { registerMode ? captchaError : null }
```
uzi: Yep, that `sitekey=` is going to have to change.

uzi:

I'll look at the sitekey stuff (and the lint bugs) -- wanna get a head start reviewing?  Thanks!

src/views/components/ReCaptcha.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+import BaseComponent from './BaseComponent';
+
+const T = React.PropTypes;
+
+class ReCaptcha extends BaseComponent {
+  static propTypes = {
+    elementId: T.string,
+    sitekey: T.string.isRequired,
+    theme: T.string, // 'light' and 'dark'
```
schwers: You can use `T.oneOf` here to add debug checks for the correct values. For theme this can be `T.oneOf(['light', 'dark'])`

src/views/components/ReCaptcha.jsx
```diff
@@ -0,0 +1,67 @@
...
+    type: T.string, // 'image' and 'audio'
+    size: T.string, // 'normal' and 'compact'
+    tabindex: T.number,
+    callback: T.func,
```
schwers: small nit, are`callback` and `expiredCallback`  required? It looks like they're passed directly to the `grecaptcha` api, not sure if it's happy with getting `undefined`

src/views/components/ReCaptcha.jsx
```diff
@@ -0,0 +1,67 @@
...
+      type,
+      size,
+      tabindex,
+      callback,
+      'expired-callback': expiredCallback,
+    });
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
+    return false;
```
schwers: 🎉  Thanks for making this explicit

src/views/components/ReCaptcha.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+  static defaultProps = {
+    elementId: 'g-recaptcha',
+    theme: 'light',
+    type: 'image',
+    size: 'normal',
+    tabindex: 0,
+  };
+
+  constructor(props) {
```
schwers: Do we need an explicit constructor here if we're only calling `super`?

src/views/components/ReCaptcha.jsx
nramadas: another nit, can we name the callbacks to something more descriptive?  maybe:
`callback` -> `onSuccess`
`expiredCallback` -> `onExpiration`

src/views/pages/login.jsx
```diff
@@ -430,6 +445,9 @@ class LoginPage extends BasePage {
     const globalError = error.global ?
       (<p className='minimalInput__error-text'>{ errorMessage }</p>) : null;
 
+    const captchaError = error.captcha ?
```
nramadas: nit, let's keep our ternary style consistent.
```es6
const captchaError = error.captch
  ? <p className='minimalInput__error-text'>Please provide a valid captcha</p>
  : null;
```

src/views/pages/login.jsx
```diff
@@ -475,6 +493,13 @@ class LoginPage extends BasePage {
                                                     error,
                                                     errorMessage,
                                                     newsletter) : null }
+          { registerMode ? <ReCaptcha
```
nramadas: same here w.r.t. ternary style.

src/views/components/ReCaptcha.jsx
uzi: Cool, will do.

src/views/pages/login.jsx
```diff
@@ -494,7 +494,7 @@ class LoginPage extends BasePage {
                                                     errorMessage,
                                                     newsletter) : null }
           { registerMode ? <ReCaptcha
-                             sitekey='XXX SITEKEY GOES HERE XXX'
+                             sitekey='6LeTnxkTAAAAAN9QEuDZRpn90WwKk_R1TRW_g-JC'
```
schwers: Should this be in `config.es6.js` instead? Also, just as a small note in either case, can you make a note as to why it's okay to hard-code this in the client?

src/views/components/ReCaptcha.jsx
uzi: I'm just trying to keep consistent with their spec.

src/views/components/ReCaptcha.jsx
uzi: No, we don't -- forgot to remove that.

src/views/pages/login.jsx
uzi: I was doing the same as the one right above it!  I can change both.

src/views/pages/login.jsx
uzi: ok

src/views/pages/login.jsx
uzi: Sure, will do on both.

src/views/components/ReCaptcha.jsx
nramadas: right, but this component is our code, and should be more useful for us. just because google gave their callbacks bad names doesn't mean we need to stick to them. giving the react component's callbacks more useful names will help a future developer understand what's going on.

schwers:
:fish: 🐠  🍹

PR #697, opened 2016-08-24
Add recentclicks2 and recent_srs to cookies

jamesmaa:
👓 : @dwick

### DIFF
--- a/src/lib/visitedSubreddits.es6.js
+++ b/src/lib/visitedSubreddits.es6.js
@@ -4,10 +4,11 @@ import cookies from 'cookies-js';
 import constants from '../constants';
 import localStorageAvailable from './localStorageAvailable';
 
-// Return an array of ids of the posts the user has visited recently.
+// Return an array of ids of the subreddits the user has visited recently.
 // Return [] if we are unable to access such a list.
-export function getVisitedPosts(username) {
-  const key = [username, constants.VISITED_POSTS_KEY].join('_');
+export function getVisitedSubreddits(username) {
+  const key = [username, constants.RECENT_SUBREDDITS_KEY].join('_');
+
   const localStorageString = localStorageAvailable() ? global.localStorage.getItem(key) : '';
   const localStorageArr = localStorageString ? localStorageString.split(',') : [];
   const cookieString = cookies.enabled ? cookies.get(key) : '';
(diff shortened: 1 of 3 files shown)


src/routes.jsx
```diff
@@ -461,8 +474,13 @@ function routes(app) {
       commentId: ctx.params.commentId,
     });
 
-    const visited = getVisitedPosts();
-    setVisitedPosts([ctx.params.listingId].concat(visited));
+    const data = props.data;
+    const user = data.get('loggedOutUser') || data.get('user');
```
dwick: if the user is logged out what happens here? don't we want it to be `_cookie_name` in that case?

src/lib/visitedSubreddits.es6.js
```diff
@@ -0,0 +1,43 @@
...
+
+// Return an array of ids of the subreddits the user has visited recently.
+// Return [] if we are unable to access such a list.
+export function getVisitedSubreddits(username) {
+  if (!localStorageAvailable()) {
+    return [];
+  }
+
+  const key = [username, constants.RECENT_SUBREDDITS_KEY].join('_');
+  const visitedString = global.localStorage.getItem(key);
```
dwick: we want to be able to grab these from the cookies as well if they were set by r2

dwick:
💅

src/routes.jsx
jamesmaa: The assumption I'm working with here is that when a user is loggedout, data will have `loggedOutUser`, whereas when a user is loggedIn, data will have `user`.

loggedOut: 
image

loggedIn:
image

jamesmaa: the cookie name is set in `setVisitedSubreddits`. it'll save as `_recent_srs`

jamesmaa:
💇

src/routes.jsx
```diff
@@ -479,7 +479,7 @@ function routes(app) {
     user.then(user => {
       const username = user.body.name;
       const visited = getVisitedPosts(username);
-      setVisitedPosts(username, [ctx.params.listingId].concat(visited));
+      setVisitedPosts(username, ['t3_' + ctx.params.listingId].concat(visited));
```
dwick: 🔕 are there no constants for types so you dont have to remember them?

dwick:
🐟 lgtm

jamesmaa:
deploy broke :( fixing

PR #698, opened 2016-08-25
Top3 Posts to SEO traffic experiment

uzi:


uzi:
Still a WIP with some things to resolve.

uzi:
Took a crack at this that I think is fairly reasonable.  The initial ask was to maintain the three comments, then show relevance and then the rest of the comments, but it felt it wasn't get best answer because:

1) It was going to be technically difficult.  Besides having to get the first three comments (which we already had), we'd also need to construct a tree of the remaining comments.  After that, we'd need to present them, and if the slice is mid-tree, we'd have to put in some work to manage maintaining the initial depth of the tree.  This is because the comment tree is turned into a DOM tree for display.  We could try inserting the relevance block, but then it'd be indented with the tree.
2) It felt like it was going to be an awkward user experience to interrupt a comment tree... let a tree finish before you slice, then show relevance, and then continue with the next top level comment tree and on.

We're looking to display at least three comments, then relevance, and then the rest of the comments.  We do that by looking at the number of total comments in the first three comment trees... then figure out where to slice the tree based on that.  If the first three comment trees have no replies, then they all have a count of 1 and we slice after the third.  If the first has 10 replies, then we slice right after it.

I'll check in with the powers that be about this solution tomorrow.

uzi:
👓  @schwers @nramadas @prashtx 

Ok, this is the behavior that was settled on with the PMs.  The real meat is in featureflags.es6.js and in listing.jsx ... the other stuff is to get rid of react complaining about component props being the wrong type or missing.

The idea here is that this should only go to SEO visitors that are not logged in.  If they are in the desired group, we:

1) Give them the first three comments
2) If the last comment tree of those first three is cut off OR if there are more comment trees, we give them a "Read More" option
3) We show the relevance info
4) We render but hide those comment trees so they show up still for SEO purposes
5) We render and show all the remaining comment trees normally

If they click the "Read More", then we show the content normally with the relevance stuff pushed to the bottom.

src/featureflags.es6.js
```diff
@@ -21,13 +23,22 @@ const config = {
...
-        variant: 'relevancy_mweb:top',
-      }],
-    },
+    or: [
+      {
+        and: {
+          seoReferrer: true,
+          loggedin: false,
+        },
+      }, {
```
prashtx: It might be worth setting up a URL flag for enabling this, so you don't have to fake/force the referrer to test it out.

Also, if this is a true experiment, then you'll need a `variant` condition, so we'll look for the API-served bucketing info and will issue the appropriate events/etc.

uzi: Since it's all in a big `or` clause, I still have the option of the url flag in /r/gaming, maintaining the previous behavior.  Is that sufficient?

I'll pick your brain on the variant condition... I've just been told to make it occur for all SEO visitors that are not logged in.

prashtx: Yup, you can still use the URL flag to force this behavior in r/gaming, which should be OK for sharing. If you need to let folks force this on for self-post-only subs (no thubnails), then it might be worth tweaking the rules a bit, but it's also possible to just intentionally navigate via search if needed.

If this isn't a true experiment, i.e. if users aren't bucketed into two controls and various treatments, then you don't need the `variant` condition.

prashtx:
:fish:

PR #699, opened 2016-08-25
Add ReCaptcha to the register page

uzi:
A few things to fix still, primarily around the loading of the code from Google, and a few other things to chat about before the eyeglasses come out... but overall it seems to work.

### DIFF
--- a/src/app/components/LoginRegistrationForm/styles.less
+++ b/src/app/components/LoginRegistrationForm/styles.less
@@ -29,13 +29,14 @@
     margin-top: 40px;
   }
 
-  &__checkbox {
-    margin-top: 15px;
+  &__checkbox-label {
+    display: block;
+    margin: 15px 0;
   }
 
-  &__checkbox-label {
-    vertical-align: top;
-    margin-left: 4px;
+  &__checkbox {
+    margin-right: 4px;
+    vertical-align: middle;
   }
 
   &__error-text {
--- a/src/app/components/ReCaptcha/index.jsx
+++ b/src/app/components/ReCaptcha/index.jsx
@@ -1,0 +1,81 @@
+import React from 'react';
+import './styles.less';
+
+const T = React.PropTypes;
+
+// A promise to serve as a barrier to ensure that we've loaded google's bits
+const grecaptchaLoaded = new Promise(resolve => {
+  // Safeguard against server rendering
+  if (typeof window !== 'undefined') {
+    window.grecaptchaOnLoad = resolve;
+  }
+});
+
+class ReCaptcha extends React.Component {
+  static propTypes = {
+    elementId: T.string,
+    sitekey: T.string.isRequired,
+    theme: T.oneOf(['light', 'dark']),
+    type: T.oneOf(['image', 'audio']),
+    size: T.oneOf(['normal', 'compact']),
+    tabindex: T.number,
+    onSuccess: T.func,
+    onExpiration: T.func,
+  };
+
+  static defaultProps = {
+    elementId: 'g-recaptcha',
+    theme: 'light',
+    type: 'image',
+    size: 'normal',
+    tabindex: 0,
+  };
+
+  async componentDidMount() {
+    const {
+      elementId,
+      sitekey,
+      theme,
+      type,
+      size,
+      tabindex,
+      onSuccess,
+      onExpiration,
+    } = this.props;
+
+    if (!window.grecaptcha) {
+      // If it's not loaded yet, let's do that...
+      // Add the script to the DOM, causing it to load...
+      const script = document.createElement('script');
+      script.src = 'https://www.google.com/recaptcha/api.js?onload=grecaptchaOnLoad&render=explicit';
+      document.body.appendChild(script);
+
+      // ... wait for the promise to get resolved
+      await grecaptchaLoaded;
+    }
+
+    window.grecaptcha.render(elementId, {
+      sitekey,
+      theme,
+      type,
+      size,
+      tabindex,
+      callback: onSuccess,
+      'expired-callback': onExpiration,
+    });
+  }
+
+  shouldComponentUpdate() {
+    return false;
+  }
+
+  render() {
+    const { elementId } = this.props;
+
+    return (
+      <div className='ReCaptcha' id={ elementId } />
+    );
+  }
+}
+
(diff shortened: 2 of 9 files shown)


uzi:
👓  @schwers @phil303

src/app/components/LoginRegistrationForm/styles.less
```diff
@@ -31,6 +31,7 @@
 
   &__checkbox {
     margin-top: 15px;
+    margin-bottom: 15px;
```
nramadas: maybe cleaner to write `margin: 15px 0;`

src/app/components/Register/index.jsx
```diff
@@ -106,6 +115,10 @@ export default class Register extends React.Component {
     this.setState({ [fieldName]: e.target.value });
   }
 
+  ReCaptchaCallback(value) {
```
nramadas: can we rename this to `setRecaptchaResponse`?

src/app/components/Register/index.jsx
```diff
@@ -184,6 +200,7 @@ export default class Register extends React.Component {
           </LoginInput>
           <div className='Register__checkbox'>
             <input
+              id='newsletter'
```
nramadas: :no_bell: i should have caught this in the original register patch, so i don't fault you if you don't want to fix it here. But with checkboxes, it's better to wrap the checkbox with a label, so that the small margin between the checkbox and the label will also trigger the checkbox select. If you feel like, instead of using an `id` with the inputs (because, yuck `id`s), you could do this:

```es6
<label className='Register__checkbox-label'>
  <input
    type='checkbox'
    name='newsletter'
    defaultChecked
  />
  Subscribe to newsletter
</label>
```

src/app/router/handlers/Register.js
```diff
@@ -12,9 +12,10 @@ export default class Register extends BaseHandler {
   }
 
   async [METHODS.POST](dispatch/*, getState, utils*/) {
-    const { username, password, email, newsletter } = this.bodyParams;
+    const { username, password, email, newsletter, gRecaptchaResponse } = this.bodyParams;
     try {
-      const newSession = await registerUser(username, password, email, newsletter);
+      const newSession = await registerUser(username, password, email,
+                                            newsletter, gRecaptchaResponse);
```
nramadas: 🔕 what's our styleguide say for multiline function calls?

is it version A (like above):
```es6
foo(thing1,
    thing2,
    thing3);
```

or option B:
```es6
foo(
  thing1,
  thing2,
  thing3,
);
```

or option C (something else)?

src/server/templates/main.jsx
```diff
@@ -45,6 +45,7 @@ export default function(data, store) {
             __html: `window.___r = ${JSON.stringify(data)}`,
           } }
         ></script>
+        <script async src='https://www.google.com/recaptcha/api.js?render=explicit'></script>
```
nramadas: super small nit, and i mean this is a doozy of a nit: since this is a react template, you can make the tag self closing:

`<script async src='https://www.google.com/recaptcha/api.js?render=explicit'/>`

src/app/components/LoginRegistrationForm/styles.less
uzi: Good call.

src/app/components/Register/index.jsx
schwers: Oh that's pretty neat, TIL

src/app/router/handlers/Register.js
schwers: I couldn't find anything on it. Off-hand, I think I've seen this and option B

src/app/components/Register/index.jsx
uzi: Cool, no problem -- I'll change that.  Had to fiddle a bit with the styles as well, of course.

src/app/components/Register/index.jsx
uzi: Sure, will do.

src/app/router/handlers/Register.js
uzi: When you figure it out, let me know. :)

src/server/templates/main.jsx
uzi: Oh, right -- had HTML on the brain.

src/config.js
```diff
@@ -46,6 +46,9 @@ const config = () => ({
   appName: process.env.APP_NAME || 'mweb',
 
   defaultCountry: process.env.DEFAULT_COUNTRY || 'US',
+
+  // Note that this is a public key, so this can be shared.
+  recaptchaSitekey: process.env.RECAPTCHA_SITEKEY || '6LeTnxkTAAAAAN9QEuDZRpn90WwKk_R1TRW_g-JC',
```
schwers: :no_bell: This might be a good time to start noting what `process.env` vars needed at build time / on the client? It would be nice to know that for building&infra purposes, and for when we fix the client env issue.

src/app/router/handlers/Register.js
schwers: Heh, personally I think this seems good for now :)

ajacksified: I like B vOv

src/config.js
ajacksified: `/CONFIG.md` is a good place

uzi:
💇

uzi:
💇

src/app/components/ReCaptcha/index.jsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import './styles.less';
+
+const T = React.PropTypes;
+
+// A promise to serve as a barrier to ensure that we've loaded google's bits
+const grecaptchaLoaded = new Promise(resolve => {
+  window.grecaptchaOnLoad = resolve;
```
nramadas: a quick guard against when this runs on the server
```es6
if (typeof window != 'undefined') {
  window.grecaptchaOnLoad = resolve;
}
```

src/config.js
schwers: Perhaps, we should port it soon ;)

phil303:
:fish: nice work

schwers:
:fish: awesome, thanks for porting this

PR #701, opened 2016-08-25
Add warning alert for archived, locked, and commentID posts

jamesmaa:
I'm actually having trouble messing around with themeify css. Could someone teach me? 
Otherwise:
**locked**
image
**single comment**
image
**archived**
image
**night mode**
image

👓  @uzi @phil303

### DIFF
--- a/src/app/components/Comment/CommentTools/index.jsx
+++ b/src/app/components/Comment/CommentTools/index.jsx
@@ -16,6 +16,8 @@ export default function CommentTools(props) {
     score,
     scoreHidden,
     voteDirection,
+    commentingDisabled,
+    votingDisabled,
     commentAuthor,
     username,
     saved,
@@ -29,10 +31,10 @@ export default function CommentTools(props) {
 
   return (
     <div className='CommentTools'>
-      { renderReply(props) }
+      { commentingDisabled ? null : renderReply(props) }
       { renderSeashells(tooltipId) }
       { renderDivider(props) }
-      { renderVote(id, score, scoreHidden, voteDirection) }
+      { renderVote(id, score, scoreHidden, votingDisabled, voteDirection) }
       { renderDropdown(tooltipId, permalinkUrl, commentAuthor, username, saved,
                        onEdit, onDelete, onToggleSave) }
     </div>
@@ -85,12 +87,13 @@ const renderSeashells = tooltipId => (
   </TooltipTarget>
 );
 
-const renderVote = (id, score, scoreHidden, voteDirection) => (
+const renderVote = (id, score, scoreHidden, hideDownvote, voteDirection) => (
   <Vote
     thingId={ id }
     classPrefix='CommentTools'
     score={ score }
     scoreHidden={ scoreHidden }
+    hideDownvote={ hideDownvote }
     voteDirection={ voteDirection }
   />
 );
--- a/src/app/components/Comment/index.jsx
+++ b/src/app/components/Comment/index.jsx
@@ -38,11 +38,18 @@ export function Comment (props) {
 }
 
 function renderFooter(props) {
-  const { commentDeleted, commentReplying, commentCollapsed, comment, preview } = props;
+  const {
+    commentDeleted,
+    commentReplying,
+    commentingDisabled,
+    commentCollapsed,
+    comment,
+    preview,
+  } = props;
 
   return [
     !commentDeleted ? renderTools(props) : null,
-    !preview && commentReplying ? renderCommentReply(props) : null,
+    !preview && commentReplying && !commentingDisabled ? renderCommentReply(props) : null,
     !preview && !commentCollapsed && comment.replies.length ? renderReplies(props) : null,
   ];
 }
@@ -125,6 +132,8 @@ function renderTools(props) {
     onToggleEditForm,
     onDeleteComment,
     onToggleSaveComment,
+    commentingDisabled,
+    votingDisabled,
   } = props;
 
   let cls = 'Comment__toolsContainer clearfix';
@@ -147,6 +156,8 @@ function renderTools(props) {
           onEdit={ onToggleEditForm }
           onDelete={ onDeleteComment }
           onToggleSave={ onToggleSaveComment }
+          commentingDisabled={ commentingDisabled }
+          votingDisabled={ votingDisabled }
         />
       </div>
     </div>
@@ -236,6 +247,16 @@ const userSelector = state => state.user;
 const nestingLevelSelector = (state, props) => props.nestingLevel;
 const commentCollapsedSelector = (state, props) => state.collapsedComments[props.commentId];
 const currentPageSelector = (state) => state.platform.currentPage;
+const commentingDisabledSelector = (state) => {
+  const postId = `t3_${state.platform.currentPage.urlParams.postId}`;
+  const post = state.posts[postId];
+  return post ? (post.archived || post.locked) : true;
+};
+const votingDisabledSelector = (state) => {
+  const postId = `t3_${state.platform.currentPage.urlParams.postId}`;
+  const post = state.posts[postId];
+  return post ? (post.archived) : true;
+};
 const commentReplyingSelector = (state, props) =>
   state.platform.currentPage.queryParams.commentReply === props.commentId;
 const commentEditingSelector = (state, props) => state.editingComment === props.commentId;
@@ -251,6 +272,8 @@ const combineSelectors = (
   nestingLevel,
   commentCollapsed,
   currentPage,
+  commentingDisabled,
+  votingDisabled,
   commentReplying,
   isEditing,
   highlightedComment,
@@ -263,6 +286,8 @@ const combineSelectors = (
   nestingLevel,
   commentCollapsed,
   currentPage,
+  commentingDisabled,
+  votingDisabled,
   commentReplying,
   isEditing,
   highlightedComment,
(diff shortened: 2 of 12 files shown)


src/app/components/CommentsPage/CommentsPageTools/index.jsx
```diff
@@ -8,8 +8,44 @@ import CommentReplyForm from 'app/components/Comment/CommentReplyForm';
 import SortSelector from 'app/components/SortSelector';
 import { SORTS } from 'app/sortValues';
 
+const renderThreadNotice = (post, hasSingleComment) => {
+  const postLoaded = !!post;
+  if (!postLoaded) {
+    return null;
+  }
```
uzi: 🔕 Could also just go with `if (!post)` since postLoaded isn't used otherwise... but it's ok if you think it helps readability.

src/app/components/CommentsPage/CommentsPageTools/index.jsx
```diff
@@ -8,8 +8,44 @@ import CommentReplyForm from 'app/components/Comment/CommentReplyForm';
...
+        </Anchor>
+      </p>
+    );
+  } else if (post.archived) {
+    message = 'Post is archived';
+  } else if (post.locked) {
+    message = <p> Comments are locked </p>;
+  } else if (subredditDetail && subredditDetail.user_is_banned) {
+    message = <p> You are banned from commenting in this community for now </p>;
+  }
```
uzi: Hmm... some messages are in <p> and some are not... is there a reason for the disparity?

src/app/components/CommentsPage/CommentsPageTools/index.jsx
```diff
@@ -8,8 +8,44 @@ import CommentReplyForm from 'app/components/Comment/CommentReplyForm';
...
 
+const renderThreadNotice = (post, hasSingleComment) => {
+  const postLoaded = !!post;
+  if (!postLoaded) {
+    return null;
+  }
 
-export default ({ replying, currentPage, id, onSortChange }) => {
+  let message;
+  const subredditDetail = post.subredditDetail;
```
uzi: Could do `const { subredditDetail } = post;` instead.

uzi:
💅 Nice!  Three birds, meet one stone!  Just a couple of nits, but this looks good!

src/app/components/CommentsPage/CommentsPageTools/index.jsx
jamesmaa: oops late night blunder

jamesmaa:
💇 Thanks for reviewing!

src/app/components/CommentsPage/CommentsPageTools/index.jsx
```diff
@@ -8,8 +8,43 @@ import CommentReplyForm from 'app/components/Comment/CommentReplyForm';
 import SortSelector from 'app/components/SortSelector';
 import { SORTS } from 'app/sortValues';
 
+const renderThreadNotice = (post, hasSingleComment) => {
+  if (!post) {
```
phil303: Where possible, one thing that I've found that makes jsx a bit clearer is to move these types of checks into the parent render function. So for instance with this particular one, the code below (~line 84) would look something like:

```jsx
{ post ? renderThreadNotice(post, hasSingleComment) : null }
```

This has the nice effect of letting future readers avoid diving down into functions unnecessarily and making it very apparent what this element requires in order to be visible.

src/app/components/CommentsPage/CommentsPageTools/styles.less
```diff
@@ -2,7 +2,35 @@
 @import (reference) '~app/less/mixins/importAll';
 @import (reference) '~app/less/themes/themeify';
 
+.themed-alert(@base-color, @background-color, @text-color) {
+  .themeify(
```
phil303: You're pretty close here.  We actually have color constants for both day mode and night mode defined in separate files. Since yours look new and don't yet exist, you'll have to define them in both `app/less/themes/dayMode.less` and its counterpart `nightMode.less`. When that's done, your code will just "do the right thing" when in either mode if it looks something like this:

```less
.themeify({
  border-color: @theme-warning-border;
  background-color: @theme-warning-body-color;
  color: @theme-warning-body-text-color;
});
```

phil303: One thing to be careful of - we do have a color style guide [here](https://www.dropbox.com/s/f9pc60gj1bkmd3e/reddit_brandbook.pdf?dl=0) (page 17). Since some of these colors are not part of it, it might be good to get some quick input from Jonathan or Derek. They might have some good feedback on the color choices.

src/app/components/CommentsPage/CommentsPageTools/index.jsx
```diff
@@ -8,8 +8,43 @@ import CommentReplyForm from 'app/components/Comment/CommentReplyForm';
...
+
+  if (message) {
+    return (
+      <div className='alert-warning'>
+        { message }
+      </div>
+    );
+  }
+
+  return null;
```
phil303: nit, returning null isn't really necessary in this case

phil303:
:nail_care:

src/app/components/CommentsPage/CommentsPageTools/styles.less
jamesmaa: I copied the styles from 1X:
https://github.com/reddit/reddit-mobile/blob/master/assets/less/components/alerts.less

phil303: Ah interesting. If you don't mind, it'd still be good to do a quick check in with them. This might've been before the style guide was introduced and would be a good opportunity to make sure we're aligned.

jamesmaa:
💇 
Re styles: the styles are consistent with 1X. Is that good enough or should I ask design?

jamesmaa:
UPDATE: So I talked to design and they are okay with the color scheme. From a UX perspective though they want us to follow the UX of the mobile apps, mainly
1. When a post is archived, the warning box doesn't show. They don't think it's useful information.
2. They want the upvote/downvote/reply buttons to be removed if a post is archived or locked.

I don't know if that is within the scope of this commit. Thoughts?

jamesmaa:
UPDATES
- Moved all styles to `dayMode.less` and `nightMode.less`
- Disabled voting and removed downvote on archived posts
- Removed "Post is archived" message per design team's request

Screenshots:
**archived**
image
image

**locked**
image
image

**single comments**
image

💇

src/app/components/Comment/CommentTools/index.jsx
```diff
@@ -85,12 +87,13 @@ const renderSeashells = tooltipId => (
...
 );
 
-const renderVote = (id, score, scoreHidden, voteDirection) => (
+const renderVote = (id, score, scoreHidden, hideDownvote, voteDirection) => (
   <Vote
     thingId={ id }
     classPrefix='CommentTools'
     score={ score }
     scoreHidden={ scoreHidden }
+    hideDownvote={ hideDownvote }
```
uzi: Hmm... it's votingDisabled in commentTools, but hideDownvote here?  Why only hide the downvote?  Seems we'd want to hide both and only show the score, right?  That's what we did on 1X at least, iirc.

src/app/components/Comment/index.jsx
```diff
@@ -236,6 +247,16 @@ const userSelector = state => state.user;
...
+const commentingDisabledSelector = (state) => {
+  const postId = `t3_${state.platform.currentPage.urlParams.postId}`;
+  const post = state.posts[postId];
+  return post ? (post.archived || post.locked) : true;
+};
+const votingDisabledSelector = (state) => {
+  const postId = `t3_${state.platform.currentPage.urlParams.postId}`;
+  const post = state.posts[postId];
+  return post ? (post.archived) : true;
+};
```
uzi: You said "They want the upvote/downvote/reply buttons to be removed if a post is archived or locked" ... so then commentingDisabledSelector and votingDisabledSelector should be the same.  Should we consolidate?

src/app/components/Vote/index.jsx
```diff
@@ -31,7 +31,7 @@ export function renderUpvote (props) {
       className={ `Vote__form ${classPrefix}__Vote__form ${voteCls}` }
     >
       <input type='hidden' value={ formDirection } name='direction' />
-      <button className={ cls } type='submit' />
+      <button className={ cls } type='submit' disabled={ hideDownvote } />
```
uzi: Yep, just confused about hiding only downvoting when I don't think they'll be able to vote at all.

uzi:
💅 Looking great, just want to double-check about the voting situation. Thanks!

src/app/components/Comment/CommentTools/index.jsx
jamesmaa: oh native-mobile, they show the upvote to indicate the score.
image

src/app/components/Comment/index.jsx
jamesmaa: A locked post should be able to upvote/downvote but not comment right? That's why I had two separate variables

src/app/components/Vote/index.jsx
jamesmaa: the logic is that when you hide the downvote, the renderUpvote button should render a disabled upvote

jamesmaa:
💇 The voting situation is that through talking with design, they want the UX to align with the native mobile experience, so they still want an upvote next to the score

uzi:
:fish: I double checked on desktop and you're right about the locked post.  Thanks again for the help!

PR #702, opened 2016-08-27
Load Recaptcha in the component, guarded by a promise

uzi:
Bring the ReCaptcha script loading into the component, protected by a promise.

👓 @schwers @phil303


schwers:
:fish: :party: 🐠

PR #703, opened 2016-08-29
Make page no. visible in URL bar.

alsuren:
This makes it easier to glance up and see how much time you've
wasted browsing /r/all. The desktop site has an incrementing count
as its first URL argument too, and I miss it when I'm on mobile.


curioussavage:
looks like this was implemented on our old app, the new one is on the 2X branch. If you implement it there Ill add it.

alsuren:
It turns out the param is unused. Here's a PR to remove it from 2X, so I can close this PR: https://github.com/reddit/reddit-mobile/pull/956

PR #705, opened 2016-08-29
Allow users to login in with reddit.com session cookies

nramadas:
This is largely just copied over from what works in modmail. Essentially, it allows users that have a reddit_session cookie (set by reddit.com) to automatically log in to 2X.

:eyeglasses: @phil303 @schwers @uzi

### DIFF
--- a/src/server/session/dispatchSession.js
+++ b/src/server/session/dispatchSession.js
@@ -10,36 +10,37 @@ export default async (ctx, dispatch, apiOptions) => {
   // try to create a session from the existing cookie
   // if the session is malformed somehow, the catch will trigger when trying
   // to access it
-  const tokenCookie = ctx.cookies.get('token');
-  const redditSessionExists = ctx.cookies.get('reddit_session');
-  if (!(tokenCookie || redditSessionExists)) {
-    return;
-  }
+  const token = ctx.cookies.get('token');
+  const redditSession = ctx.cookies.get('reddit_session');
 
   let session;
   let sessionData;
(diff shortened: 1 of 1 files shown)


src/server/session/dispatchSession.js
```diff
@@ -10,36 +10,37 @@ export default async (ctx, dispatch, apiOptions) => {
...
+    setSessionCookies(ctx, session);
   }
 
   // if the session is invalid, try to use the refresh token to grab a new
   // session.
-  if (!session.isValid) {
+  if (session && sessionData && !session.isValid) {
     const data = await PrivateAPI.refreshToken(apiOptions, sessionData.refreshToken);
-    session = makeSessionFromData({ ...data, refresh_token: sessionData.refreshToken });
+    session = makeSessionFromData({ ...data, refreshToken: sessionData.refreshToken });
```
schwers: Should we put `sessionData.refreshToken` first so `data` can override with a new one?

schwers:
💅  Just a small question, looks good otherwise

src/server/session/dispatchSession.js
nramadas: that's a good point. also, i realized this line is wrong. it should be camel_cased, because that's what the function expects.

nramadas:
:haircut:

schwers:
:fish:

schwers:
@nramadas I was wrong, this was merged but the PR wasn't closed.

PR #706, opened 2016-08-29
Grab LOIDs from the server

nramadas:
Blocked by node-api-client PR here: https://github.com/reddit/node-api-client/pull/150

This patch pulls in the loids created by r2 api into state. It also makes sure that the "user fetch" call is highly blocking so that we can be sure `loid`s are in state before any further fetches happen (in case we need to add tracking for any subsequent api calls we make).


src/app/router/handlers/CommentsPage.js
```diff
@@ -45,11 +45,12 @@ export default class CommentsPage extends BaseHandler {
     const state = getState();
     if (state.platform.shell) { return; }
 
+    // block on user data to make sure loid gets loaded in
+    await fetchUserBasedData(dispatch);
```
nramadas: this is highly repetitive, and i'm open to suggestions on how to clean this up. i noticed we're doing a similar thing with checking for 'shell rendering' right above this. a cleanup might not be appropriate for an initial launch, but we should definitely review this chunk of boilerplate

uzi: At least it's a one-liner ... in python land, it would probably be a decorator.

nramadas: true, but even with a decorator i feel really un easy. it's not hard for someone else to try to create a new handler, and forget to add this blocking call. Or, worse yet, they see that all the handlers include it, and simply copy the above lines into the new handler without really understanding what they are for.

nramadas:
:haircut:

src/server/initialState/permanentCookieOptions.js
```diff
@@ -2,6 +2,7 @@ const COOKIE_OPTIONS = {
   httpOnly: false,
   overwrite: true,
   maxAge: 1000 * 60 * 60,
+  domain: process.env.NODE_ENV === 'production' ? '.reddit.com' : null,
```
phil303: should this go into config.js?

src/server/initialState/dispatchInitialUser.js
```diff
@@ -0,0 +1,29 @@
...
+import * as loidActions from 'app/actions/loid';
+import { permanentCookieOptions } from './permanentCookieOptions';
+
+export const dispatchInitialUser = async (ctx, dispatch, getState) => {
+  // first, if we already have loid cookies, toss those into state.
+  // the lack of camel casing on the 'loidcreated' cookie name is intentional.
+  const oldLoid = ctx.cookies.get('loid');
+  const oldLoidCreated = ctx.cookies.get('loidcreated');
+
+  if (oldLoid || oldLoidCreated) {
```
phil303: This combined with the [string check](https://github.com/reddit/reddit-mobile/pull/706/files#diff-a52797dd8284198695b7649f5b8c21bbR12) in the reducer feels like an odd check. It seems like you definitely need `oldLoidCreated` to exist right? Your reducer handles the undefined `oldLoid` case but doesn't handle the undefined `oldLoidCreated` case.

src/server/initialState/dispatchInitialUser.js
```diff
@@ -0,0 +1,29 @@
...
+    dispatch(loidActions.setLOID(oldLoid, oldLoidCreated));
+  }
+
+  // fetchMyUser pulls in the loid and loidCreated fields
+  await dispatch(userActions.fetchMyUser());
+  const { loid, loidCreated } = await getState().loid;
```
phil303: is `getState().loid` an object with keys `loid` and `loidCreated`?

src/lib/apiOptionsFromState.js
```diff
@@ -1,12 +1,26 @@
 import APIOptions, { optionsWithAuth } from '@r/api-client';
+import merge from '@r/platform/merge';
 
 export const apiOptionsFromState = (state) => {
   if (!state) { return APIOptions; }
 
-  const { session } = state;
+  let options = { ...APIOptions };
```
phil303: Just to get across that `options` can only start in one of two states, what do you think of:
```js
const { session, meta } = state;
const options = (session && session.accessToken) 
  ? optionsWithAuth(session.accessToken) 
  : { ...APIOptions };
```

src/lib/apiOptionsFromState.js
```diff
@@ -1,12 +1,26 @@
...
+  const { session, meta } = state;
+
   if (session && session.accessToken) {
-    return optionsWithAuth(session.accessToken);
+    options = optionsWithAuth(session.accessToken);
+  }
+
+  // grab loids if we have them, and set the cookie if on the server
+  const { loid, loidCreated } = state.loid;
```
phil303: same comment here as in the other case of this

src/server/initialState/permanentCookieOptions.js
nramadas: potentially, but i think i'm going to use whatever schwers does with his cookies patch and set this cookie the same way. thanks for pointing it out though, it'll serve as a reminder to switch it over when the time comes

src/Server.js
```diff
@@ -82,16 +82,16 @@ export function startServer() {
     template: main,
     reducers,
     reduxMiddleware,
-    dispatchBeforeNavigation: async (ctx, dispatch/*, getState, utils*/) => {
+    dispatchBeforeNavigation: async (ctx, dispatch, getState/*, utils*/) => {
```
phil303: nit (and I know it might not be your code) but i dislike that we do the commenting out thing for remaining args. If our docs aren't making this obvious enough (which they might not be) then we should update the docs.

src/server/initialState/dispatchInitialUser.js
nramadas: good point. i'll have the reducer handle the undefined `oldLoidCreated` case

phil303:
:nail_care:

nramadas:
:haircut:

src/server/initialState/permanentCookieOptions.js
schwers: so I'm doing https://github.com/reddit/reddit-mobile/pull/713/files#diff-53d96610cf9c22a0b604e7c888acf5f3R9, and it lives outside of `permanentCookieOptions`. My thought was not all cookies are going to want to live on `.reddit.com`. If we like that, I'd be happy to make a separate, `rootPermanentCookieOptions.js` that does this automatically for you so we can more easily share this.

src/server/session/setSessionCookies.js
```diff
@@ -3,6 +3,7 @@ const COOKIE_OPTIONS = {
   httpOnly: false,
   overwrite: true,
   maxAge: 1000 * 60 * 60,
+  domain: process.env.NODE_ENV === 'production' ? '.reddit.com' : null,
```
schwers: 🔕  Make sure we merge this with #713  properly

src/lib/apiOptionsFromState.js
```diff
@@ -1,12 +1,23 @@
...
+  const options = (session && session.accessToken)
+    ? optionsWithAuth(session.accessToken)
+    : { ...APIOptions };
+
+  // grab loids if we have them, and set the cookie if on the server
+  const { loid: { loid, loidCreated }, meta } = state;
+  if (loid && meta.env !== 'CLIENT') {
+    return merge(options, {
+      headers: {
+        'Cookie': `loid=${loid}; loidcreated=${loidCreated}`,
```
schwers: Do we intentionally put loid on state, and pass it, when the users is logged in?

src/Server.js
```diff
@@ -82,16 +82,16 @@ export function startServer() {
...
       dispatchInitialShell(ctx, dispatch);
-      dispatchInitialLoid(ctx, dispatch);
       await dispatchSession(ctx, dispatch, ConfigedAPIOptions);
       dispatchInitialTheme(ctx, dispatch);
       dispatchInitialCollapsedComments(ctx, dispatch);
       dispatchInitialCompact(ctx, dispatch);
       dispatchInitialMeta(ctx, dispatch);
       dispatchInitialOver18(ctx, dispatch);
       dispatchInitialRecentSubreddits(ctx, dispatch);
+      dispatchInitialUser(ctx, dispatch, getState);
```
schwers: Do we need an `await` here? I remember having weird issues with `dispatchSession`. The platform bug with `dispatchBeforeNavigation` might have been my real issue though.

schwers:
💅

src/lib/apiOptionsFromState.js
nramadas: Since on the client, any loid cookies that exist will be sent with the request, i've decided to leave this as is for the server case as well, just to be consistent.

nramadas:
:haircut:

src/app/actions/user.js
```diff
@@ -2,5 +2,5 @@ import * as accountActions from 'app/actions/accounts';
 
 export const fetchMyUser = () => async (dispatch, getState) => {
   const state = getState();
-  dispatch(accountActions.fetch({ name: 'me', loggedOut: !state.session.accessToken }));
+  await dispatch(accountActions.fetch({ name: 'me', loggedOut: !state.session.accessToken }));
```
phil303: can you leave a comment on what's happening here?

phil303:
:fish: small comment and a reminder to watch out for [PR713](https://github.com/reddit/reddit-mobile/pull/706/files#r77932027)

src/app/actions/user.js
schwers: Upovte for comment, it'd be nice to know that this blocks other data fetches for consistent loid / logging

schwers: Actually  wait, there's a comment in `dispatchInitialUser` which is more important. disregard :)

schwers:
:fish: but lets talk about merging with #713 before you merge

nramadas:
:haircut: wanna take a quick look since i'm setting cookies on `.reddit.com` now?

phil303:
:fish:

schwers:
:fish:

PR #707, opened 2016-08-29
Interest targeting fix 1 x

jamesmaa:
There was an error occurring on the server side. 

My understanding is that `indexPage` in routes.jsx is executed both on the client and server side. This comes from the following screenshot in which I put debugging code in the method and found the statement printed both in my browser and terminal
This was what was occurring in production: I was getting a bunch of server errors but was confused as to why the cookie set anyways. The above explanation is my best hypotheses as why that is occurring, although I have not been able to locate the reason why routes.jsx runs both on client side and server side

**Solution**
I check to make sure cookies are enabled. This lets the client running the code to set the cookie, while when the server runs it, because there are no cookie or localstorage functionality available, the method becomes a silent no-op instead of an error. 

The first commit is the original commit. The second commit is the fix.

*Note*:  I missed the error message the first time because I thought it was noise coming from another part of the source code. Realizing it's a bad idea now

👓 @nramadas

### DIFF
--- a/src/constants.es6.js
+++ b/src/constants.es6.js
@@ -44,6 +44,12 @@ export default {
   VISITED_POST_COUNT: 20,
   VISITED_POSTS_KEY: 'visitedPosts',
 
+  RECENT_SUBREDDIT_COUNT: 10,
+  RECENT_SUBREDDITS_KEY: 'recent_srs',
+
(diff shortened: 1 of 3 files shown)


nramadas:
:fish:

PR #708, opened 2016-08-30
More comments

phil303:
It's open season baby! Feel free to review this in a more granular way - some of it will be subject to change after our live review.

### DIFF
--- a/src/app/actions/comment.js
+++ b/src/app/actions/comment.js
@@ -1,6 +1,9 @@
-import { endpoints, models } from '@r/api-client';
+import { endpoints, models, errors } from '@r/api-client';
 
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+
+const { SavedEndpoint, CommentsEndpoint } = endpoints;
+
 
 export const TOGGLE_EDIT_FORM = 'COMMENT__TOGGLE_EDIT_FORM';
 export const toggleEditForm = id => ({ type: TOGGLE_EDIT_FORM, id });
@@ -31,7 +34,7 @@ export const toggleSave = id => async (dispatch, getState) => {
   const state = getState();
   const comment = state.comments[id];
   const method = comment.saved ? 'del' : 'post';
-  await endpoints.SavedEndpoint[method](apiOptionsFromState(state), { id });
+  await SavedEndpoint[method](apiOptionsFromState(state), { id });
   // the response doesn't have anything in it (yay api), so emit a new model
   // on the client side;
   const newComment = models.CommentModel.fromJSON({ ...comment.toJSON(), saved: !comment.saved });
@@ -42,7 +45,7 @@ export const del = id => async (dispatch, getState) => {
   const state = getState();
   const comment = state.comments[id];
   const apiOptions = apiOptionsFromState(state);
-  await endpoints.CommentsEndpoint.del(apiOptions, id);
+  await CommentsEndpoint.del(apiOptions, id);
   // the response doesn't have anything in it, so we're going to guess what the
   // comment should look like
   const newComment = models.CommentModel.fromJSON({
@@ -59,8 +62,61 @@ export const report = (id, reason) => async (dispatch, getState) => {
   console.log(id, reason, getState());
 };
 
-export const LOADING_MORE = 'LOADING_MORE';
-export const LOADED_MORE = 'LOADED_MORE';
-export const loadMore = ids => async (dispatch, getState) => {
-  console.log(ids, getState());
+export const MORE_COMMENTS_FETCHING = 'COMMENTS__MORE_FETCHING';
+export const MORE_COMMENTS_RECEIVED = 'COMMENTS__MORE_RECEIVED';
+export const MORE_COMMENTS_FAILURE = 'COMMENTS__MORE_FAILURE';
+
+export const fetching = parentCommentId => ({ parentCommentId, type: MORE_COMMENTS_FETCHING });
+export const received = (parentCommentId, comments) => {
+  return {
+    comments,
+    parentCommentId,
+    type: MORE_COMMENTS_RECEIVED,
+  };
+};
+
+export const failure = (parentCommentId, error) => {
+  return {
+    error,
+    parentCommentId,
+    type: MORE_COMMENTS_FAILURE,
+  };
+};
+
+export const loadMore = comment => {
+  return async (dispatch, getState) => {
+    dispatch(fetching(comment.uuid));
+
+    try {
+      const resp = await CommentsEndpoint.get(apiOptionsFromState(getState()), {
+        linkId: comment.linkId,
+        commentIds: comment.loadMoreIds,
+        sort: 'confidence',
+      });
+
+      // Doing this in here since doing it in api-client is messy. Since we
+      // need to mix the current comment replies with the new, the `get` api
+      // would need to accept the comment. This is tricky since that function
+      // actually covers multiple endpoints and would need special handling of
+      // the comment object.
+      const newComment = comment.set({
+        replies: [ ...comment.replies, ...resp.results ],
+        loadMoreIds: [],
+        loadMore: false,
+      });
+
+      dispatch(received(
+        comment.uuid,
+        { ...resp.comments, [newComment.uuid]: newComment },
+      ));
+
+    } catch (e) {
+      if (e instanceof errors.ResponseError) {
+        dispatch(failure(e));
+      } else {
+        throw e;
+      }
+    }
+
+  };
 };
--- a/src/app/components/Comment/index.jsx
+++ b/src/app/components/Comment/index.jsx
@@ -6,36 +6,70 @@ import { createSelector } from 'reselect';
 import { models } from '@r/api-client';
 
 import mobilify from 'lib/mobilify';
-
 import * as commentActions from 'app/actions/comment';
-
+import { DEFAULT_COMMENT_REQUEST } from 'app/reducers/moreCommentsRequests';
+import cx from 'lib/classNames';
 import CommentsList from 'app/components/CommentsList';
-
 import CommentHeader from './CommentHeader';
 import CommentTools from './CommentTools';
 import CommentReplyForm from './CommentReplyForm';
 import RedditLinkHijacker from 'app/components/RedditLinkHijacker';
 
-// import CommentSeeMore from './CommentSeeMore';
-// import CommentEditForm from './CommentEditForm';
 
 const T = React.PropTypes;
-const { CommentModel } = models;
 
-const collapsible = props => !(props.preview || props.userActivityPage);
-const collapsed = props => props.commentCollapsed && collapsible(props);
+const LOAD_MORE_COMMENTS = 'More Comments';
+const LOADING_MORE_COMMENTS = 'Loading...';
+const NESTING_STOP_LEVEL = 6;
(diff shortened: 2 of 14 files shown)


src/app/reducers/moreCommentsRequests.js
```diff
@@ -0,0 +1,25 @@
...
+import * as commentActions from 'app/actions/comment';
+
+
+const DEFAULT = {};
+
+export default (state=DEFAULT, action={}) => {
+  switch (action.type) {
+    case commentActions.MORE_COMMENTS_FETCHING: {
+      return merge(state, {
+        [action.parentCommentId]: { loading: true, loaded: false },
```
phil303: You could technically get away with a schema like:
```js
{
   [parentCommentId]: false|true
}
```
since a component checking here would have three states available to it. `undefined` being not loading/loaded. Defined as `true` as "loading" and defined as `false` as "loaded". I didn't like the subtle nature of that structure so I defined it as it is above.

src/app/reducers/moreCommentsRequests.js
```diff
@@ -0,0 +1,25 @@
...
+      });
+    }
+
+    case commentActions.MORE_COMMENTS_RECEIVED: {
+      return merge(state, {
+        [action.parentCommentId]: { loading: false, loaded: true },
+      });
+    }
+
```
phil303: TODO: error state and tests

src/app/components/Comment/index.jsx
```diff
@@ -192,135 +145,149 @@ function renderCommentReply(props) {
...
-  postCreated: T.number,
-  comment: T.instanceOf(CommentModel),
-  user: T.object,
+  user: T.object.isRequired,
+  currentPage: T.object.isRequired,
+  comment: T.instanceOf(models.CommentModel).isRequired,
+  moreCommentStatus: T.object.isRequired,
+  commentCollapsed: T.bool.isRequired,
+  commentReplying: T.bool.isRequired,
+  isEditing: T.bool.isRequired,
```
phil303: maybe alphabetize within groups

src/app/components/Comment/index.jsx
```diff
@@ -192,135 +145,149 @@ function renderCommentReply(props) {
...
-      votingDisabledSelector,
-      commentReplyingSelector,
-      commentEditingSelector,
-      highlightedCommentSelector,
-    ],
-    combineSelectors,
-  );
-};
+const selector = createSelector(
+  state => state.user,
```
phil303: maybe add a comment here about what this is

phil303: common selector somewhere?

src/app/reducers/moreCommentsRequests.js
phil303: Is there a case where a parentComment has more than one "load more"?

src/app/reducers/comments.js
```diff
@@ -36,6 +36,20 @@ export default function(state=DEFAULT, action={}) {
...
+    case commentActions.MORE_COMMENTS_RECEIVED: {
+      const { parentCommentId, apiResponse: { comments, results } } = action;
+
+      const parentComment = state[parentCommentId];
+      if (!parentComment) {
+        return state;
+      }
+
+      const newReplies = [ ...parentComment.replies, ...results ];
+      const newParent = parentComment.set('replies', newReplies);
```
phil303: consider emptying `loadMoreIds` and deriving the "loaded" state.

phil303:
look at api-client for "continuation thread" notion

src/app/actions/comment.js
```diff
@@ -59,8 +62,35 @@ export const report = (id, reason) => async (dispatch, getState) => {
...
+    try {
+      const resp = await CommentsEndpoint.get(apiOptionsFromState(getState()), {
+        linkId: linkId,
+        commentIds: loadMoreIds,
+        sort: 'confidence',
+      });
+
+      dispatch(received(uuid, resp));
+    } catch (e) {
+      dispatch(failure(e));
```
phil303: look into possible distinguishing between error types

src/app/reducers/moreCommentsRequests.js
phil303: followup: spot checking desktop, I haven't found a place where this is the case. If it _is_ possible, it's an edge case we can address at a later point.

src/app/components/Comment/index.jsx
phil303: followup to the comments idea: I tried this out, I think it's actually making the coder harder to parse since there's now an intermingling of comments every other line.

phil303:
:eyeglasses: @nramadas @schwers @uzi 

I believe I've addressed all comments so this is ready for some eyes.

src/app/actions/comment.js
```diff
@@ -59,8 +62,57 @@ export const report = (id, reason) => async (dispatch, getState) => {
...
+export const MORE_COMMENTS_FAILURE = 'COMMENTS__MORE_FAILURE';
+
+export const fetching = parentCommentId => ({ parentCommentId, type: MORE_COMMENTS_FETCHING });
+export const received = (parentCommentId, comments) => {
+  return {
+    comments,
+    parentCommentId,
+    type: MORE_COMMENTS_RECEIVED,
+  };
+};
```
uzi: :no_bell: Any reason why these are formatted differently?

src/app/actions/comment.js
```diff
@@ -59,8 +62,57 @@ export const report = (id, reason) => async (dispatch, getState) => {
...
+        commentIds: comment.loadMoreIds,
+        sort: 'confidence',
+      });
+
+      // doing this in here since doing it in api-client would be a bit hairy
+      const newComment = comment.set({
+        replies: [ ...comment.replies, ...resp.results ],
+        loadMoreIds: [],
+        loadMore: false,
+      });
```
uzi: Elaborate in the comment what "this" is and/or why it'd be hairy to do in api-client?

src/app/components/Comment/index.jsx
```diff
@@ -6,37 +6,71 @@ import { createSelector } from 'reselect';
...
+          onToggleCollapse={ onToggleCollapse }
+        />
+      </div>
+
+      <RedditLinkHijacker>
+        <div
+          className={ bodyClasses }
+          dangerouslySetInnerHTML={ { __html: mobilify(props.comment.bodyHTML) } }
+        />
+      </RedditLinkHijacker>
```
uzi: :no_bell: So the bits here were previously in separate render methods ... this `RedditLinkHijacker` part in a `renderBody()` that would vanish versus `renderEditForm()` if `isEditing` was set.  Seems to me that when we reintroduce editing that this is potentially going to get split out again.  Not really a problem, just an observation.

uzi:
💅 Looks good from where I sit. Just a couple of nits.

src/app/actions/comment.js
phil303: just line length issues

src/app/components/Comment/index.jsx
```diff
@@ -192,135 +144,149 @@ function renderCommentReply(props) {
...
+
+  const loadingText = moreCommentStatus.loading
+    ? LOADING_MORE_COMMENTS
+    : `${LOAD_MORE_COMMENTS} (${comment.numReplies})`;
+
+  return (
+    <div className='Comment__loadMore' onClick={ onLoadMore }>
+      <div className='icon icon-caron-circled' />
+      <a href={ comment.permalink }>{ loadingText }</a>
     </div>
```
uzi: I know this bit is copied code, but that `<a>` tag is technically incorrect here.  The click will be swallowed by the onClick for the surrounding `<div>` and we use `<Anchor>` in 2X anyways.

The other thing is that we're missing the dots before the caron like we have in 1X.

src/app/actions/comment.js
```diff
@@ -59,8 +62,57 @@ export const report = (id, reason) => async (dispatch, getState) => {
...
+    try {
+      const resp = await CommentsEndpoint.get(apiOptionsFromState(getState()), {
+        linkId: comment.linkId,
+        commentIds: comment.loadMoreIds,
+        sort: 'confidence',
+      });
+
+      // doing this in here since doing it in api-client would be a bit hairy
+      const newComment = comment.set({
+        replies: [ ...comment.replies, ...resp.results ],
```
nramadas: nit, but this might look cleaner if it was written:
```es6
replies: comment.replies.concat(resp.results),
```

src/app/components/Comment/index.jsx
```diff
@@ -6,37 +6,71 @@ import { createSelector } from 'reselect';
...
+    'm-hidden': commentCollapsed && !isUserActivityPage,
+  });
 
   return (
-    <div className={ `Comment ${!preview ? 'in-comment-tree' : ''}` }>
-      { renderHeader(props) }
-      { isEditing ? renderEditForm(props) : renderBody(props) }
-      { !userActivityPage ? renderFooter(props) : null }
+    <div className={ cx('Comment', { 'in-comment-tree': !preview }) }>
+
```
nramadas: nit, extra newline

src/app/components/Comment/index.jsx
```diff
@@ -6,37 +6,71 @@ import { createSelector } from 'reselect';
...
+        <CommentHeader
+          id={ comment.id }
+          author={ comment.author }
+          authorType={ authorType }
+          topLevel={ nestingLevel === 0 && isCollapsible }
+          dots={ Math.max(nestingLevel - 6, 0) }
+          flair={ comment.authorFlairText }
+          created={ comment.createdUTC }
+          gildCount={ comment.gilded }
+          collapsed={ commentCollapsed && isCollapsible }
```
nramadas: 🔕  We don't need to change this now, but this does feel weird that a comment that _isn't_ collapsable could still have `commentCollapsed` be `true`.

src/app/components/Comment/index.jsx
```diff
@@ -192,135 +144,149 @@ function renderCommentReply(props) {
...
-  reportComment: (reason) => dispatch(commentActions.report(commentId, reason)),
-  loadMore: (ids) => dispatch(commentActions.loadMore(ids)),
+  reportComment: reason => dispatch(commentActions.report(commentId, reason)),
+  onLoadMore: comment => dispatch(commentActions.loadMore(comment)),
+  onToggleCollapse: commentCollapsed => dispatch(commentActions.toggleCollapse(commentId, !commentCollapsed)),
 });
 
-export default connect(makeConnectedCommentSelector, mapDispatchToProps)(Comment);
+
+function _determineAuthorType(comment, user, op) {
```
nramadas: super nit, but i'm not a huge fan of using underscores in function names unless it's a private method on an instance. for example, i'm not sure how much more "private" this function is compared to `mergeProps` or `mapDispatchToProps`; both are "privatized" by the module scope.

nramadas:
:nail_care:

src/app/actions/comment.js
uzi: Hmm... I thought it looked cleaner as is.

src/app/components/Comment/index.jsx
phil303: oh i think i agree. good catch. if it's not too much trouble i'll try to change it in this patch

nramadas:
LGTM :fish:

PR #709, opened 2016-08-30
Set the token, loid, and loid_created to the root domain.

telaviv:
At a high level the reason this PR exists is that `amp.reddit.com` will need to know if users already have a `loid` or `token` cookie so that we don't double count users that visit mweb and amp. The reason this is not possible is that currently the domain of these cookies are set to `m.reddit.com`. This means those cookies aren't readable by `amp.reddit.com`. We need them to be set to the root level `.reddit.com` to count correctly on amp.

In order to guarantee that we create new root level cookies we have to delete the old ones. The reason is that ideally we would just do this:

```
if (isNotOnRootLevelDomain(ctx.cookies.get('loid'))) {
    ctx.cookies.set('loid');
}
```

Unfortunately there's no way to tell what domain the cookie is on. So we have to simply delete the non-root cookies so that we can then regenerate them correctly.   

You'll note that this change will globally log out all users and generate new loids for existing users. Luckily we aren't live yet :D

👓 : @schwers @prashtx

### DIFF
--- a/src/server/initialState/clearDeprecatedCookies.js
+++ b/src/server/initialState/clearDeprecatedCookies.js
@@ -1,0 +1,10 @@
+export const clearDeprecatedCookies = (ctx) => {
+  // these cookies are deprecated on non root domains.
+  ctx.cookies.set('loid');
+  ctx.cookies.set('loidcreated');
+  ctx.cookies.set('token');
+
+  ctx.cookies.set('tokenExpires');
+  ctx.cookies.set('refreshToken');
+  ctx.cookies.set('reddit_session');
+};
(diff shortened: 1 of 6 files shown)


src/server/initialState/clearDeprecatedCookies.js
```diff
@@ -0,0 +1,10 @@
+export const clearDeprecatedCookies = (ctx) => {
+  // these cookies are deprecated on non root domains.
+  ctx.cookies.set('loid');
+  ctx.cookies.set('loidcreated');
+  ctx.cookies.set('token');
+
+  ctx.cookies.set('tokenExpires');
+  ctx.cookies.set('refreshToken');
+  ctx.cookies.set('reddit_session');
```
schwers: I think this will clear your `.reddit.com` reddit_session cookie

schwers:
Hi @telaviv, are we okay to close this now that we have https://github.com/reddit/reddit-mobile/pull/713 and https://github.com/reddit/reddit-mobile/pull/712?

schwers:
Note for loid there's also https://github.com/reddit/reddit-mobile/pull/706 and some 1X work will be done  as well.

PR #710, opened 2016-08-31
Add statsd, and request and response error code stats

uzi:
We log stats to graphite via statsd -- so add that
packages along with its base configuration.

Initial stats are counting the number of requests
as well as the response error codes.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -98,6 +98,7 @@
     "curly": 2,
     "dot-location": [2, 'property'],
     "dot-notation": 2,
+    "eol-last": 2,
     "eqeqeq": 2,
     "no-alert": 2,
     "no-else-return": 2,
@@ -137,7 +138,7 @@
     "no-undef-init": 2,
     "no-undef": 2,
     "no-unused-vars": 2,
-    "callback-return": 2,
+    "callback-return": [2, ["callback", "cb"]],
 
     // es6-specific
     "prefer-template": 1,
--- a/src/Server.js
+++ b/src/Server.js
@@ -26,6 +26,7 @@ import { dispatchInitialOver18 } from 'server/initialState/dispatchInitialOver18
 import { dispatchInitialTheme } from 'server/initialState/dispatchInitialTheme';
 import { dispatchInitialRecentSubreddits } from 'server/initialState/dispatchInitialRecentSubreddits';
 import metaRoutes from 'server/meta';
+import statsRouterMiddleware from 'server/meta/stats';
 
 import dispatchInitialCollapsedComments from
   'server/initialState/dispatchInitialCollapsedComments';
@@ -97,6 +98,9 @@ export function startServer() {
       buildFiles,
     ],
     getServerRouter: router => {
+      // middleware
+      statsRouterMiddleware(router, ConfigedAPIOptions);
+
       // private routes for login, logout, register, and token refresh
       loginproxy(router, ConfigedAPIOptions);
       logoutproxy(router, ConfigedAPIOptions);
--- a/src/server/meta/stats.js
+++ b/src/server/meta/stats.js
@@ -1,0 +1,43 @@
+import StatsdClient from 'statsd-client';
+
+import config from 'config';
+import errorLog from 'lib/errorLog';
+
+
+let statsdConfig;
+if (config.statsdHost) {
+  statsdConfig = {
+    host: config.statsdHost,
+    port: config.statsdPort,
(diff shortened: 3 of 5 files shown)


uzi:
Still need to test this and perhaps rearrange the code a little... but this is probably the general idea.

uzi:

Can't test -- but hey, that's what staging is for, right?  Let me know if you guys would like the code arranged a little differently.

kjoconnor:
You can just run `netcat` locally listening on udp port 8125 to see things get emitted.

uzi:
Ahh! Great suggestion -- thanks!

uzi:
Huzzah!

```
$ netcat -lup 8125
mweb2x.staging.server.request:1|c
mweb2x.staging.server.response.200:1|c
```

uzi:
Note: This patch is for counting requests and the types of responses.  There's also the desire to get the number of activeRequests, and that's a bit more complicated. Computing the number of activeRequests is handled in 1X via node's EventEmitter and the cluster stuff. Doesn't look too terrible, but I'll do that as a part of a separate patch. ("Yo dawgs, I heard you like smaller patches to review...")

src/server/meta/stats.js
```diff
@@ -0,0 +1,40 @@
...
+    port: config.statsdPort,
+    debug: config.statsdDebug,
+    prefix: config.statsdPrefix,
+    socketTimeout: config.statsdSocketTimeout,
+  };
+}
+
+const statsd = new StatsdClient(statsdConfig || {
+  _socket: { send: ()=>{}, close: ()=>{} },
+});
```
uzi: This statsd stuff could/should probably be moved to another place. Opinions as to where?

.eslintrc
```diff
@@ -137,7 +137,7 @@
     "no-undef-init": 2,
     "no-undef": 2,
     "no-unused-vars": 2,
-    "callback-return": 2,
+    "callback-return": [2, ["callback", "cb"]],
```
phil303: what prompted this change? the `next` case?

src/server/meta/stats.js
phil303: why do you think so? seems ok in here for now.

src/server/meta/stats.js
```diff
@@ -0,0 +1,40 @@
...
+        userAgent: 'SERVER',
+      }, {
+        hivemind: config.statsURL,
+      });
+    }
+
+    const status = ctx.response.status;
+    statsd.increment(`response.${status}`);
+  });
+};
```
phil303: nit, can you add a newline here

phil303:
:nail_care: small nitpicks. The code seems kosher but the larger patch goals I'll leave up to more knowledgeable minds.

src/server/meta/stats.js
uzi: Oh, just if we want it to be in a more generalized location.  Or we can instantiate a StatsdClient wherever we need one.

.eslintrc
uzi: Yeah, and it seems a koa-ism, so I think it's something we'd run into often.

src/server/meta/stats.js
uzi: Will do!

uzi:
💇

src/server/meta/stats.js
```diff
@@ -0,0 +1,43 @@
...
+    socketTimeout: config.statsdSocketTimeout,
+  };
+}
+
+const statsd = new StatsdClient(statsdConfig || {
+  _socket: { send: ()=>{}, close: ()=>{} },
+});
+
+export default router => {
+  router.use(async function (ctx, next) {
```
nramadas: a tiny nit, but could we make this an es6 style anonymous function?
`router.use(async (ctx, next) => {`

src/Server.js
```diff
@@ -103,6 +104,7 @@ export function startServer() {
       registerproxy(router, ConfigedAPIOptions);
       refreshproxy(router, ConfigedAPIOptions);
       metaRoutes(router, ConfigedAPIOptions);
+      metaStatsRoutes(router, ConfigedAPIOptions);
```
nramadas: another nit, can we call this: `metaStatsRouterMIddleware`? It's a mouthful, but unlike the other functions above it, this does not declare a route. instead, it sets middleware on our koa router, so i think the name should reflect that.

nramadas:
LGTM :fish:

PR #711, opened 2016-08-31
Compose: add escape hatch link

umbrae:
Compose is not yet ready on mweb, so we add a link to desktop compose as an alternative.

Fixes https://reddit.atlassian.net/browse/MOBILEWEB-634

Not very pretty, but it looks like this:

image

👓  @phil303

### DIFF
--- a/src/views/pages/messageCompose.jsx
+++ b/src/views/pages/messageCompose.jsx
@@ -3,10 +3,10 @@ import React from 'react';
 import MessageNav from '../components/MessageNav';
 import BasePage from './BasePage';
 
+const composeLink = 'https://www.reddit.com/message/compose';
+
 class MessageComposePage extends BasePage {
   render () {
-    const composeLink = "https://www.reddit.com/message/compose";
-
     return (
       <div className={ `message-page message-${this.props.view}` }>
         <div>
@@ -16,7 +16,9 @@ class MessageComposePage extends BasePage {
               <div className='col-xs-12 col-sm-6'>
                 <div className='well well-lg'>
                   <h3>
-                    Sorry, this isn’t ready yet! You can <a href={ composeLink }>compose a message on the desktop site</a> instead.
+                    Sorry, this isn’t ready yet! You can
+                    <a href={ composeLink }>compose a message on the desktop site</a>
+                    instead.
                   </h3>
                 </div>
               </div>
--- a/src/views/pages/messageCompose.jsx
+++ b/src/views/pages/messageCompose.jsx
@@ -1,9 +1,11 @@
 import React from 'react';
 
+import config from '../../config';
+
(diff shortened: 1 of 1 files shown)


src/views/pages/messageCompose.jsx
```diff
@@ -5,6 +5,8 @@ import BasePage from './BasePage';
 
 class MessageComposePage extends BasePage {
   render () {
+    const composeLink = "https://www.reddit.com/message/compose";
```
phil303: Mostly a style nit, can you make this a constant outside of the class? Also note as part of the [js style guide](https://github.com/reddit/styleguide/tree/master/javascript#strings), we only use single quotes.

phil303:
:fish: looks good but it's failing our linter on the single quote issue.

src/views/pages/messageCompose.jsx
umbrae: Hm I can, what's the thinking behind pulling it outside of the function just for learning purposes for me?

umbrae:
:haircut: on those just for explicitness

src/views/pages/messageCompose.jsx
phil303: It's mostly stylistic as any performance gains are negligible. But if it's a constant variable that is "tweakable" and constant across function calls, I generally pull it out and put it at the top of a file. It's a slippery slope obviously (you could probably apply this thinking to the static text below for instance) so I'm never super gung-ho about it.

Also, sorry for not being explicit here, can you make it `COMPOSE_LINK`?

phil303:
:nail_care:

src/views/pages/messageCompose.jsx
schwers: Upvote for the casing as `COMPOSE_LINK`, it's really nice to have it be obvious from the variable name that its a constant.

🔕  As a super minor nit, you can use the `reddit` property defined on `config` to build this variable, which would harden this for use in staging environments that have both mweb and desktop configured.

umbrae:
💇  Ah, fixed those things, thanks. As an ES6 noob think I was mixing up `const` and `let`.

phil303:
:fish:

powerlanguage:
@umbrae - does this preserve the url params for `to` and `subject`?

E.g. https://m.reddit.com/message/compose?to=%2Fr%2Freddit.com

umbrae:
As discussed verbally with @powerlanguage, it doesn't, but that's fine because it would be very difficult for someone to get to that URL on mobile now that we're unwhitelisting the compose resource. (They'd essentially have to be linked directly to m.reddit.com compose with those parameters, which is unlikely)

schwers:
@umbrae is it safe to close this PR now?

umbrae:
Yes, thanks

PR #712, opened 2016-09-01
Convert 1X cookies to use 2X format. Move to root domain.

schwers:
👓  @nramadas @phil303

### DIFF
--- a/assets/js/client.es6.js
+++ b/assets/js/client.es6.js
@@ -78,8 +78,8 @@ function initialize(bindLinks) {
   const $body = document.body || document.getElementsByTagName('body')[0];
   setAppMethods(app, $body, render, history);
 
-  if (app.getState('token')) {
-    app.setTokenRefresh(app, app.getState('tokenExpires'));
+  if (app.getState('redditSession')) {
+    app.setTokenRefresh(app, app.getState('redditSession').expires);
   } else if (!cookies.get('loid')) {
     setLoggedOutCookies(cookies, app);
   }
--- a/assets/js/clientLib.es6.js
+++ b/assets/js/clientLib.es6.js
@@ -118,31 +118,26 @@ export function refreshToken (app) {
   return makeRequest
     .get('/oauth2/refresh')
     .then(res => {
-      const token = res.body;
+      const session = res.body;
 
-      if (token && token.token) {
+      if (session && session.accessToken) {
         const now = new Date();
-        const expires = new Date(token.tokenExpires);
+        const expires = new Date(session.expires);
 
-        Object.assign(app.getState('ctx'), {
-          token: token.token,
-          tokenExpires: token.tokenExpires,
+        Object.assign(app.getState('ctx').redditSession, {
+          accessToken: session.accessToken,
+          expires: session.expires,
         });
 
-        app.setState('refreshingToken', false);
-        app.emit('token:refresh', token);
-
-        window.setTimeout(function() {
-          refreshToken(app).then(function() {
-            Object.assign(app.getState('ctx'), {
-              token: token.token,
-              tokenExpires: token.tokenExpires,
-            });
-
-            app.setState('refreshingToken', false);
-            app.emit('token:refresh', token);
+        if (session.refreshToken) {
+          Object.assign(app.getState('ctx').redditSession, {
+            refreshToken: session.refreshToken,
           });
-        }, (expires - now) * 0.9);
+        }
+        app.setState('refreshingToken', false);
+        app.emit('token:refresh', session);
+
+        window.setTimeout(refreshToken, (expires - now) * 0.9);
       } else {
         app.setState('refreshingToken', false);
       }
--- a/src/server/oauth.es6.js
+++ b/src/server/oauth.es6.js
@@ -6,30 +6,21 @@ import url from 'url';
 import has from 'lodash/object/has';
 
 import constants from '../constants';
+import {
+  getSession,
+  clearSessionCookies,
+  oauthTokenToSession,
+  setSessionCookies,
+} from './session';
 
 const SCOPES = 'history,identity,mysubreddits,read,subscribe,vote,submit,' +
                'save,edit,account,creddits,flair,livemanage,modconfig,' +
                'modcontributors,modflair,modlog,modothers,modposts,modself,' +
                'modwiki,privatemessages,report,wikiedit,wikiread';
 
-function nukeTokens(ctx) {
-  ctx.cookies.set('token');
-  ctx.cookies.set('tokenExpires');
-  ctx.cookies.set('refreshToken');
-}
-
 // set up oauth routes
 const oauthRoutes = function(app) {
-  app.nukeTokens = nukeTokens;
-
   const router = app.router;
-
-  const cookieOptions = {
-    secure: app.getConfig('https'),
-    secureProxy: app.getConfig('httpsProxy'),
-    httpOnly: true,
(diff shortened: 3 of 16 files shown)


src/server/session.es6.js
```diff
@@ -0,0 +1,120 @@
...
+export const COOKIE_OPTIONS = {
+  secure: config.https,
+  secureProxy: config.httpsProxy,
+  httpOnly: true,
+  maxAge: 1000 * 60 * 60 * 24 * 365,
+};
+
+export const setSessionCookies = (ctx, session) => {
+  const payloadBase64 = btoa(JSON.stringify(session));
+  const token = `${payloadBase64}.2`; // version 2
```
schwers: version constants

assets/js/clientLib.es6.js
```diff
@@ -118,31 +118,26 @@ export function refreshToken (app) {
...
-        Object.assign(app.getState('ctx'), {
-          token: token.token,
-          tokenExpires: token.tokenExpires,
+        Object.assign(app.getState('ctx').redditSession, {
+          accessToken: session.accessToken,
+          expires: session.expires,
         });
 
+        if (session.refreshToken) {
+          Object.asseign(app.getState('ctx').redditSession, {
```
schwers: `Object.assign`

assets/js/client.es6.js
```diff
@@ -78,7 +78,7 @@ function initialize(bindLinks) {
   const $body = document.body || document.getElementsByTagName('body')[0];
   setAppMethods(app, $body, render, history);
 
-  if (app.getState('token')) {
+  if (app.getState('session')) {
```
nramadas: 🔕  just want to make sure `session` is intentional, and it's not supposed to be `redditSession`.

src/server/oauth.es6.js
```diff
@@ -499,7 +473,8 @@ const oauthRoutes = function(app) {
           resolve();
         }
 
-        const message = obj.message ? `&message=${obj.message}` : '';
+        const message = obj.message ?
```
nramadas: 🔕  why the change?

src/server/session.es6.js
```diff
@@ -0,0 +1,123 @@
...
+  if (tokenCookie.indexOf('.') === -1) {
+    if (!oldExpiresCookie || !oldRefreshCookie) {
+      // we have a token cookie that looks like the new format, but we
+      // don't have the expiration and refreshToken, just clear auth cookies
+      clearSessionCookies(ctx.cookies);
+      return;
+    }
+
+    // clear the old tokenExpires and refreshToken cookies
+    clearSessionCookies(ctx.cookies);
```
nramadas: up to you, but an alternate organization of this code could look like:

```es6
if (tokenCookie.indexOf('.') === -1) {
  // this is an old style token cookie. we want to go ahead and clear all the pertinent
  // session cookies.
  clearSessionCookies(ctx.cookies);

  // if we have both an 'expires' cookie and a 'refresh' cookie, create a new session out of it.
  if (oldExpiresCookie && oldRefreshCookie) {
    const newSession = {
      accessToken: tokenCookie,
      expires: oldExpiresCookie,
      refreshToken: oldRefreshCookie,
    };

    setSessionCookies(ctx, newSession);
    return newSession;
  }
}

```

src/server/session.es6.js
```diff
@@ -0,0 +1,123 @@
...
+};
+
+
+export const SESSION_COOKIES = [
+  'token',
+  'tokenExpires',
+  'refreshToken',
+];
+
+export const clearSessionCookies = cookies => {
```
nramadas: :no_bell: Nit, but I think it would be nice for `clearSessionCookies` to take in `ctx`, so that its signature is complementary to `setSessionCookies`.

src/server/session.es6.js
```diff
@@ -0,0 +1,123 @@
...
+    if (!session || !session.accessToken || !session.refreshToken || !session.expires) {
+      clearSessionCookies(ctx.cookies);
+      return;
+    }
+
+    return session;
+  } catch (e) {
+    // clear the malformed cookies
+    clearSessionCookies(ctx.cookies);
+    // implicit return
```
nramadas: 🔕  It might be reasonable to just `return` here instead of leaving a comment explaining that the return is implicit :p

src/server/session.es6.js
```diff
@@ -0,0 +1,123 @@
...
+  }
+};
+
+export const COOKIE_OPTIONS = {
+  secure: config.https,
+  secureProxy: config.httpsProxy,
+  httpOnly: true,
+  maxAge: 1000 * 60 * 60 * 24 * 365,
```
nramadas: 1x code, i know, but should this number live in a constant?

src/views/pages/login.jsx
```diff
@@ -263,15 +263,19 @@ class LoginPage extends BasePage {
 
       // do some redirection here.
       if (res && res.body) {
-        const { token } = res.body.token;
+        const { redditSession } = res.body;
+
+        const ctx = app.getState('ctx');
```
nramadas: 🔕  so, in 1X, state that comes out of `getState` isn't a new reference, right? so, to keep things consistent with how it exists elsewhere, this could read:

```es6
Object.assign(app.getState('ctx'), {
  ...ctx,
  redditSession: {
    ...ctx.redditSession,
    ...redditSession,
  },
});
```

nramadas:
:nail_care:

src/server/session.es6.js
schwers: I think I may have been overzealous with comment details in this, do you have any objection to just removing the comment and not having an explicit return? When I look at it now it doesn't seem necessary

src/server/session.es6.js
schwers: Sure, that's more readable

src/server/session.es6.js
nramadas: nope, that's fine by me too

src/server/session.es6.js
schwers: Having the parameter be `ctx` would be some satisfying symmetry. I liked only requiring `cookies` so callers understand `clearSessionCookies` only needs to use `ctx.cookies`. With that in mind, would you still prefer it taking `ctx`?

src/server/oauth.es6.js
schwers: bad-edit, thanks

src/server/session.es6.js
nramadas: i think i would still prefer `ctx`. The other reason being, we understand `ctx` to be something we mutate on a regular basis (and has a clear meaning on our codebase: it's koa ctx). However, `cookies` isn't quite as clear cut. And, in this particular case, though we are calling methods on `cookies`, ultimately, it's to mutate the koa `ctx` object, so passing in `ctx` feel appropriate.

assets/js/client.es6.js
schwers: good catch

src/views/pages/login.jsx
schwers: Yes, it's not technically a new a reference, I dislike that assumption in 1X code so I'll leave it as is. (From a quick look `setState` actually seems more consistent)

nramadas: I'm fine with that 👍 . I was just curious for my own understanding

schwers:
💇

nramadas:
one question, but other than that, LGTM :fish:

PR #713, opened 2016-09-01
Convert 1X style cookies

schwers:
1X uses a different style of token cookie than 2X. There's a corresponding patch for 1X that does this work as well, but we need to do it in 2X in case the user hasn't visited in a while and still has a 1X era cookie. The format of the new cookie is `${Base64EncodedJson}.${versionNumber}`; Where `.` is a separator that will allow extra data past versionNumber in the future. For now, the versionNumber is 2 (the old format is implicitly 1). The encoded data should contain `accessToken`, `expires`, `refreshToken`, and scope information.

See https://github.com/reddit/reddit-mobile/pull/712 for the 1X patch

👓  @nramadas @phil303


src/server/session/dispatchSession.js
```diff
@@ -33,7 +59,11 @@ export default async (ctx, dispatch, apiOptions) => {
   // session.
   if (session && sessionData && !session.isValid) {
```
schwers: consider refreshing 1X converted cookies here

src/server/session/clearSessionCookies.js
```diff
@@ -0,0 +1,20 @@
...
+
+  SESSION_COOKIES.map(cookieName => {
+    // clear the cookie and it's signed version for `m.reddit`
+    cookies.set(cookieName);
+    cookies.set(`${cookieName}.sig`);
+    // clear the cookie and it's signed version from `.reddit`
+    cookies.set(cookieName, undefined, rootOptions);
+    cookies.set(`${cookieName}.sig`, undefined, rootOptions);
+  });
+};
```
schwers: V delete constants below

src/server/session/clearSessionCookies.js
```diff
@@ -0,0 +1,20 @@
...
+export const SESSION_COOKIES = [
+  'token', // v1+ cookie
+  'tokenExpires', // v1 only
+  'refreshToken', // v1 only
+];
+
+export default cookies => {
+  const rootOptions = { domain: config.rootReddit };
+
+  SESSION_COOKIES.map(cookieName => {
```
schwers: `forEach` instead of `map`

src/server/session/sessionDataFromCookies.js
```diff
@@ -0,0 +1,94 @@
...
+  const [payloadBase64, version] = tokenCookie.split(SEPERATOR);
+  if (parseInt(version, 10) !== VERSION) { // old format is implicit version 1
+    // this is mostly a sanity check. if the version is not 2 (current) then the
+    // the cookie is malformed somehow, or is a new format we don't expect.
+    clearSessionCookies(ctx.cookies);
+    return;
+  }
+
+  const session = JSON.parse(atob(payloadBase64));
+  if (!session || !session.accessToken || !session.refreshToken || !session.expires) {
```
nramadas: suggestion:

```es6
import has from 'lodash/has';
/* ... */
if (!session || !has(session, ['accessToken', 'refreshToken', 'expires']) { /* ... */ }
```

nramadas: nvm! `has` doesn't work like that apparently :(

schwers:
💇

nramadas:
this lgtm :fish:

PR #715, opened 2016-09-06
Ads bugs

dwick:
👓 @schwers 

Fixed some easy ones from https://reddit.atlassian.net/browse/ADS-692

### DIFF
--- a/src/app/actions/ads.js
+++ b/src/app/actions/ads.js
@@ -4,6 +4,8 @@ import uniqueId from 'lodash/uniqueId';
 import config from 'config';
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import isFakeSubreddit from 'lib/isFakeSubreddit';
+import { AD_LOCATION } from 'app/constants';
+
 
 const { PostsEndpoint } = endpoints;
 const { PostModel } = models;
@@ -119,6 +121,7 @@ export const fetchAddBasedOnResults = async (dispatch, state, adId, postsList, p
     dt,
     site,
     platform: 'mobile_web',
+    placement: `feed-${AD_LOCATION}`,
     raw_json: '1',
   };
 
--- a/src/app/actions/ads.js
+++ b/src/app/actions/ads.js
@@ -4,7 +4,7 @@ import uniqueId from 'lodash/uniqueId';
 import config from 'config';
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import isFakeSubreddit from 'lib/isFakeSubreddit';
-import { AD_LOCATION } from 'app/constants';
+import adLocationForPostRecords from 'lib/adLocationForPostRecords';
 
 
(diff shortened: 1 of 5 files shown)


src/app/actions/ads.js
```diff
@@ -123,9 +125,10 @@ export const fetchAddBasedOnResults = async (dispatch, state, adId, postsList, p
   const site = getSite(pageParams);
 
   const data = {
-    dt,
     site,
+    dt: dt.join(','),
     platform: 'mobile_web',
+    placement: `feed-${AD_LOCATION}`,
```
schwers: 🔕  Small question, because this will only affect listings that have less than `AD_LOCATION` items. Would it be helpful to compute what slot the ad unit will occupy in practice? (again, this only seems to matter for listings without much content) [actual position computed here](https://github.com/reddit/reddit-mobile/blob/2X/src/app/components/PostsList/index.jsx#L75)

src/app/actions/ads.js
```diff
@@ -140,6 +143,7 @@ export const fetchAddBasedOnResults = async (dispatch, state, adId, postsList, p
     const ad = await getAd(apiOptionsFromState(state), data);
     dispatch(received(adId, ad));
   } catch (e) {
+    console.log('add error', e);
```
schwers: Extra `.log`. 

🔕  In dev we see this ad fetch getting called from time-to-time, it would be awesome to have your insight on what we should consider an error here. e.g. is getting a response of 'no ad for this listing' actually an error we should be bucketing into mweb logs

schwers:
💅

src/app/actions/ads.js
```diff
@@ -160,9 +164,15 @@ export const getAd = (apiOptions, data) => {
   return new Promise((resolve, reject) => {
     // why is this a post??
     rawSend(apiOptions, 'post', config.adsPath, data, 'form', (err, res) => {
-      if (!res || !res.body || res.status !== 200) {
+      if (err) {
```
dwick: i think i missed the case for when there simply is no response (no ad available). can you check that?

src/app/actions/ads.js
dwick: yeah, i was struggling to figure out how to pass that data around, but the `postList` is above so we could probably factor the function out that's used in the view.  would definitely be better to get the accurate position

src/app/actions/ads.js
dwick: i think it's a 200 with a blank response

src/app/actions/ads.js
dwick: no ad isn't an error, sometimes there is just nothing available. it'd probably want to make the promise resolve and do a null check on `ad` above before dispatching. does that make sense?

schwers:
👓  @phil303 or @nramadas

nramadas:
:fish:

PR #716, opened 2016-09-07
Add kage manifests

spladug:
These are the supporting files so we can do containerized staging environments.

Once these are on the main 2X branch, any branches based on it will be stagable with kage.

:eyeglasses: @gtaylor @schwers

### DIFF
--- a/.kage/manifests/deployment.yml
+++ b/.kage/manifests/deployment.yml
@@ -1,0 +1,66 @@
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: DEPLOYMENT_NAME
+  namespace: NAMESPACE_NAME
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        # The branch's Service uses these labels to find the pods to route
+        # traffic to.
+        kage.repo.branch.name: BRANCH_NAME
+      annotations:
+        # These will be populated as the pod starts.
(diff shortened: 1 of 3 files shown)


gtaylor:
:fish:

gtaylor:
cc'ing @ckwang8128 for reference. Also, @nramadas @phil303. @prashtx for funzies.

.kage/manifests/deployment.yml
```diff
@@ -0,0 +1,66 @@
...
+            - name: PORT
+              value: '4444'
+            - name: HTTPS
+              value: 'false'
+
+            - name: API_HEADERS
+              value: 'false'
+            - name: API_USER_AGENT
+              value: switcharoo-reddit
+            - name: API_USER_AGENT
```
schwers: Is this a duplicate from the line var above?

schwers:
💅  Super small question about the config

.kage/manifests/deployment.yml
gtaylor: This is probably my fault. Indeed it is. I'm going to be nuking pretty much everything under `env` as soon as I can finish up some cluster shuffling, though.

spladug:
:haircut:

schwers:
:fish:

spladug:
7a283637410c42b9b0aa45b68cf9258d6a849e7d

PR #718, opened 2016-09-07
Add .ackrc

uzi:
Ack is a useful find/grep sort of utility. It can
help folks understand out codebase quicker.  This
hides generated files from searches to make things
friendlier.

### DIFF
--- a/.ackrc
+++ b/.ackrc
@@ -1,0 +1,4 @@
+--ignore-dir=bin
+--ignore-dir=build
+--ignore-dir=node_modules
+--ignore-dir=.test


phil303:
I think all we need is `bin` and `build`. `node_modules` I ack from time to time, `.test` should be getting deleted after tests are run I believe - if not, there's a bug, and .git should never be a problem.

phil303:
:nail_care:

schwers:
I'm actually in favor of having `.test` explicitly ignored. The likelihood of it popping up is exceptionally rare, but it seems good to rule out this potential source of confusion.

schwers:
💅

uzi:
If you want to search node_modules, you can just `cd` into it.

uzi:
💇 I took out the `.git` directory but left `node_modules` in.  The reason is my use case is that I don't want to search `node_modules` unless I explicitly intend to, which can be done by `cd`ing into it.

phil303:
ok sounds reasonable to me :fish:

PR #720, opened 2016-09-08
Use loids generated by the api

nramadas:
Through clever manipulation of 1X code (read: hacks and frustration), I think I have a solution for using api generated loids in 1X.

Essentially, I repurposed the current middleware used to create loids, and instead had the middleware yield on the route handling first so that the loids can be loaded by the api.

Blocked on: https://github.com/reddit/node-api-client/pull/151

:eyeglasses: @phil303 || @schwers || @uzi

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -326,6 +326,20 @@ function routes(app) {
       this.props.adsEnabled = !feature.enabled(constants.flags.NO_ADS);
     });
 
+    const loggedOutUserPromise = this.props.data.get('loggedOutUser');
+
+    if (loggedOutUserPromise) {
+      // we need to yield here to ensure that the logged out user data gets
+      // loaded in.
+      yield loggedOutUserPromise.then(user => {
+        this.props.loid = user.body.loid;
+        this.props.loidcreated = String(user.body.loid_created);
+      });
+    } else {
+      this.props.loid = '';
+      this.props.loidcreated = '';
+    }
+
     return yield next;
   }
 
--- a/src/server/index.jsx
+++ b/src/server/index.jsx
@@ -222,11 +222,15 @@ class Server {
       server.use(koaStatic(`${__dirname}/../../build`));
     }
 
+    // these first two yield first _before_ taking action, so they are ordered
+    // in "reverse" and at the top;
+    server.use(this.setExperiments(app));
+    server.use(this.setLOID(app));
+
+    // these do work first before yielding.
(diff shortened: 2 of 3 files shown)


src/server/index.jsx
```diff
@@ -305,6 +315,9 @@ class Server {
         return;
       }
 
+      // yield first since the route handling might set loids
+      yield next;
```
nramadas: we need to let the `setLOID` middleware do its thing first. so, we'll yield, then read the loids after.

src/server/index.jsx
```diff
@@ -278,22 +282,28 @@ class Server {
 
   setLOID (app) {
     return function * (next) {
+      // if we have loid cookies, set them on the context. certain actions will
+      // depend on these priliminary loids. if they are wrong, they will get
```
phil303: it's "preliminary" NEWB

phil303:
:runner: gonna pass this onto someone who know this code better

nramadas:
:eyeglasses: @schwers ? have a minute to take a look?

src/routes.jsx
```diff
@@ -328,6 +328,18 @@ function routes(app) {
       this.props.adsEnabled = !feature.enabled(constants.flags.NO_ADS);
     });
 
+    const loggedOutUserPromise = this.props.data.get('loggedOutUser');
+
+    if (loggedOutUserPromise) {
+      loggedOutUserPromise.then(user => {
+        this.props.loid = user.body.loid;
+        this.props.loidcreated = String(user.body.loid_created);
```
schwers: Should we move `yield next` inside this then block? I think we want to to make sure the loggedOutUser promises loads first before anything else, so `buildAPIOptions` passes the correct `loid` to api calls. Though, this is 1X so I may be missing something "clever" ;)

schwers:
💅

nramadas:
:haircut:

src/routes.jsx
```diff
@@ -328,6 +328,18 @@ function routes(app) {
       this.props.adsEnabled = !feature.enabled(constants.flags.NO_ADS);
     });
 
+    const loggedOutUserPromise = this.props.data.get('loggedOutUser');
+
+    if (loggedOutUserPromise) {
+      yield loggedOutUserPromise.then(user => {
```
schwers: Can you add a comment for why we yield here?

schwers:
:fish:

PR #721, opened 2016-09-08
Add "Continue this thread" support

uzi:
There is the side-effect that clicking the link and then navigating back will then show the additional comments in the tree, but my feeling it's better to have this problem than the other way around.

👓 @schwers @nramadas @phil303


phil303:
uzi, can you slap some pics onto this when you get a chance?

src/app/components/Comment/index.jsx
```diff
@@ -153,6 +154,15 @@ function renderReplies(props) {
     'm-no-indent': nestingLevel >= NESTING_STOP_LEVEL,
   });
 
+  let moreContent = null;
+  if (comment.loadMore) {
+    if (comment.loadMoreIds && comment.loadMoreIds.length) {
```
phil303: You shouldn't need the first `loadMoreIds` check here since that is now a default property on the comment model

src/app/components/Comment/index.jsx
```diff
@@ -161,9 +171,22 @@ function renderReplies(props) {
...
+    </div>
+  );
+}
 
-      { comment.loadMoreIds.length ? renderMoreCommentsButton(props) : null }
+function renderContinueThisThread(props) {
+  const { comment, nestingLevel } = props;
 
+  const className = cx('Comment__continueThread', {
+    'm-no-indent': nestingLevel >= NESTING_STOP_LEVEL,
```
phil303: One thing I pondered was having a derived prop in `mergeProps` that does this for you, maybe something like `isIndented`. Since I believe this is the third check of this type in the file, it might be time to consider that.

src/app/components/Comment/index.jsx
phil303: The logic in this block seems not obvious and tied to the whims of our api. What if we had a more explicit boolean in api-client - `canContinueThread`? (needs work) 

I also kind of wonder if we can just remove `loadMore` then as it'd be a derived value.

phil303:
:nail_care:

src/app/components/Comment/index.jsx
uzi: Yeah, I put in that extra check as a just in case, but you're right.

The thing with loadMore is we're seeing that yes there is more to load but we don't know which ids... in which case we know to "Continue this thread".  I'll think about what else we can do.

src/app/components/Comment/index.jsx
uzi: Noted. Also I noticed that the dots aren't really working... without this patch as well.

src/app/components/Comment/index.jsx
phil303: I understand that. I think we can help the next person who has to look at this code by just giving out a very easy to reason about flag. The `loadMore` thing is an artifact that I think is safe to remove now.

Put it this way:
If you were coming across the comment object for the first time and you saw `loadMore: true` and `loadMoreIds: []`, is it a completely 100% safe assumption that this means that you need a "continue this thread"? Or would you follow up on it and do your homework to make sure that's actually the case?

PR #723, opened 2016-09-09
Check and update state.finished atomically

prashtx:
We may have data promises that lead to other data promises, which result
in multiple checks for `this.state.finished` when a page has finished
rendering. `this.setState` doesn't act immediately, so both checks will
see a false value, and we'll take page-finished actions twice. Instead,
we should check and set the state as an atomic operation.

### DIFF
--- a/src/views/pages/BasePage.jsx
+++ b/src/views/pages/BasePage.jsx
@@ -27,8 +27,9 @@ class BasePage extends BaseComponent {
       data: {},
       meta: {},
       loaded: !!props.dataCache,
-      finished: false,
     };
+
+    this.finished = false;
 
     const state = this.state;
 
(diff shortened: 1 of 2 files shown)


prashtx:
👓  @uzi  @schwers

src/views/pages/BasePage.jsx
```diff
@@ -145,10 +145,12 @@ class BasePage extends BaseComponent {
       this.props.app.emit('pageload', data);
 
       // pageview fires when we should track a pageview
-      if (this.state.finished === false && this.track) {
-        this.props.app.emit('pageview', data);
-        this.setState({ finished: true });
-      }
+      this.setState((oldState) => {
```
schwers: tiny nit, we've been removing parens on single argument arrow functions

prashtx:
💇

schwers:
:fish:

prashtx:
Accidentally pulled in an unrelated patch that's on master...

💇  /cc @schwers

schwers:
:fish: Can you make sure to add a description of what was going on and what we tried to the squashed commit?

PR #724, opened 2016-09-09
Count filtered r/all as a fake subreddit

schwers:
We get a lot of hits to `r/all-<subreddit-A>-<subreddit-B>-...-<subreddit-Z>`. This filtered view of `r/all` is valid for gold users, otherwise we return the normal list of posts. However our fake subreddit detector doesn't consider this filtered view to be faking, causing an error when we try to fetch subreddit info.

👓  @prashtx @nramadas


prashtx:
:fish:

PR #725, opened 2016-09-09
Fix unhandledrejection listener passing undefined ctx

uzi:
This ends up bubbling to the error handler which
wants to check ctx.env, but ctx is undefined.  Instead
we pull ctx from app.state.

👓  @schwers @nramadas


schwers:
:100: for thoughtful commit messages, thank you

schwers:
:fish:

PR #726, opened 2016-09-10
Collect shell render RUM timings

uzi:
Gathers statistics on loading the initial state of
the app, which is before any of the API requests
have happened.

We send the data to hivemind which ends up logging to graphite something like

```
service_time.web.m2.server.shell.frontend.request:65.00005722045898|ms
service_time.web.m2.server.shell.frontend.response:0.9999275207519531|ms
service_time.web.m2.server.shell.frontend.dom_loading:83.9998722076416|ms
service_time.web.m2.server.shell.frontend.mount:90|ms
```

Note that we'll need to set the `process.env.ACTION_NAME_SECRET` in configuration for this to work.

👓 @schwers @phil303 @nramadas

### DIFF
--- a/src/lib/timing.js
+++ b/src/lib/timing.js
@@ -47,23 +47,23 @@ export function getTimes() {
 }
 
 export function sendTimings(beginRender) {
-    if (Math.random() > 0.1) { // 10% of requests
-      return;
-    }
+  if (Math.random() > 0.1) { // 10% of requests
+    return;
+  }
 
-    let actionName = `m2.server.shell`;
+  const actionName = 'm2.server.shell';
 
-    const timings = Object.assign({
-      actionName: actionName,
-    }, getTimes());
+  const timings = Object.assign({
+    actionName,
+  }, getTimes());
 
-    timings.mountTiming = (Date.now() - beginRender) / 1000;
+  timings.mountTiming = (Date.now() - beginRender) / 1000;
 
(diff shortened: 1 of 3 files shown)


src/lib/timing.js
```diff
@@ -1,4 +1,8 @@
-function getTimes() {
+import { DEFAULT_API_TIMEOUT } from 'app/constants';
```
schwers: 🔕  This patch made me realize we should be passing `DEFAULT_API_TIMEOUT` in `apiOptionsFromState`

src/server/meta/index.js
```diff
@@ -151,6 +151,9 @@ export default (router, apiOptions) => {
         .send({ rum: timings })
         .timeout(DEFAULT_API_TIMEOUT)
         .end(err => err ? reject(err) : resolve());
+
+    ctx.body = null;
```
schwers: 🔕  Just my own curiosity, why the explicit `null` and return here?

schwers:
:fish:

src/server/meta/index.js
uzi: Following what other methods did here... also, it bombs out without the ctx.body line ... not sure about the return.

src/lib/timing.js
uzi: That entails a change to api-client, right?

PR #727, opened 2016-09-11
Add originalUrl to loginRedirect

jamesmaa:
👓 @uzi 

Problem: Any action that requires an account redirects to the registration page. Unfortunately, when a user comes in from SEO or direct link causes `ctx.originalUrl` not to be set and closing the register window causes them to be redirected to the frontpage instead of the original post. 

The problem is that `ctx.originalUrl` is not set when a comment page is directly accessed, so a login page cannot access that property and therefore falls back to returning to frontpage.

The fix is easy. Whenever actions are called that requires a logged in user, add a query parameter to the redirect if it exists, and if the path is the frontpage, silence the query parameter for `ctx.originalUrl` to override.

### DIFF
--- a/assets/js/appMethods.es6.js
+++ b/assets/js/appMethods.es6.js
@@ -1,5 +1,6 @@
 import ReactDOM from 'react-dom';
 import cookies from 'cookies-js';
+import querystring from 'querystring';
 
 import { setMetaColor, setTitle, refreshToken } from './clientLib';
 import constants from '../../src/constants';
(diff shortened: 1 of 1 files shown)


schwers:
@jamesmaa I wonder if there's a way we can make koa-router see referrer across in-app navigations. In theory the referrer check in `routes.jsx::loginRegisterOriginalUrl` should handle this, but you're right, we need something like this too unfortunately.

Small nit, your PR has a really good description! Can you put it in the commit message body too? Knowing this kind of stuff is super helpful in the future if we're spelunking through commit logs.

jamesmaa:
💇 Cool done

schwers:
Woo! thanks @jamesmaa :fish:

uzi:
:fish:

jamesmaa:
ee16269a8a5dd3353185665ce719feef05ed3969

PR #714, opened 2016-09-01
Add recent clicks and recent srs.

jamesmaa:
👓 @nramadas Same commit as last time, but readded the route I accidentally removed.

## TODOS:
- [x] move the post visits over to cookies instead of localstorage
- [x] be 100% sure these cookies are compatible with desktop cookies
- [x] use the options object to set the domain correctly

PR #728, opened 2016-09-11
Store and convert localStorage.visitedPosts in JSON

uzi:
We were writing a comma separated string in 1X and
it's a JSON array in 2X. This will store things as
an array and convert from old to new.

👓 @schwers @phil303 @nramadas


jamesmaa:
Does this conflict with https://github.com/reddit/reddit-mobile/pull/714 ?

I'm storing it as a cookie because it allows .reddit.com to access them

nramadas:
I don't know much about the visited posts code, but from what I can tell, this LGTM :fish:

PR #729, opened 2016-09-11
Support old style localStorage.visitedPosts

uzi:
If we see the old style visitedPosts, we're able to
read them. Also added a bit of code to handle the
initial state a bit better.

👓 @schwers @phil303 @nramadas

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -64,20 +64,34 @@ const client = Client({
 
     data.preferences = models.Preferences.fromJSON(data.preferences);
 
-    data.collapsedComments = {};
     data.meta.env = 'CLIENT';
 
+    // Pull some defaults from localStorage (if available)
     if (isLocalStorageAvailable()) {
       try {
-        data.collapsedComments = JSON.parse(window.localStorage.collapsedComments);
+        const collapsedComments = window.localStorage.collapsedComments;
+        if (collapsedComments !== undefined) {
+          data.collapsedComments = JSON.parse(collapsedComments);
+        }
       } catch (e) { console.warn(e); }
 
       try {
-        data.expandedPosts = JSON.parse(window.localStorage.expandedPosts);
+        const expandedPosts = window.localStorage.expandedPosts;
+        if (expandedPosts !== undefined) {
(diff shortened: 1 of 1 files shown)


src/Client.js
```diff
@@ -60,20 +60,39 @@ const client = Client({
 
     data.preferences = models.Preferences.fromJSON(data.preferences);
 
-    data.collapsedComments = {};
     data.meta.env = 'CLIENT';
 
+    // Set defaults for values we pull from localStorage ahead of time
+    // in case we don't have localStorage to turn to.
+    data.collapsedComments = {};
```
schwers: I think we shouldn't be setting these here. State should already have proper default values as defined by reducers. In theory this function should only be used for state manipulation reducers can't handle like turning JSON into JS classes or pulling data our of localstorage (our current use case).

(Just as a side-note, if we do this, we should export the `DEFAULT` constant defined in the reducer so we don't have a divergence in type or value)

src/Client.js
```diff
@@ -60,20 +60,39 @@ const client = Client({
...
+          data.expandedPosts = JSON.parse(expandedPosts);
+        }
       } catch (e) { console.warn(e); }
 
       try {
-        data.visitedPosts = JSON.parse(window.localStorage.visitedPosts);
+        const visitedPosts = window.localStorage.visitedPosts;
+        if (visitedPosts !== undefined) {
+          if (!visitedPosts.startsWith('[')) {
+            // Old format -- comma separated string
```
schwers: Should we make a patch to 1X to start reading/writing the new format? I'm worried about what happens if a user changes buckets.

schwers:
💅

src/Client.js
uzi: See #728

src/Client.js
uzi: I'm fine with not setting them there ... I'll check if things are still happy if this isn't set, but coming in cold without any of these set, we end up throwing a few errors...

screen shot 2016-09-13 at 8 55 36 pm

uzi:
💇

schwers:
:fish:

PR #730, opened 2016-09-12
Pageview events for most popular pages

phil303:
The pages this adds events to are the following:

- subreddits
- special subreddits (frontpage, all, etc)
- profiles
- messages
- login
- registration
- posting

This list is derived from Justin's comment on this [ticket](https://reddit.atlassian.net/browse/DA-638?focusedCommentId=55474&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-55474).

This does things differently than the original events patch. The original patch had middleware listening for `SET_PAGE`. The problem with this approach is that pages have very different field requirements. Sometimes the data for those fields wasn't even available unless it was asynchronously fetched for. The code was fast becoming a mess of special handling statements.

The approach this takes is to ask for the required data explicitly in the handler and wait for their responses to make the event request. This obviously comes at the expense of DRY-ness but I think overall is a win for clarity/maintainability.

One thing to note, is that tracking should only be happening server side. Each tracking call is protected by an "is client" check. One thing we might want to consider is putting this check into the eventTracker code directly.

TODO:
- [x] `screenview` name should have `mweb` in it? (check with Justin on this)
- [x] blocked on @umbrae checking with data team on the multi-logging issue
- [x] correct tracking endpoint and credentials

### DIFF
--- a/src/app/actions/ads.js
+++ b/src/app/actions/ads.js
@@ -70,7 +70,7 @@ const nextAdId = () => (uniqueId('ad_'));
 // We could look up page params from state, but having them passed in
 // is more explicit and safer if the page were to change while we await.
 export const fetchNewAdForPostsList = (postsListId, pageParams) =>
-  async (dispatch, getState, { waitForState }) => {
+  async (dispatch, getState) => {
     const state = getState();
     const adRequest = state.adRequests[postsListId];
     if (adRequest && adRequest.loading) { return; }
@@ -78,13 +78,6 @@ export const fetchNewAdForPostsList = (postsListId, pageParams) =>
     const adId = nextAdId();
 
     dispatch(fetching(adId, postsListId));
-
-    // We don't want to show ads if there are no posts to render, so we need
-    // to wait for the postsList to load.
-    await waitForState(state => {
-      const postList = state.postsLists[postsListId];
-      return postList && !postList.loading;
-    });
 
     const loadedState = getState();
     const postsList = loadedState.postsLists[postsListId];
--- a/src/app/router/handlers/CommentsPage.js
+++ b/src/app/router/handlers/CommentsPage.js
@@ -4,10 +4,13 @@ import * as platformActions from '@r/platform/actions';
 import { models } from '@r/api-client';
 
 import { cleanObject } from 'lib/cleanObject';
-import * as commentsPageActions from 'app/actions/commentsPage';
-import * as replyActions from 'app/actions/reply';
 import { fetchUserBasedData } from './handlerCommon';
+import * as commentsPageActions from 'app/actions/commentsPage';
+import * as subredditActions from 'app/actions/subreddits';
+import * as replyActions from 'app/actions/reply';
+import { getBasePayload, buildSubredditData, convertId, logClientScreenView } from 'lib/eventUtils';
 import { paramsToCommentsPageId } from 'app/models/CommentsPage';
+
 
 const { POST_TYPE } = models.ModelTypes;
 const PostIdRegExp = new RegExp(`^${POST_TYPE}_`);
@@ -18,7 +21,9 @@ const ensurePostTypePrefix = postId => {
   return `${POST_TYPE}_${postId}`;
 };
 
+
 export default class CommentsPage extends BaseHandler {
+
   static pageParamsToCommentsPageParams({ urlParams, queryParams}) {
     let { postId } = urlParams;
     const { commentId } = urlParams;
@@ -41,23 +46,28 @@ export default class CommentsPage extends BaseHandler {
     });
   }
 
-  async [METHODS.GET](dispatch, getState, { waitForState }) {
+  async [METHODS.GET](dispatch, getState) {
     const state = getState();
     if (state.platform.shell) { return; }
 
     const commentsPageParams = CommentsPage.pageParamsToCommentsPageParams(this);
-
-    dispatch(commentsPageActions.fetchCommentsPage(commentsPageParams));
-    fetchUserBasedData(dispatch);
-
-    dispatch(commentsPageActions.fetchRelevantContent());
-    dispatch(commentsPageActions.visitedCommentsPage(this.urlParams.postId));
-
     const commentsPageId = paramsToCommentsPageId(commentsPageParams);
-    await waitForState(state => {
-      const commentsPage = state.commentsPages[commentsPageId];
-      return commentsPage && !!commentsPage.responseCode;
-    }, state => dispatch(setStatus(state.commentsPages[commentsPageId].responseCode)));
+    const { subredditName } = state.platform.currentPage.urlParams;
+
+    await Promise.all([
+      dispatch(commentsPageActions.fetchCommentsPage(commentsPageParams)),
+      fetchUserBasedData(dispatch),
+
+      dispatch(commentsPageActions.fetchRelevantContent()),
+      dispatch(commentsPageActions.visitedCommentsPage(this.urlParams.postId)),
+
+      dispatch(subredditActions.fetchSubreddit(subredditName)),
+    ]);
+
+    dispatch(setStatus(getState().commentsPages[commentsPageId].responseCode));
+
+    const _tempState = getState();
+    logClientScreenView(buildScreenViewData(getState()), _tempState);
   }
 
   async [METHODS.POST](dispatch, getState, { waitForState }) {
@@ -87,4 +97,22 @@ export default class CommentsPage extends BaseHandler {
       }
     });
   }
+}
+
+
+function buildScreenViewData(state) {
+  const { currentPage: { queryParams, urlParams } } = state.platform;
+  const fullName =`t3_${urlParams.postId}`;
+  const post = state.posts[fullName];
+
+  return cleanObject({
+    target_fullname: fullName,
+    target_id: convertId(post.id),
+    target_type: post.isSelf ? 'self' : 'link',
+    target_sort: queryParams.sort || 'confidence',
+    target_url: post.cleanUrl,
+    target_filter_time: queryParams.sort === 'top' ? (queryParams.time || 'all') : null,
+    ...buildSubredditData(state),
+    ...getBasePayload(state),
+  });
 }
--- a/src/app/router/handlers/PostsFromSubreddit.js
+++ b/src/app/router/handlers/PostsFromSubreddit.js
@@ -5,11 +5,15 @@ import * as adActions from 'app/actions/ads';
 import * as postsListActions from 'app/actions/postsList';
 import * as subredditActions from 'app/actions/subreddits';
 import { paramsToPostsListsId } from 'app/models/PostsList';
+import { FAKE_SUBS } from 'lib/isFakeSubreddit';
 import { cleanObject } from 'lib/cleanObject';
 import { listingTime } from 'lib/listingTime';
 import { fetchUserBasedData } from './handlerCommon';
+import { getBasePayload, buildSubredditData, convertId, logClientScreenView } from 'lib/eventUtils';
+
 
 export default class PostsFromSubreddit extends BaseHandler {
+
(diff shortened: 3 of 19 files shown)


src/app/actions/ads.js
```diff
@@ -79,13 +79,6 @@ export const fetchNewAdForPostsList = (postsListId, pageParams) =>
 
     dispatch(fetching(adId, postsListId));
 
-    // We don't want to show ads if there are no posts to render, so we need
-    // to wait for the postsList to load.
-    await waitForState(state => {
```
phil303: This is no longer needed as we're `await`ing on this data earlier in the code flow. See this [line](https://github.com/reddit/reddit-mobile/pull/730/files#diff-13314b044d0d3ddc8a9ddb78dbe58412R53).

src/config.js
```diff
@@ -38,10 +38,11 @@ const config = () => ({
   adsPath: process.env.ADS_PATH || '/api/request_promo.json',
   manifest: {},
 
-  trackerKey: process.env.TRACKER_KEY,
-  trackerEndpoint: process.env.TRACKER_ENDPOINT,
-  trackerClientSecret: process.env.TRACKER_SECRET,
-  trackerClientAppName: process.env.TRACKER_CLIENT_NAME,
+  // TODO:
+  trackerKey: process.env.TRACKER_KEY || 'XXX',
```
phil303: @schwers do you know if I can just lift this directly from 1X?

src/lib/eventTracker.js
```diff
@@ -0,0 +1,67 @@
...
+function calculateHash(key, string) {
+  const hmac = crypto.createHmac('sha256', key);
+  hmac.setEncoding('hex');
+  hmac.write(string);
+  hmac.end();
+
+  return hmac.read();
+}
+
+function post({url, data, query, headers, done}) {
```
phil303: Should I stub this out for dev mode? Or is there a throwaway endpoint?

src/app/router/handlers/CommentsPage.js
```diff
@@ -88,3 +108,21 @@ export default class CommentsPage extends BaseHandler {
...
   }
 }
+
+
+function buildPageViewData(state) {
+  const { currentPage: { queryParams, urlParams } } = state.platform;
+  const fullName =`t3_${urlParams.postId}`;
+  const post = state.posts[fullName];
+
+  return omitBy({
```
schwers: Could we use `cleanObject` here ?

src/config.js
schwers: For the `env` variables, yes, I think we should be able to use the same name. Do you need the variable on the client? You might run into the webpack issue I hit ([MOBILEWEB-574](https://reddit.atlassian.net/browse/MOBILEWEB-574?jql=text%20~%20%22client%20env%22))

src/app/router/handlers/CommentsPage.js
```diff
@@ -41,23 +50,34 @@ export default class CommentsPage extends BaseHandler {
...
-      const commentsPage = state.commentsPages[commentsPageId];
-      return commentsPage && !!commentsPage.responseCode;
-    }, state => dispatch(setStatus(state.commentsPages[commentsPageId].responseCode)));
+      dispatch(subredditActions.fetchSubreddit(subredditName)),
+    ]);
+
+    dispatch(setStatus(getState().commentsPages[commentsPageId].responseCode));
+
+    // only track on the client
+    if (process.env.ENV === 'client') {
```
schwers: The duplicated env checks and inlined strings for firing screenviews seems like a bit of a code-smell. Would you be open to an abstraction of some sort for firing the tracking? A quick idea just as an example (not attached to it): `logScreenView(<state>, <optional: state->pageViewData>)`. (The second arg could be `getBasePayload` by default, hence optionality). `logScreenView` could handle checking the right environment, getting the tracker, and defining the event details

src/app/router/handlers/PostsFromSubreddit.js
```diff
@@ -27,27 +35,85 @@ export default class PostsFromSubreddit extends BaseHandler {
...
-      const postsList = state.postsLists[postsListId];
-      return postsList && !!postsList.responseCode;
-    }, state => dispatch(setStatus(state.postsLists[postsListId].responseCode)));
+  let target_id = null;
+  let target_fullname = null;
+  let listing_name = null;
+
+  // There's some unfortunate special handling here. The frontpage doesn't
+  // have a subredditName, multi's are delimited by a '+', and some "fake"
+  // subreddits have special listing names.
```
schwers: 🔕  Do you know if we need to have special handling of `r/all` filtering. e.g. `r/all-apple-android-javascript` is technically a valid subreddit listing, but is not a subreddit. NOTE: we need to add it to the 2X `isFakeSubreddit` check

src/lib/eventTracker.js
schwers: I believe 1X doesn't log anything if you don't have a tracker endpoint, but there is indeed a test endpoint I can point you to.

src/lib/eventUtils.js
```diff
@@ -0,0 +1,54 @@
...
+import isFakeSubreddit from 'lib/isFakeSubreddit';
+
+const ID_REGEX = /(?:t\d+_)?(.*)/;
+
+export function convertId(id) {
+  const unprefixedId = ID_REGEX.exec(id)[1];
+  return parseInt(unprefixedId, 36);
+}
+
+
```
schwers: nit: extra newline

src/lib/eventUtils.js
```diff
@@ -0,0 +1,54 @@
...
+    const subreddit = state.subreddits[subredditName.toLowerCase()];
+    return {
+      sr_id: convertId(subreddit.name),
+      sr_name: subreddit.uuid,
+    };
+  }
+  return {};
+}
+
+
```
schwers: nit: extra newline

src/lib/eventUtils.js
```diff
@@ -0,0 +1,54 @@
...
+export function getBasePayload(state) {
+  const { user, accounts } = state;
+  const referrer = state.platform.currentPage.referrer || '';
+  const domain = window.location.host;
```
schwers: Can you add a note that this is only to be used in the client? (As an aside --  You could also parse `config.origin` if `window.location.host` is undefined i.e. if we needed to run this on the server)

src/lib/eventUtils.js
```diff
@@ -0,0 +1,54 @@
...
+    dnt: !!window.DO_NOT_TRACK,
+    compact_view: state.compact,
+  };
+
+  if (userAccount) {
+    payload.user_name = userAccount.name;
+    payload.user_id = convertId(userAccount.id);
+  } else {
+    payload.loid = state.loid.loid;
+    payload.loid_created = state.loid.loidcreated;
```
schwers: 🔕  This will be `loidCreated` when #706 lands

schwers:
@phil303 could you turn this into a drone branch so we could all do event QA together with data?

src/app/router/handlers/CommentsPage.js
phil303: yea i think you're right. I'll refactor.

src/app/router/handlers/PostsFromSubreddit.js
phil303: Oh, I had no idea that was a thing. Thanks for the catch. I'll investigate.

src/lib/eventUtils.js
phil303: I actually separated those by two spaces on purpose. Since it's common to use one line to separate ideas inside functions this gives passing eyeballs further distinction. This has some precedent in [python-land](https://www.python.org/dev/peps/pep-0008/#blank-lines)

src/lib/eventUtils.js
phil303: I'll watch for timing. Thanks for the catch.

src/config.js
phil303: So this already works on the client via the DefinePlugin that you describe in that ticket. [Here](https://github.com/reddit/node-build/blob/master/lib/configs/DefaultProductionClientConfig.js#L31) is the client config, and [here](https://github.com/reddit/node-build/blob/master/lib/generators/plugins/SetNodeEnvironmentPlugin.js) is the relevant plugin that line points to. The todo here would be to get these values for prod/staging/etc and make sure they're set correctly in those environments.

Do you happen to know what the values should be here and also where/how I can set them (maybe they're already there in the right start.sh?). Maybe I need to talk to @gtaylor?

phil303:
@schwers Yes, I could do that. Let's chat tomorrow about it since I'm not sure the value of that over just deploying it and QA'ing it on snooguts.

src/lib/eventUtils.js
schwers: Ah, I see now. Okay cool, I'm actually in favor of this kind of logical separation -- I tend to do it as well, potentially at the past influence of other pythonistas (without even realizing it).

src/app/router/handlers/PostsFromSubreddit.js
phil303: Following up on this for anyone curious. This does require special handling as the `listing_name` becomes "all (filtered)". Is there anything else like this that you can think of?

src/config.js
gtaylor: @phil303 Here are the current staging values: https://github.com/reddit/reddit-kage-manifests/blob/master/kage-namespaces/staging-reddit-mobile/deployment-config.yml#L37

src/lib/eventTracker.js
phil303: Follow up. Talked to data, going to stub this out to do nothing for local dev instance.

phil303:
:nail_care: I believe I've addressed everything. This patch is now dependent on [this PR](https://github.com/reddit/node-build/pull/26).

phil303:
:haircut:

src/lib/isFakeSubreddit.js
```diff
@@ -1,8 +1,13 @@
...
 
-export { fakeSubs, randomSubs };
+export { FAKE_SUBS, RANDOM_SUBS };
 
 export default function isFakeSubreddit(subredditName) {
-  return !subredditName || (subredditName.indexOf('+') > -1 || fakeSubs.includes(subredditName));
+  return (
+    !subredditName ||                     // the frontpage
+    subredditName.indexOf('+') > -1 ||    // multis
+    subredditName.indexOf('-') > -1 ||    // reddit gold filtering
```
schwers: 💯 💯 💯  ty

src/app/router/handlers/UserProfile.js
```diff
@@ -1,15 +1,37 @@
...
+    ]);
+
+    const _tempState = getState();
+    logClientScreenView(buildScreenViewData(getState()), _tempState);
   }
 }
+
+function buildScreenViewData(state) {
+  const { userName: name } = state.platform.currentPage.urlParams;
+  const user = find(state.accounts, (_, k) => k.toLowerCase() === name.toLowerCase());
```
phil303: Bro, do you even lodash? `state.accounts` is an object 😛

src/Server.js
```diff
@@ -84,6 +84,7 @@ export function startServer() {
     reducers,
     reduxMiddleware,
     dispatchBeforeNavigation: async (ctx, dispatch, getState) => {
+      dispatchInitialMeta(ctx, dispatch);
```
schwers: Whoops, duplicate. I added this a while ago too :)

phil303:
:haircut: @schwers updated the commit message as we chatted about (see second note) and removed the duplicate dispatch.

schwers:
@phil303 the idea of staging was just in the spirit of using the fancy new drone stagings, its pretty easy to get setup. There's an issue with 2X right now though, so I think deploying and QA'ing on snooguts is good

schwers:
:fish:

PR #732, opened 2016-09-12
Add canonical link tags when NOT shell rendering

nramadas:
Add canonical link tags when the html is server-rendered. It only targets server renders because it assumes we don't need to worry about canonicals in non seo-bot situations.

:eyeglasses: @prashtx

### DIFF
--- a/src/lib/createCanonicalLinkFromState.js
+++ b/src/lib/createCanonicalLinkFromState.js
@@ -1,0 +1,91 @@
+import { matchRoute } from '@r/platform/navigationMiddleware';
+import values from 'lodash/values';
+
+import config from 'config';
+
+export const listingCanonical = (currentPage, state) => {
+  const { subredditName } = currentPage.urlParams;
+  const subreddit = state.subreddits[subredditName];
+
+  if (subreddit) {
+    return subreddit.url;
+  }
+
+  // we might not have a subreddit loaded because it is a multi-reddit. in that
+  // case, we're going to make our best attempt at canonicalizing the url by
+  // bulding a map of lowercased subreddit names to their canonicalized
+  // versions. this map will pull the canonical subreddit names from our "Posts"
+  // data.
+  // NOTE: if there is a very obscure subreddit in the multi-reddit, we might
+  // not have any posts loaded that are in that subreddit. meaning, our attempt
+  // at canonicalization will fail. this is something we've accepted for now.
+  const canonicalSubredditNames = values(state.posts).reduce((prev, post) => {
+    if (post.subreddit) {
+      prev[post.subreddit.toLowerCase()] = post.subreddit;
+    }
+    return prev;
+  }, {});
+
+  // split the multi-reddit and sort
+  let divider = '';
+  if (subredditName.indexOf('+') > -1) { // additive multi-reddit
+    divider = '+';
+  } else if (subredditName.indexOf('-') > -1) { // subtractive multi-reddit
+    divider = '-';
+  }
+
+  // make sure we can actually understand the multi-reddit. if it's some other
+  // format, just canonicalize it to the frontpage.
+  if (divider) {
+    const newName = subredditName
+      .split(divider)
+      .sort()
+      .map(s => canonicalSubredditNames[s.toLowerCase()] || s)
+      .join(divider);
+
+    return `/r/${newName}/`;
+  }
+
+  // desperation case. we don't know how to handle this url, so just
+  // canonicalize to the frontpage.
+  return '/';
+};
+
+export const commentsCanonical = (currentPage, state) => {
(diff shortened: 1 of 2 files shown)


src/lib/createCanonicalLinkFromState.js
```diff
@@ -0,0 +1,44 @@
...
+  ['/r/:subredditName/comments/:postId/comment/:commentId', commentsCanonical],
+  ['/r/:subredditName/comments/:postId/:postTitle/:commentId', commentsCanonical],
+  ['/r/:subredditName/comments/:postId/:postTitle?', commentsCanonical],
+  // user pages
+  ['/u/:userName/activity', userCanonical],
+  ['/u/:userName/gild', userCanonical],
+  ['/u/:userName/:savedOrHidden(saved|hidden)', userCanonical],
+  ['/u/:userName', userCanonical],
+  // catch-all
+  ['*', baseCanonical],
```
nramadas: i don't know if this is something we want to do or not. some feedback here would be appreciated.

src/lib/createCanonicalLinkFromState.js
```diff
@@ -0,0 +1,44 @@
...
+  : `${currentPage.url}/`;
+
+const CANONICAL_ROUTES = [
+  // listings
+  ['/r/:subredditName', listingCanonical],
+  ['/r/:subredditName/about', listingCanonical],
+  // comment pages
+  ['/r/:subredditName/comments/:postId/comment/:commentId', commentsCanonical],
+  ['/r/:subredditName/comments/:postId/:postTitle/:commentId', commentsCanonical],
+  ['/r/:subredditName/comments/:postId/:postTitle?', commentsCanonical],
```
prashtx: We also need to cover the cases of the above that omit `r/:subredditName`, for example `/comments/:postId/comment/:commentId` . Those will start to become redirects to the version that includes subreddit names, but in the meantime it's important for those to have the same `commentsCanonical` handling.

src/lib/createCanonicalLinkFromState.js
prashtx: I think this is helpful, since we don't redirect to the trailing slash version yet. As long as we'll always have `currentPage.url` and won't fall back to a canonical of `https://www.reddit.com/` for random other pages.

src/server/templates/main.jsx
```diff
@@ -13,6 +14,12 @@ const JS_FILE = manifest[`${CLIENT_NAME}.js`];
 const { assetPath } = config;
 
 export default function(data, store) {
+  const state = store.getState();
+
+  const canonicalLink = !state.platform.shell
+    ? createCanonicalLinkFromState(state)
```
prashtx: It looks like this will produce relative paths for the canonical. Is that accurate? If so, we should pull in the "primary" Reddit scheme and hostname, so we have canonicals of the form `https://www.reddit.com/r/AskReddit/`.

src/lib/createCanonicalLinkFromState.js
```diff
@@ -0,0 +1,44 @@
+import { matchRoute } from '@r/platform/navigationMiddleware';
+
+export const listingCanonical = (currentPage, state) => {
+  const { subredditName } = currentPage.urlParams;
+  const subreddit = state.subreddits[subredditName];
```
prashtx: I think this might break for multis.

prashtx:
💅

src/lib/createCanonicalLinkFromState.js
nramadas: could i just use `r/${subredditName}/`, or do i need to canonicalize the url in someway?

prashtx:
For multireddits, we should make a best effort attempt to pull the correct subreddit capitalization from the list of posts, and we should sort the subreddits. For multis that have many subreddits, we'll likely get the capitalization wrong for several subs, and that's OK for now.

src/lib/createCanonicalLinkFromState.js
```diff
@@ -0,0 +1,48 @@
+import { matchRoute } from '@r/platform/navigationMiddleware';
+
+import config from 'config';
+
+export const listingCanonical = (currentPage, state) => {
+  const { subredditName } = currentPage.urlParams;
+  const subreddit = state.subreddits[subredditName];
+  return subreddit ? subreddit.url : `/r/${subredditName}/`;
```
nramadas: DND until we figure out how we're going to deal with the multi reddit case

nramadas:
:haircut:

src/server/templates/main.jsx
nramadas: i've updated the function to use the `reddit` in config, which will point to `https://www.reddit.com` in production

prashtx:
:fish:

PR #736, opened 2016-09-13
Point kage deployment at a more thorough ConfigMap.

gtaylor:
Rather than commit potentially sensitive secrets, tokens, headers, and URLs to public repos, we've broken configuration for staging environments out to config maps. This PR also makes the kage reddit-mobile staging environment more closely resemble the current staging environment's configs.

Since kage uses your local configs to determine how to carry out a staging deploy, you can still change values locally (without committing them) and staging the branch. For example, let's say I wanted to change my babel cache path. Here's the current value:
```
- name: BABEL_CACHE_PATH
  valueFrom:
    configMapKeyRef:
      name: deployment-config
      key: babel.cache.path
```
Here's an arbitrarily updated, un-committed change that I'm testing:
```
- name: BABEL_CACHE_PATH
  value: /some/path
```
When you're done, checkout over your changes and you are back to normal.

:eyeglasses: @nramadas @spladug @ckwang8128

### DIFF
--- a/.kage/manifests/deployment.yml
+++ b/.kage/manifests/deployment.yml
@@ -26,22 +26,141 @@ spec:
           # have an image with that tag name pulled.
           imagePullPolicy: Always
           env:
-            - name: NAMESPACE
+            # The values for these configMap keys can be found at:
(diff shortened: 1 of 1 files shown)


spladug:
:fish:

PR #737, opened 2016-09-13
Second attempt: use server set loids

nramadas:
:eyeglasses: @schwers

Most of the error were being thrown by our health check. `loid` didn't exist because there is no user associated with it. before, since we were always setting `loid`s on the client side, it was no big deal if we were dealing with a real user or not, because "hey, just give 'em an loid". But now, with api generated loids, we need to be more careful.


schwers:
:fish:

PR #738, opened 2016-09-14
Put Varnish in front of Node server via docker-compose. (CHAN-159)

wting:
This puts a Varnish server (:4301) in front of the Node server (:4444). This
provides the ability to test custom VCL (e.g. URL rerouting) in a development
environment.

:eyeglasses: @gtaylor


Dockerfile
```diff
@@ -1,22 +0,0 @@
...
-
-ENV DEBIAN_FRONTEND noninteractive
-
-RUN echo "deb-src http://archive.ubuntu.com/ubuntu trusty main" >> /etc/apt/sources.list \
-    && apt-get update \
-    && apt-get install -y curl python2.7 git \
-    && apt-get update \
-    && apt-get upgrade -y \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
wting: FYI most of these `apt-get` steps are unnecessary because the base Ubuntu image (ubuntu/vivid) takes care of this.

From [Dockerfile best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/):

>The official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.

gtaylor:
I can look this over more tomorrow, but it's important to note that the current master branch is the old 1X deployment. The Dockerfile you'll find in there is leftover from some tinkering that was done long ago.

Most of the action happens in 2X now: https://github.com/reddit/reddit-mobile/tree/2X

You'll find a much better Dockerfile situation in there. If you need to do this in master and 2X, no problem. Just keep in mind that the two have diverged substantially, especially with the Dockerfile situation.

wting:
This needs to happen in both 1X and 2X, and I figured that it's easier to get it working in 1X and port it over to 2X afterwards.

I've updated `Dockerfile.dev` to mirror 2X's `Dockerfile.staging`, with some tweaks.

wting:
I've pushed the Docker image and updated the Dockerfile to use `quay.io/reddit/varnish-fastly`.

:haircut:

gtaylor:
Awesome! :fish:

wting:
Merged in 77dad5c963.

PR #739, opened 2016-09-14
Fix a small bug in the mobilify helper function

uzi:
👓 @schwers @phil303 @nramadas @umbrae

### DIFF
--- a/src/lib/mobilify.es6.js
+++ b/src/lib/mobilify.es6.js
@@ -1,6 +1,6 @@
 function mobilify(url, origin) {
   if (url) {
-    return url.replace(/https?:\/\/(?:np\.)?(?:www\.)?reddit.com/g, origin || '');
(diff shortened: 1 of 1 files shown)


src/lib/mobilify.es6.js
```diff
@@ -1,6 +1,6 @@
 function mobilify(url, origin) {
   if (url) {
-    return url.replace(/https?:\/\/(?:np\.)?(?:www\.)?reddit.com/g, origin || '');
+    return url.replace(/https?:\/\/(?:np\.)?(?:www\.)?reddit.com\//g, origin || '');
```
JordanMilne: nit: the `.` in `reddit.com` should be escaped as well

JordanMilne: The replacement should be `(origin || '') + '/'` or this will break links.

schwers:
:fish:

PR #741, opened 2016-09-15
Put Varnish in front of Node via docker-compose.

wting:
Same as #738 (cherry-picked it over). Verified working on my machine.

:eyeglasses: @gtaylor


gtaylor:
:fish:

PR #742, opened 2016-09-15
Get "go to desktop" working

phil303:
This works by first settings the `mweb-no-redirect` cookie on click, and then
going to the www analogue of the current url using a normal anchor.

Assumptions:

- This assumes that the onclick handler fires synchronously before the browser
  takes the user to the new location. From browser testing, this seems true.

Notes:

- Right now, we define the www url as separate than the m. url. Eventually,
  when both clients are on the www urls, we can (if we wanted to - there might
  be reasons not to) update this patch to just use the relative path.

This is semi-blocked on https://github.com/reddit/reddit-mobile/pull/730, which has some DefinePlugin updates. I could technically just add `REDDIT` to `process.env` right now but that'd result in a merge conflict. So I'll just wait.

### DIFF
--- a/src/app/components/OverlayMenu/index.jsx
+++ b/src/app/components/OverlayMenu/index.jsx
@@ -4,11 +4,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import * as overlayMenuActions from 'app/actions/overlayMenu';
-
-import {
-  OVERLAY_MENU_CSS_CLASS,
-  OVERLAY_MENU_CSS_TOP_NAV_MODIFIER,
-} from 'app/constants';
+import cx from 'lib/classNames';
 
 const T = React.PropTypes;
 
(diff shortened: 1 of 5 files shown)


src/app/components/OverlayMenu/index.jsx
```diff
@@ -4,30 +4,17 @@ import React from 'react';
...
-  let className = OVERLAY_MENU_CSS_CLASS;
-  if (!props.fullscreen) {
-    className += ` ${OVERLAY_MENU_CSS_TOP_NAV_MODIFIER}`;
-  }
-
-  return className;
-};
-
-export const OverlayMenu = (props) => (
+export const OverlayMenu = props => (
```
phil303: Just some cleanup in this file.

src/app/components/SettingsOverlayMenu/index.jsx
```diff
@@ -1,24 +1,25 @@
...
 import OverlayMenu from 'app/components/OverlayMenu';
 import { LinkRow, ButtonRow, ExpandoRow } from 'app/components/OverlayMenu/OverlayMenuRow';
-
 import menuItems from './SettingsOverlayMenuItems';
-
 import { themes } from 'app/constants';
+import { userAccountSelector } from 'app/selectors/userAccount';
+import config from 'config';
+
+// TODO: update process.env.REDDIT in defineplugin
```
schwers: 👍

schwers:
:fish: from me, I'm fine with us delaying the sort'ing url differences between mweb and desktop until later

schwers:
:fish: 🍣  🐠

PR #743, opened 2016-09-15
2 x auth stuff

schwers:
I have this as two separate commits with reasoning in their messages. But essentially this fixes the login issue and the cookie issue on staging.

👓  @nramadas || @uzi


nramadas:
:fish:

uzi:
:fish:

PR #745, opened 2016-09-16
Keep the client definePlugin fallbacks in config

phil303:
It's confusing to possibly have these types of fallbacks in two different places, so move them into config.js.

:eyeglasses: @schwers


schwers:
:fish: ❗

PR #748, opened 2016-09-19
NSFW Interstitial doesn't need a redirect

uzi:
When the interstitial is being shown, we're already
being routed to the page we want. With the code as
is, we're getting redirected back to the root.

@schwers 

🔕  For this patch, but should we be setting a cookie or the like of what the user requests?  If they refresh the app, they need to select to see NSFW stuff again.


uzi:
👓  @schwers

schwers:
:fish:

RE: the 🔕  `src/reddit/r/reddit-mobile/src/app/side-effect-components/CookieSync.js` should be taking `state.preferences.over18` out of state and writing it to a cookie, and `dispatchInitialOver18` in `src/reddit/r/reddit-mobile/src/Server.js` should be putting that into initial state. It sounds like that's been broken from what you're describing

uzi:
Oh, this is with a logged out user I've been testing with.

schwers:
closed via https://github.com/reddit/reddit-mobile/commit/8baa1e832366bd410df75312d9b7b5d40202c1e3

PR #749, opened 2016-09-19
Fix a bug in pageview tracking

phil303:
Problem:
When shell was set to false, we were server rendering. Also, since shell
was set to false, we were running a function that was building tracking
data for us. This function relied on the window object, which obviously
doesn't exist server-side, and was causing an exception.

Fix:
Because we want to make a client side check before we build this data,
I'm now passing the data building function as a callback to the
`logClientScreenView` function.

:eyeglasses: @uzi


uzi:
:fish:

schwers:
closed via https://github.com/reddit/reddit-mobile/commit/3fb798c47c7e8c59de6c37481415babf71f56a1b

PR #750, opened 2016-09-20
Pass loidcreated cookie header in UTC format.

schwers:
The api expects loidcreated to be in UTC format, passing
unix time as we've been doing makes all of our server-side api
calls fail. This only really matters for googlebot right now,
hence why we didn't notice it sooner.

Test this by making a fake device profile in chrome with user agent 'googlebot', and making server side requests.

👓  @nramadas @uzi


uzi:
:fish:

nramadas:
:fish:

PR #751, opened 2016-09-20
Rum timings refinement

uzi:
- Send timings based on the onLoad event.
- Detect if we're shell or seo rendered.
- Resolve the promise when we exit early.
- Make server and client messaging more clear.

Examples of data sent when we're shell rendered:
```
 { actionName: 'm2.server.shell',
  startTiming: 0.02200007438659668,
  requestTiming: 0.2989997863769531,
  responseTiming: 0.002000093460083008,
  domLoadingTiming: 0.09200000762939453,
  mountTiming: 0.101 }
```

And when we're rendered for seo:
```
 { actionName: 'm2.server.seo',
  requestTiming: 1.38100004196167,
  responseTiming: 0.003000020980834961,
  domLoadingTiming: 0.17599987983703613,
  mountTiming: 0.378 }
```

👓 @schwers @nramadas @phil303

### DIFF
--- a/src/Client.js
+++ b/src/Client.js
@@ -18,6 +18,16 @@ import reduxMiddleware from 'app/reduxMiddleware';
 import { sendTimings } from 'lib/timing';
 import Session from 'app/models/Session';
 
+// Bits to help in the gathering of client side timings to relay back
+// to the server
+const beginMount = Date.now();
+let isShell;
+
+window.onload = () => {
+  const endMount = Date.now();
+  sendTimings(beginMount, endMount, isShell);
+}
+
 // register window.onError asap so we can catch errors in the client's init
 window.onerror = (message, url, line, column) => {
   errorLog({
@@ -46,7 +56,6 @@ window.onunhandledrejection = rejection => {
 };
 
 // start the app now
-const beginRender = Date.now();
 const client = Client({
   routes,
   reducers,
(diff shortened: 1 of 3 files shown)


src/Client.js
```diff
@@ -18,6 +18,15 @@ import reduxMiddleware from 'app/reduxMiddleware';
 import { sendTimings } from 'lib/timing';
 import Session from 'app/models/Session';
 
+// Bits to help in the gathering of client side timings to relay back
+// to the server
+const beginRender = Date.now();
```
nramadas: let's move this to the top, then rename the timing metric something akin to 'scriptParseTime' or something like that (i'm not creative)

src/lib/timing.js
```diff
@@ -46,12 +46,17 @@ export function getTimes() {
   return timings;
 }
 
-export function sendTimings(beginRender) {
+export function sendTimings(beginRender, isShell) {
```
nramadas: start + end times :)

src/Client.js
uzi: Spoke with @schwers and the problem is that we don't know that that Date.now() will be run at the head of the script.  We've decided to leave this as mountTiming for now and in place, and revisit later.

src/lib/timing.js
uzi: Made this change

schwers:
:fish:

PR #753, opened 2016-09-22
Support /user/* routes

uzi:
I've implemented this as additional valid urls in order to get this in and supported in a timely fashion.  The real thing we'll want to do of course is have the /u/* variants be 301 redirects to the /user/* ones, but this is a start.

I threw in a patch to change many of the /u/* links in the system to /user/* ones.  I'm not sure if we want any of these, but I at least wanted to call it out.

Also tossed in a small cosmetic issue while I was in the neighborhood.

👓 @schwers @phil303 @nramadas @umbrae

### DIFF
--- a/src/app/components/UserProfileSummary/index.jsx
+++ b/src/app/components/UserProfileSummary/index.jsx
@@ -59,7 +59,7 @@ const GoldInfo = props => {
   }
 
   return (
-    <Anchor href={ `/u/${user.name}/gild` }>
+    <Anchor href={ `/user/${user.name}/gild` }>
       <UserProfileRow>
         <UserProfileBadge
           text={ `Give ${user.name} gold` }
--- a/src/app/components/UserProfileSummary/index.jsx
+++ b/src/app/components/UserProfileSummary/index.jsx
@@ -60,12 +60,14 @@ const GoldInfo = props => {
 
   return (
     <Anchor href={ `/u/${user.name}/gild` }>
-      <UserProfileBadge
-        text={ `Give ${user.name} gold` }
-        subtext='show your appreciation'
-      >
-        <UserProfileBadgeIcon iconName='gold-snoo' color='gold' />
-      </UserProfileBadge>
+      <UserProfileRow>
+        <UserProfileBadge
+          text={ `Give ${user.name} gold` }
+          subtext='show your appreciation'
+        >
+          <UserProfileBadgeIcon iconName='gold-snoo' color='gold' />
+        </UserProfileBadge>
+      </UserProfileRow>
     </Anchor>
   );
 };
--- a/src/app/router/handlers/UserReroute.js
+++ b/src/app/router/handlers/UserReroute.js
@@ -1,0 +1,13 @@
+import { setPage } from '@r/platform/actions';
+import { BaseHandler, METHODS } from '@r/platform/router';
+
+export default class UserReroute extends BaseHandler {
+  async [METHODS.GET](dispatch, getState) {
+    const { platform: { currentPage }} = getState();
+    const { urlParams, queryParams, hashParams, referrer } = currentPage;
+    let { url } = currentPage;
+
+    url = url.replace('/u/', '/user/');
+    dispatch(setPage(url, { urlParams, queryParams, hashParams, referrer }));
+  }
+}
--- a/src/app/components/Comment/CommentDropdown/index.jsx
+++ b/src/app/components/Comment/CommentDropdown/index.jsx
@@ -32,7 +32,7 @@ export default function CommentDropdown(props) {
         ? <DropdownRow icon='save' text={ isSaved ? 'Saved' : 'Save' } isSelected={ isSaved } onClick={ onToggleSave }/>
         : null }
       { !userIsAuthor
-        ? <DropdownLinkRow href={ `/u/${commentAuthor}` } icon='user-account' text={ `${commentAuthor}'s profile` }/>
+        ? <DropdownLinkRow href={ `/user/${commentAuthor}` } icon='user-account' text={ `${commentAuthor}'s profile` }/>
         : null }
       { username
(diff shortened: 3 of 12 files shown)


uzi:
👓 @prashtx @wting

umbrae:
As a noob, I'm surprised it's considered hard to introduce a 301 from the `/u/` pages.

umbrae:
Unrelated, but in digging into this I noticed that our canonical tags are pointing to snooguts when googlebot rendering:

image

Is that known?

uzi:
You too, huh?  Yeah, I don't think it would be hard to do something on client side to forward them from /u/ to /user/, but that wouldn't generate a 301... but do we care about the client side doing it, or just the server side if it's an initial render?

Of course maybe it is easy and I'm overlooking something -- I'm just trying to do something quick in the meantime and get this in now so we at least have the routes and then look at fixing it at an appropriate time.

For the canonical stuff, it looks like we're making use of the config.reddit variable which is defined as

```const reddit = process.env.REDDIT || 'https://www.reddit.com';```

I imagine that'll be configured to m.reddit.com when in production... so that's also wrong. Easy fix and we'll take care of it today. Great find!

**Comment by umbrae** [comment]
If I'm understanding how the router works, it looks like issuing a 301 should be pretty easy unless I'm missing something.

Couldn't you add a route here: https://github.com/reddit/reddit-mobile/blob/2X/src/app/router/index.js#L58

That matched all `/u/` URLs and returned a 301 status like the 404 handler does?

**Comment by uzi** [comment]
That's a good point, but the 404 only sets a status -- there's no specified redirect url. Buuuuuut... I guess I didn't noticed that there's a reroutePage as well which is totally not used in 2X yet:

https://github.com/reddit/node-platform/blob/master/lib/modules/actions.js#L44

I'll take a look at that in a bit.

**Comment by wting** [comment]
I wouldn't worry about the /u/ 301s if it's not a quick fix.

Longer term, it might be easier to do 301s _only_ in Fastly than maintaining two separate routing rules between www and m web.

**Comment by uzi** [comment]
💇  Ok friends -- I made a Reroute handler which can do the heavy lifting here, and I've made it generalized enough that we can use it in more cases.

The question remains if we want to do all of the extra /u/ -> /user/ changes in the third commit, so let me know your thoughts.  Thanks!

File: src/app/router/handlers/Reroute.js [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+import { setPage, reroutePage } from '@r/platform/actions';
+
+const urlTransform = (url) => {
+  if (url.startsWith('/u/')) {
+    return url.replace('/u/', '/user/');
+  }
+  throw new Error('Redirect failure with route "`${url}`"');
+};
+
+export default class Reroute extends BaseHandler {
```
phil303: ok now i understand what your concern was when we were chatting yesterday. I didn't realize this was called `Reroute`, instead of say, `UserReroute`. If you're goal is to make this generic, this can be done differently.

Some pseudo code:
```js
import redirect from 'lib/redirect';

export default class UserReroute extends BaseHandler {
  async [METHODS.GET](dispatch, getState) {
    const { currentPage } = getState()
    const newUrl = urlTransform(currentPage.url);
    redirect(newUrl, currentPage, { status: 301 });
  }
}
```

```js
function redirect(newUrl, currentPage, options) {
  dispatch(setPage(newUrl, currentPage));
  dispatch(reroutePage());
}
```

TBH, I'm not even sure you need to be this generic but it does remove some places for mistakes to have this be in a function.

phil303: Being specific about this handler's name also allows you to remove the checks in `urlTransform`.

uzi:
💇

src/app/components/UserProfileSummary/index.jsx
```diff
@@ -59,13 +59,15 @@ const GoldInfo = props => {
...
   return (
-    <Anchor href={ `/u/${user.name}/gild` }>
-      <UserProfileBadge
-        text={ `Give ${user.name} gold` }
-        subtext='show your appreciation'
-      >
-        <UserProfileBadgeIcon iconName='gold-snoo' color='gold' />
-      </UserProfileBadge>
+    <Anchor href={ `/user/${user.name}/gild` }>
+      <UserProfileRow>
```
phil303: is this the cosmetic change you mentioned in your description?

src/app/router/handlers/UserReroute.js
```diff
@@ -0,0 +1,17 @@
+import { BaseHandler, METHODS } from '@r/platform/router';
+import redirect from 'lib/redirect';
+
+export default class UserReroute extends BaseHandler {
+  async [METHODS.GET](dispatch, getState) {
+    const { platform: { currentPage }} = getState();
+
+    let { url } = currentPage;
+    if (url.startsWith('/u/')) {
```
phil303: why do we need this check? Isn't this guaranteed now?

src/lib/redirect.js
```diff
@@ -0,0 +1,10 @@
+import { setPage, reroutePage } from '@r/platform/actions';
+
+function redirect(url, currentPage, dispatch) {
+  // Helper for redirecting routes to other routes
```
phil303: :no_bell: total nit, should this be above the function?

phil303:
small comments, otherwise :fish:

src/app/components/UserProfileSummary/index.jsx
uzi: Yes, the second commit is cosmetic, the third is changing links

src/app/router/handlers/UserReroute.js
uzi: It should be -- I can change it

src/lib/redirect.js
uzi: I'm used to it being inside the function at the top ... python docstring style (or with what you do with most languages, really)

PR #754, opened 2016-09-22
Gather and send handler timings to the timings endpoint

uzi:
We add an optional third argument, an object, to the
routes where we can specify arbitrary route metadata
information. In this case, a "name" key specifies the
name of the route for capturing timing info on the
duration of a handler.

Hook into this new feature of @r/platform, prepare
and validate the data and send it down.

NOTE: This will require a small patch to hivemind as well.
Also, I'm not sold on the "apiRequestTiming" name for
the timing (which I was planning on mapping to "api" in
graphite), so please chime in with better ideas.  I just didn't
want to put the word "handler" in there in case it's a thing
that eventually goes away.

👓 @schwers @phil303 @nramadas


src/Client.js
```diff
@@ -131,6 +131,9 @@ const client = Client({
   },
   appComponent: <App/>,
   debug: (process.env.NODE_ENV || 'production') !== 'production',
+  onHandlerComplete: (data) => {
```
schwers: tinsiest nit: you don't need the parens here

src/lib/timing.js
```diff
@@ -46,8 +48,18 @@ export function getTimes() {
...
+    .timeout(DEFAULT_API_TIMEOUT)
+    .send({
+      rum: timings,
+    })
+    .then();
+}
+
 export function sendTimings(beginMount, endMount, isShell) {
-  if (Math.random() > 0.1) { // 10% of requests
+  if (Math.random() > CLIENT_PERCENTAGE) {
```
schwers: 🔕  Should we always compute the random here, only logging 10% of every pageview, or should we compute it once and only log timings for 10% of sessions?

src/app/router/index.js
```diff
@@ -21,24 +21,44 @@ import { PostSubmitHandler, PostSubmitCommunityHandler } from './handlers/PostSu
...
-  ['/', PostsFromSubredditHandler],
-  ['/r/:subredditName', PostsFromSubredditHandler],
-  ['/u/:user/m/:multi', PostsFromSubredditHandler],
-  ['/r/:subredditName/comments/:postId/comment/:commentId', CommentsPageHandler],
-  ['/r/:subredditName/comments/:postId/:postTitle/:commentId', CommentsPageHandler],
-  ['/r/:subredditName/comments/:postId/:postTitle?', CommentsPageHandler],
+  ['/', PostsFromSubredditHandler, { name: 'index' }],
+  ['/r/:subredditName', PostsFromSubredditHandler, { name: 'listing' }],
+  ['/u/:user/m/:multi', PostsFromSubredditHandler, { name: 'listing' }],
+  [
```
phil303: How long is this line with it all one line? sometimes its ok to go over the 80 character limit and in this case, I think it'd be preferable to be able to fit it into one line.

src/lib/timing.js
```diff
@@ -46,8 +48,18 @@ export function getTimes() {
...
 }
 
+export function makeTimingsRequest(timings) {
+  makeRequest
+    .post('/timings')
+    .timeout(DEFAULT_API_TIMEOUT)
+    .send({
+      rum: timings,
+    })
+    .then();
```
phil303: why the empty then?

src/server/meta/index.js
```diff
@@ -4,16 +4,24 @@ import superagent from 'superagent';
 
 import { formatLogJSON } from 'lib/errorLog';
 import { DEFAULT_API_TIMEOUT } from 'app/constants';
+import routes from 'app/router';
 
+// Generate the client actionNames from the routes
+const mClientName = name => `m2.client.${name}`;
+const routeNames = routes.filter(r => r[2] && r[2].name).map(n => n[2].name);
```
phil303: nit, why switch to `n` instead of keeping the `r`?

phil303: :no_bell: I think these can both be considered constants

phil303: How about:
```js
const SERVER_NAMES = new Set(['shell', 'seo']).map(name => `m2.client.${name}`)
```

src/lib/timing.js
schwers: How much do we care about logging request failures here? If it's okay for this to fail we should add a `.catch(() => {})`

src/lib/timing.js
uzi: That's only going to get triggered with the onLoad event -- so only 10% of every initial load.

src/lib/timing.js
uzi: Good question -- this is 1X code that was copied over.

src/server/meta/index.js
```diff
@@ -4,16 +4,24 @@ import superagent from 'superagent';
...
+const CLIENT_NAMES = new Set(routeNames.map(mClientName));
+
+// Generate the server actionNames
 const mServerName = name => `m2.server.${name}`;
-const ALLOWED_ACTION_NAMES = new Set([
-  'shell',
-  'seo',
-].map(mServerName));
+const SERVER_NAMES = new Set(['shell', 'seo'].map(mServerName));
```
phil303: Since there's an added (albeit most likely trivial) cost to creating the array and then the sets, do these need to be sets?

src/server/meta/index.js
uzi: Sure, I'll do that

src/app/router/index.js
schwers: looks like its 103 with it, just over our warning limit. That being said, it doesn't seem too terrible being a little over, but I'm fine either way

src/lib/timing.js
uzi: Sure, I'll add that.  As it is, that endpoint always returns a 200 or 204, I think.

src/server/meta/index.js
uzi: It does for deduping in the end, but it doesn't have to be in the initial construction unless something is going to use CLIENT_NAMES or SERVER_NAMES later.

src/lib/timing.js
schwers: This is through the `makeRequest` wrapper, it looks like you need to call `.then()` to call `superagent.Request::end` so the request is kicked off. I think the idea here is you can do arbitrary method chaining to set headers, timeouts, and etc in any order and then `.then()` when you want to really kick off the request

phil303: just an fyi, timeouts seem to trigger the catch statement: https://visionmedia.github.io/superagent/#request-timeouts

src/lib/timing.js
schwers: Sure, for now that's the case. I left it nobell just because this seems like something that we may want to revisit in the near future where each route could be logigng

src/lib/timing.js
```diff
@@ -46,8 +48,19 @@ export function getTimes() {
...
 
+export function makeTimingsRequest(timings) {
+  makeRequest
+    .post('/timings')
+    .timeout(DEFAULT_API_TIMEOUT)
+    .send({
+      rum: timings,
+    })
+    .then()
+    .catch(() => {});
```
phil303: comment here on why this is so kooky looking

src/server/meta/index.js
phil303: yea that's my thinking, since it doesn't look like this is being used in other places. Up to you.

src/lib/timing.js
uzi: Is the comment I've added good enough?

src/lib/timing.js
uzi: Not with sendTimings... it's the `onHandlerCompleteTimings` where we might do that... and I don't think it's terrible to have a 10% sample size of that.

src/server/meta/index.js
uzi: I'd rather play it safe with those variables and leave as is.

src/Client.js
```diff
@@ -131,6 +131,9 @@ const client = Client({
   },
   appComponent: <App/>,
   debug: (process.env.NODE_ENV || 'production') !== 'production',
+  onHandlerComplete: data => {
+    onHandlerCompleteTimings(data);
```
phil303: this can just be `onHandlerComplete: onHandlerCompleteTimings,`

phil303:
:fish:

src/lib/timing.js
schwers: Oh I see, you're right this is the wrong line, just similar.  Howerver part of me still thinks its weird that we even do this randomization, I feel like we should be able to just log all the timings and have more data overall.

src/Client.js
phil303: NEWB

schwers:
🐠  🍹

uzi:
bf8940c6931454d6eadb90dc33431894559e40ba

PR #756, opened 2016-09-22
Error toaster

phil303:
Feature:
Generic error toaster that can pop up upon error conditions, whether that be user based or exception based.

Implementation:
The goal here was flexibility. The toaster currently only works for posting validation errors/exceptions but it can easily work for any other error flow. All an author has to do is add their action case to the toaster reducer and also specify a message as part of their action.

Additionally, this can be used for more than just errors. The code and CSS was written to give authors that leeway should they need some other kind of (read-only) popup toaster.

Additional details:

- This adds `react-motion` as a dependency. [React-motion](https://github.com/chenglou/react-motion) (MIT license) is the defacto standard for adding animations to React code. It's a fairly light library and the code quality is solid.

### How it looks
animation

:eyeglasses: @schwers @uzi


phil303:
👓 @schwers @uzi

src/app/components/Toaster/index.jsx
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable react/jsx-curly-spacing */
```
phil303: deals with lines 21 through 25

umbrae:
image

src/app/components/Toaster/index.jsx
uzi: Was there no other way?  I'm surprised we need this here.

src/app/components/Toaster/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+  );
+}
+
+function chooseBody(props) {
+  switch (props.type) {
+    case TYPES.ERROR:
+      return renderError(props);
+    default:
+      return null;
+  }
```
uzi: Doesn't really need to be a switch...

`return (props.type == TYPES.ERROR) ? renderError(props) : null;`

src/app/components/Toaster/index.jsx
phil303: I could have put the body of that code into another function but it'd be a super thin function and you'd only lose clarity.

src/app/components/Toaster/index.jsx
phil303: yep, it doesn't. The goal of that is to allow new toaster types to be easily added and to make it obvious how to do so.

src/app/reducers/toaster.js
```diff
@@ -0,0 +1,45 @@
...
+      });
+    }
+
+    case postingActions.FAILURE: {
+      return merge(state, {
+        isOpen: true,
+        type: toasterActions.TYPES.ERROR,
+        message: GENERIC_ERROR,
+      });
+    }
```
uzi: 🔕  You could merge postingActions.VALIDATION_FAILURE and postingActions.FAILURE, and put a

`const message = (action.type === postingActions.VALIDATION_FAILURE) : action.message : GENERIC_ERROR;`

src/app/actions/toaster.js
```diff
@@ -0,0 +1,7 @@
+export const TYPES = {
+  ERROR: 'ERROR',
+};
+
+export const CLOSE = 'TOASTER__CLOSE';
+
+export const closeToaster = () => ({ type: CLOSE });
```
uzi: Maybe comment here why there's only a close?  "We only need the generic close action as it's opened by several external actions..."?

uzi:
💅  Looks good!  Just a few nits from my end.

src/app/reducers/toaster.js
phil303: So the goal of this patch is to provide the first layer for a generic toaster. I can imagine many situations this toaster would be useful for, and yes while I can do that, it artificially constraints the flexibility I'm trying to build into this code. For instance, this check immediately fails should I add a comment failure case.

src/app/actions/toaster.js
phil303: This is how redux is supposed to work. It feels overly specific to include it in this one place.

src/app/reducers/toaster.js
schwers: I agree with Phil, we should keep this as flexible as possible. However I do like the idea of actions conforming to interfaces (this case makes me want to try adding TypeScript support again, as we could define interfaces and specify that certain action objects must conform to them). In this case, if you did want to merge them together, I would propose an interface for actions that trigger the toaster

```ts
interface OpensToaster {
    toasterType?: string; // you can specify a kind of toaster, reducer defaults to 'toasterActions.TYPES.ERROR`
    message?: string; // you can specify a message for toaster, reducer defaults to `GENERIC_ERROR`
}
```

Then your reducer could look like

```javascript
case postingActions.VALIDATION_FAILURE:
case postingActions.FAILURE: 
case commentActions.FAILURE: 
case login.SUCCESS: {
  return merge(state, {
    isOpen: true,
    type: action.toasterType || toasterActions.TYPES.ERROR,
    message: action.message || GENERIC_ERROR,
  });
}
```

I'm sure there be some cases were you want to handle an action in its own `case` statement if it breaks the mold, but to me a scheme a like this makes it really easy to add new toaster actions.

Regardless I think what Phil has is fine for now.

src/app/actions/toaster.js
schwers: I think it's fine to omit a comment here. However it illustrates how its not entirely obvious how the app works without digging into the implementation of reducers. Maybe reducers could have a top-of-file-comment of "what state is the reducer supposed to handle, and how do you change it". This could provide new devs with some docs about how all of the little parts of the app work. In this case it would be really clear, there are some actions that the toaster reducer listens to open the toaster, and when you navigate pages or fire the close action, it closes.

schwers:
:fish: I think this patch is a welcome addition to 2X. There is some cleanup we could do on the open actions, but it sounds like I've encouraged you to think about it more. I think it's fine to land now so I'll leave the :fish: and let you decide when to change that.

src/app/reducers/toaster.js
phil303: @schwers @uzi fyi, decided to combine these together

PR #757, opened 2016-09-22
Send timings data on all visits

uzi:
Since we'll do a smaller launch of 2X initially,
we'll get numbers for all visitors instead of 10%
for now and scale back later if needed.

👓  @schwers @umbrae


umbrae:
🐟  🙇

schwers:
The only problem with this commit, is that I can't leave a :100: on the commit message :( :(

:fish:

PR #759, opened 2016-09-22
Don't pass loidCreated header if we don't have it.

schwers:
It's possible to have loid and not loidCreated cookies
from inbound requests. In fact, this will always be the
case when we have server-signed loids. This patch fixes
an exception when we try to call `toISOString()` on an
invalid date, that was being created because we passed the
default empty string value for `loidCreated` to the Date
constructor. In general, we just shouldn't pass the
cookie header if we don't really have a loidCreated anyway.

👓  @uzi


uzi:
:fish:

PR #760, opened 2016-09-23
Various comments page cleanups

phil303:
- Instead of a horizontal line dividing comments, have a vertical line
  encapsulating comment branches. This is aligns with how native clients
  show comments.
- Fix the over-indendation of the "more comments" text.
- Make the collapse caron animate as it rotates from down to right.
- Make the caron for "more comments" the same size as the rest of the
  comment carons.

### Old comments
screen shot 2016-09-22 at 5 07 24 pm

### New comments
screen shot 2016-09-22 at 5 08 44 pm

:eyeglasses: @schwers @uzi @dhunten

### DIFF
--- a/src/app/components/Comment/styles.less
+++ b/src/app/components/Comment/styles.less
@@ -6,7 +6,6 @@
   &.in-comment-tree {
     .themeify({
       background-color: @theme-comment-body-color;
-      border-top: @theme-border;
     });
   }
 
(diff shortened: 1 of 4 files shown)


src/app/components/Comment/styles.less
```diff
@@ -59,10 +58,16 @@
...
+    padding-left: 2 * @grid-size;
+
+    .themeify({
+      border-left: @theme-border;
+    });
 
     &.m-no-indent {
-      margin-left: 0;
+      padding-left: 0;
+      // can't get the specificity high enough to override this otherwise
```
schwers: 🔕  Does this work without `important` if you wrap it in `@themeify({})`?

schwers:
:fish:

src/app/components/Comment/styles.less
phil303: oh yes, that probably would work. I hadn't even thought of that.

PR #761, opened 2016-09-23
Update config and respect server headers

schwers:
This adds support for the env variables `API_HEADERS` and`API_PASS_THROUGH_HEADERS`. The former is a list of default key/value pairs to send as headers on server-side api requests. The latter is a list of headers to copy from the client request onto server-side api requests.

I included an update for the root cookie domain update we wanted to make. I can take it out if we want to get internal testing going asap, but if we like it and want to include it we'll have to get this puppet patch landed too: https://github.com/reddit/puppet/compare/master...schwers:2X-root-cookie-domain?expand=1 

👓  @phil303 @uzi

### DIFF
--- a/src/app/reducers/apiRequestHeaders.js
+++ b/src/app/reducers/apiRequestHeaders.js
@@ -1,0 +1,18 @@
+// We need to pass along some headers on all api requests made on the server
+// We store them here in a dictionary of header name to value so that
+// `apiOptionsFromState` can add them to the api request options
+
+import * as apiRequestHeadersActions from 'app/actions/apiRequestHeaders';
+
+export const DEFAULT = {};
+
+export default function(state=DEFAULT, action={}) {
+  switch (action.type) {
+    case apiRequestHeadersActions.SET: {
+      const { headers } = action;
+      return headers;
+    }
+
+    default: return state;
+  }
+}
--- a/src/config.js
+++ b/src/config.js
@@ -1,9 +1,29 @@
 /*eslint max-len: 0*/
 
-// This configuration is shared with the client. Any hidden, server-only config
-// belongs in ./server instead.
+// This config is shared with the client -- but the client will only see ENV
+// variables that are defined in `blueprints.config.js` -- so be sure to have
+// sane defaults
 
 import localStorageAvailable from 'lib/localStorageAvailable';
+
+
+// takes a ';' separated string of key value pairs and turns them into an
+// object of { key -> value}
+const parseSemiColonKeyValues = (list='') => {
+  return list.split(';').reduce((obj, pair) => {
+    // check that we have a key value pair with an '=' in the middle.
+    // the '=' can't be the first character because there'd be no key
+    if (pair && pair.indexOf('=') > 0) {
+      const [key, value] = pair.split('=');
+      obj[key.trim()] = value.trim();
+    }
+
+    return obj;
+  }, {});
+};
+
+// takes a ';' separated string and returns a list
+const parseSemiColonList = (list='') => list.split(';');
 
 const reddit = process.env.REDDIT || 'https://www.reddit.com';
 
@@ -27,6 +47,9 @@ const config = () => ({
   postErrorURL: '/error',
 
   minifyAssets: process.env.MINIFY_ASSETS === 'true',
+
+  apiHeaders: parseSemiColonKeyValues(process.env.API_HEADERS),
+  apiPassThroughHeaders: parseSemiColonList(process.env.API_PASS_THROUGH_HEADERS),
 
   assetPath: process.env.STATIC_BASE || '',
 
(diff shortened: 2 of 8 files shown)


src/app/reducers/apiRequestHeaders.js
```diff
@@ -0,0 +1,16 @@
+// We need to pass along some headers on all api requests made on the server
+// We store them here in a dictionary of header name to value so that `apiOptionsFromState` can add them to the api request options
```
phil303: nit, you probably need a newline here

schwers: Thanks! Done.

src/config.js
```diff
@@ -1,23 +1,33 @@
...
+// object of { key -> value}
+const parseObject = list => {
+  if (!list) { return {}; }
+  const obj = {};
 
-// Get the last two parts if the domain has multiple subdmaoins
-const rootReddit = redditDomainParts.length < 2
-  ? redditDomainParts.join('.')
-  : redditDomainParts.splice(redditDomainParts.length - 2, 2).join('.');
+  list.split(';').forEach(function (l) {
```
phil303: :no_bell: should we generally default to es6 style callbacks? Also, should we use `reduce` here?

src/config.js
```diff
@@ -1,23 +1,33 @@
...
+    if (l && l.indexOf('=') > 0) {
```
phil303: I think you want `>=` right?

phil303: isn't there always going to be an `l`? Even if it's an empty string, this code should still work

src/config.js
```diff
@@ -1,23 +1,33 @@
...
+      const split = l.split('=');
```
phil303: I believe you can destructure this
```js
const [key, val] = l.split('=');
```

src/config.js
```diff
@@ -1,23 +1,33 @@
...
 
-// NOTE: It's very important that this is the root domain and not any
-// subdomain. Used for setting cookies, could cause issues like
-// losing authentication or infinite redirect loops if it doesn't work.
-const redditDomainParts = reddit
-  .match(/^https?:\/\/([^\/]+)/)[1]
-  .split('.');
+// takes a ';' separated string of key value pairs and turns them into an
+// object of { key -> value}
+const parseObject = list => {
```
phil303: can we call this `parseHeaders` since it seems really specific to that use case?

src/config.js
```diff
@@ -1,23 +1,33 @@
...
+      const split = l.split('=');
+      obj[split[0].trim()] = split[1].trim();
+    }
+  });
+
+  return obj;
+};
+
+// takes a ';' separated string and returns a list
+const parseList = list => {
```
phil303:
```js
const parseList = (list=[]) => list.split(';');
```

src/config.js
phil303: you can default list to [] and remove the `if list` check

src/config.js
schwers: We want to make sure there's at least one letter on the left side of the `=` sign, I can add a comment

src/config.js
```diff
@@ -37,8 +50,10 @@ const config = () => ({
   nonAuthAPIOrigin: process.env.NON_AUTH_API_ORIGIN || 'https://www.reddit.com',
   authAPIOrigin: process.env.AUTH_API_ORIGIN || 'https://oauth.reddit.com',
 
-  reddit,
-  rootReddit,
+  reddit: process.env.REDDIT || 'https://www.reddit.com',
+  // NOTE:: it's import that this is configured properly so we don't lose
```
phil303: 👍 one nit, D-D-DOUBLE COLONS

src/config.js
schwers: if you don't pass anything that would try to call 'split' on an array, which will fail

phil303:
general ask. you might do this already - when you squash this down, can you add your PR description to the description of the commit?

src/config.js
schwers: I feel like that would become confusing because then we'll have `parseList` and `parseHeaders` both being used to `apiHeaders` and `apiPassThroughHeaders`. e.g. why don't they both use `parseHeaders` if they're 'headers'

src/config.js
phil303: also "infinite"

src/config.js
schwers: using `''` 👍

phil303:
:nail_care:

uzi:
Looks good to me after addressing @phil303's feedback

src/config.js
schwers: Reverted back to `reddit` and `rootReddit` vars and comments. Thankfully they have the right spelling :)

schwers:
💇

src/config.js
```diff
@@ -1,10 +1,30 @@
...
+// This config is shared with the client -- but the client will only see ENV
+// variables that are defined in `blueprints.config.js` -- so be sure to have
+// sane defaults
 
 import localStorageAvailable from 'lib/localStorageAvailable';
 
+
+// takes a ';' separated string of key value pairs and turns them into an
+// object of { key -> value}
+const parseSemiColonKeyValues = (list='') => {
```
phil303: :heart: thanks

schwers: <3 Thank you for convincing me

phil303:
:fish:

PR #762, opened 2016-09-23
Allow self posts to have empty text fields

phil303:
Problem:
Currently, self posts require both the title and the text field to be
filled. That's not an actual requirement and in fact, some subreddits,
like AskReddit require the text field to be empty.

Fix:
On self posts, just check for the subreddit and the title. For link
posts, continue to check for subreddit, title, and the link.

:eyeglasses: @schwers @uzi


uzi:
:fish:

PR #763, opened 2016-09-23
Don't parse loidCreated or pass it to Date constructor

schwers:
The value we get is from cookies, by default a string.
We don't have to send the ISO string anymore so there's
no reason to parse it, and circumvents an issue when
we tried to call the Date constructor on a unixtimestamp
string (it wants integers).

👓  @uzi


uzi:
:fish:

PR #764, opened 2016-09-26
Login - fix validation issue

phil303:
This was failing because a real error object was expected by the
handling code. To fix it, I throw a ValidationError and check for that
type in the handling code. If it's that type, then handle gracefully.

NOTE: this is fairly hacky for the time being as we need this up asap.

:eyeglasses: @schwers @nramadas @uzi

### DIFF
--- a/src/app/router/handlers/Login.js
+++ b/src/app/router/handlers/Login.js
@@ -1,5 +1,6 @@
 import { BaseHandler, METHODS } from '@r/platform/router';
 import * as platformActions from '@r/platform/actions';
+import { errors } from '@r/api-client';
 
 import Session from 'app/models/Session';
 import * as sessionActions from 'app/actions/session';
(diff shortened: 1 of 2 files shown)


phil303:
:haircut: @schwers

src/app/router/handlers/Login.js
```diff
@@ -22,10 +23,12 @@ export default class Login extends BaseHandler {
...
 
     } catch (e) {
-      const error = JSON.parse(e);
-      dispatch(sessionActions.sessionError(error.error));
-
-      return; // do nothing until session is better
+      if (e instanceof errors.ValidationError) {
+        dispatch(sessionActions.sessionError('ERROR NOT USED'));
+      } else {
+        // TODO: swallow it for now since we don't have good error handling
```
schwers: How confident are you in the `e` being an instance of `ResponseError`? Maybe we could rethrow in that case to have an idea of when something's wrong with the api

schwers:
I like this patch's improvement, thanks Phil! Could you also take a look at `src/server/session/loginproxy.js`? If you enter a wrong password that throws a server error causing the request to 404, even though the code looks like it's trying to throw a 401 💅

src/app/router/handlers/Login.js
phil303: That's a good point. I'm not confident but like we discussed, I'm not sure what else it could be at that point.

phil303:
@schwers I think that's a deeper issue - possibly with platform. For now, I'd like to get this out (after I make the `ResponseError` change) and then come back to it.

phil303:
:haircut:

schwers:
@phil303 okay, as long as we can try to get to it soon. FWIW it looks like the code in `reddit-mobile` just isn't using koa correctly, the modmail version `src/server/session/loginproxy.js` looks to be setting and error and returning json correclty

schwers:
:fish:

PR #766, opened 2016-09-26
Put timings messages on one line in the log

uzi:
This lets us parse out the timings as well as
pase them in general.  In addition, the client
doesn't need to know.

👓  @schwers


schwers:
:fish:

PR #768, opened 2016-09-26
Fix a few of event tracking errors

phil303:
- Profile page - handle an error where `name` is undefined. Occurs when
  someone hits a profile page for a non-existent user.
- Comments page - handle an error where the post is undefined. Haven't
  seen yet but it's probably possible so being defensive here.
- Listings page - handle an error where the listing page is undefined.
  Occurs when a user doesn't have access to that subreddit.

Details:
If the important info is not present, return `null` in its stead. When
we're about to make the tracking call, check for `null` and if it
exists, don't make the call.

There might be more of these out there but I'll start with this as a foundation. 
:eyeglasses: @schwers @uzi @nramadas

### DIFF
--- a/src/app/router/handlers/CommentsPage.js
+++ b/src/app/router/handlers/CommentsPage.js
@@ -104,6 +104,10 @@ function buildScreenViewData(state) {
   const fullName =`t3_${urlParams.postId}`;
   const post = state.posts[fullName];
 
+  if (!post) {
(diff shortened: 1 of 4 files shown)


phil303:
:eyeglasses: @schwers @uzi @nramadas

nramadas:
:fish:

PR #769, opened 2016-09-26
Stop loginproxy 401 from causing errors

schwers:
👓  @phil303 similar to your patch, this will fix the server from throwing errors when the user enters a wrong password. The errors we get back from node-private seem to always be 'WRONG_PASSWORD' so I'm logging the error when it's anything _but_ that.

### DIFF
--- a/src/server/meta/stats.js
+++ b/src/server/meta/stats.js
@@ -40,7 +40,7 @@ export default router => {
       errorLog({
         error,
         requestUrl: ctx.request.url,
-        userAgent: 'SERVER',
+        userAgent: ctx.headers['user-agent'],
       }, {
         hivemind: config.statsURL,
       });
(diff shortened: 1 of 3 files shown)


src/server/meta/stats.js
```diff
@@ -40,7 +40,7 @@ export default router => {
       errorLog({
         error,
         requestUrl: ctx.request.url,
-        userAgent: 'SERVER',
+        userAgent: ctx.headers['user-agent'],
```
phil303: :no_bell: reason for this change?

phil303:
:fish:

src/server/meta/stats.js
schwers: This is just fixing an unintentional bug when I changed around `errorLog` -- `userAgent` is supposed to be the `userAgent` of the device making a request. (`errorLog` used to rely on this userAgent field having the environment prefixed onto the `userAgent` key, but it does that on it's own with `process.env.ENV` now)

PR #770, opened 2016-09-26
A little more timings endpoint hardening.

uzi:
Detect the cases where timings aren't sent or
we're missing an actionName.

👓  @schwers


schwers:
:fish:

PR #772, opened 2016-09-27
Fix error handling for Register

jamesmaa:
I looked into the koa docs and found some examples I was modeling off of. It seems to produce simpler and more readable code and it works.

https://github.com/koajs/koa#koa-v2
http://stackoverflow.com/questions/22059775/using-callbacks-with-nodejs-in-koa

The problem with the Login error handling fix is that ultimately it is convoluted to differentiate between different types of errors, and requires additional JSON parsing.

Happy to discuss whether this is the right approach or not

👓 @phil303

### DIFF
--- a/src/app/router/handlers/Login.js
+++ b/src/app/router/handlers/Login.js
@@ -23,8 +23,8 @@ export default class Login extends BaseHandler {
       dispatch(platformActions.navigateToUrl(METHODS.GET, '/'));
 
     } catch (e) {
-      if (e instanceof errors.ValidationError) {
-        dispatch(sessionActions.sessionError('ERROR NOT USED'));
+      if (e instanceof errors.ValidationError && e.errors && e.errors[0]) {
+        dispatch(sessionActions.sessionError(e.errors[0].error));
       } else {
         throw e;
       }
--- a/src/lib/errorLog.js
+++ b/src/lib/errorLog.js
@@ -1,5 +1,6 @@
 import { errors } from '@r/api-client';
 import makeRequest from './makeRequest';
+import config from 'config';
 
 const { ResponseError } = errors;
 
@@ -10,7 +11,7 @@ const ENV = (process.env.ENV || 'server').toUpperCase();
 const isAPIFailure = details => details.error instanceof ResponseError;
 
 
-export default function (details={}, errorEndpoints={}, options={ SHOULD_RETHROW: true}) {
+export default function errorLog(details={}, errorEndpoints={}, options={ SHOULD_RETHROW: true}) {
   // parse the stack for location details if we're passed
   // an Error or PromiseRejectionEvent
   const { error, rejection } = details;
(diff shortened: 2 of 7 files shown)


phil303:
@jamesmaa Regarding the error handling of the Login patch - while I agree that it's not ideal, I'm not sure there's another way. Even in this patch, you're doing type checking. It seems like it'd be more explicit to define errors into a "class" of error and check for that.

phil303:
Thanks for this patch! Looks good but let's chat about the error handling stuff :nail_care:

jamesmaa:
💇 Adding @schwers because I made a change to Login error handling

src/app/router/handlers/Login.js
```diff
@@ -24,7 +24,9 @@ export default class Login extends BaseHandler {
 
     } catch (e) {
       if (e instanceof errors.ValidationError) {
-        dispatch(sessionActions.sessionError('ERROR NOT USED'));
+        if (e.errors && e.errors[0]) {
```
phil303: for this line and the similar one in Register, can you move the checks into the conditional above.
```js
if (e instanceof errors.ValidationError && e.errors && e.errors[0]) {
  dispatch(sessionActions.sessionError(e.errors[0].error));
}
```

src/server/session/registerproxy.js
```diff
@@ -11,16 +14,20 @@ export default (router, apiOptions) => {
...
+      ctx.status = 401;
+      if (error && (typeof error === 'string')) {
+        ctx.type = 'application/json';
+        ctx.body = { error };
+      } else { // we're not sure what this error is, log it for now
+        errorLog({
+          error,
+          requestUrl: ctx.request.url,
+          userAgent: ctx.headers['user-agent'],
+        },{
```
phil303: nit, space between your comma and your last curly

phil303:
:nail_care: small comments and give a chance for @schwers to weigh in

src/server/session/registerproxy.js
```diff
@@ -11,16 +14,20 @@ export default (router, apiOptions) => {
...
-    } catch (e) {
-      console.log(e);
-      let errormsg = 'UNKNOWN_ERROR';
-
-      if (e && (typeof e === 'string')) {
-        errormsg = e;
+    } catch (error) {
+      ctx.status = 401;
+      if (error && (typeof error === 'string')) {
+        ctx.type = 'application/json';
```
schwers: 🔕  I think Koa will do this for you automatically if you set body to a dictionary like you're doing in the line below.

src/server/session/registerproxy.js
```diff
@@ -11,16 +14,20 @@ export default (router, apiOptions) => {
...
+        ctx.body = { error };
+      } else { // we're not sure what this error is, log it for now
+        errorLog({
```
schwers: 🔕  If we keep doing this we should make a helper for it, something like `logServerError(ctx)`

schwers:
💅 agree with Phil's comments, left a few comments on non-blocking stuff, overall patch looks good to me.

jamesmaa:
💇

src/lib/errorLog.js
```diff
@@ -173,6 +174,16 @@ export const formatLogJSON = logJSON => {
   return JSON.stringify(logJSON, null, 2).replace(/\\n/g, '\n');
 };
 
+export const logServerError = (error, ctx) => {
```
schwers: Thanks for adding the helper! Could you change https://github.com/reddit/reddit-mobile/blob/2X/src/server/meta/stats.js#L40 to use it too?

phil303:
:fish: thanks!

schwers:
The helper is nice, left a comment on one other place we should use it. I'll optimistically :fish: but make sure @phil303 is on-board too

jamesmaa:
7c1c92c5d0cea472168c2fad69efc157b17f05f4

PR #774, opened 2016-09-27
Sanitize json before putting it into the dom

phil303:
There's a possible XSS vector via query parameters in the url and script
tags. So we need to escape them. We'll do this via their unicode escape
code.

I've added the `safelyStringifyData` function, which has a counterpart
in modmail and am running it against the json data before it's put into
the dom.

:eyeglasses: @nramadas


nramadas:
:fish:

PR #775, opened 2016-09-27
Add retries to EventTracker's event posting

uzi:
We're seeing noise in the logs when EventTracker's
post fails.  Instead, retry two times with a
random delay and then give up silently.


phil303:
Can you give us more context in the commit please

uzi:
💇

src/lib/eventTracker.js
```diff
@@ -15,14 +17,20 @@ function calculateHash(key, string) {
   return hmac.read();
 }
 
-function post({url, data, query, headers, done}) {
+function post({url, data, query, headers, done}, retry = 0) {
```
phil303: nit, no spaces before/after equals

src/lib/eventTracker.js
```diff
@@ -15,14 +17,20 @@ function calculateHash(key, string) {
...
     .post(url)
     .query(query)
     .set(headers)
     .timeout(DEFAULT_API_TIMEOUT)
     .send(data)
-    .then(done);
+    .then(done)
+    .catch(() => {
+      if (retry < POST_RETRIES) {
+        const delay = Math.floor(1000 + Math.random() * 1000);
```
phil303: :no_bell: so always between 1 and 2 seconds? Any reasoning for that?

src/lib/eventTracker.js
```diff
@@ -15,14 +17,20 @@ function calculateHash(key, string) {
...
   return makeRequest
     .post(url)
     .query(query)
     .set(headers)
     .timeout(DEFAULT_API_TIMEOUT)
     .send(data)
-    .then(done);
+    .then(done)
+    .catch(() => {
+      if (retry < POST_RETRIES) {
```
phil303: should we log if this get exceeded? Might be able to tease out what's going on

phil303:
:nail_care:

src/lib/eventTracker.js
uzi: Just used the same logic as session refreshing... granted the comment there is wrong:

```
    // retry, only once, somewhere between 1 and 3 seconds later
    setTimeout(() => dispatch(refresh()), Math.floor(1000 + Math.random() * 1000));
```

Didn't want it too long or too short, basically.

src/lib/eventTracker.js
uzi: Sure.

src/lib/eventTracker.js
uzi: There's not much to log.  I checked e and the only thing in there is the timeout value.  That's before I did this retry code.

phil303: oh i dont know about that. we could log the data, headers, query params. There's a lot going on here other than just the error we get back. I'm wondering if there might be a malformed request causing the issue

uzi:
Abandoning for now

PR #778, opened 2016-09-28
Don't produce a canonical when we're missing data

uzi:
In the case that we're rendering for SEO and they
were hitting a banned subreddit (or the like), we
were barfing on trying to get the cleanPermalink
(in the comments case).  Instead, when looking an
comments and listings of potentially banned subreddits,
we recognize that the data is missing and just don't
make the canonical links.

👓  @nramadas @schwers @prashtx

### DIFF
--- a/src/lib/createCanonicalLinkFromState.js
+++ b/src/lib/createCanonicalLinkFromState.js
@@ -46,14 +46,20 @@ export const listingCanonical = (currentPage, state) => {
     return `/r/${newName}/`;
   }
 
-  // desperation case. we don't know how to handle this url, so just
-  // canonicalize to the frontpage.
-  return '/';
+  // we don't know how to handle this url, probably because of an issue
+  // getting data from the server (like a banned subbreddit), so give up.
+  return null;
 };
 
 export const commentsCanonical = (currentPage, state) => {
   const { postId } = currentPage.urlParams;
-  return state.posts[`t3_${postId}`].cleanPermalink;
+  const post = state.posts[`t3_${postId}`];
+  if (!post || !post.cleanPermalink) {
+    // We were unable to get the info from the server, which probably means
+    // the subreddit was banned or the like.
+    return null;
+  }
+  return post.cleanPermalink;
 };
 
 export const userCanonical = currentPage => {
@@ -87,5 +93,6 @@ const CANONICAL_ROUTES = [
 export default state => {
   const currentPage = state.platform.currentPage;
   const { handler } = matchRoute(currentPage.url, CANONICAL_ROUTES);
-  return `${config.reddit}${handler(currentPage, state)}`;
+  const route = handler(currentPage, state);
+  return route ? `${config.reddit}${route}` : null;
 };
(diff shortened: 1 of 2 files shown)


src/lib/createCanonicalLinkFromState.js
```diff
@@ -87,5 +93,6 @@ const CANONICAL_ROUTES = [
 export default state => {
   const currentPage = state.platform.currentPage;
   const { handler } = matchRoute(currentPage.url, CANONICAL_ROUTES);
-  return `${config.reddit}${handler(currentPage, state)}`;
+  const route = handler(currentPage, state);
+  return route ? `${config.reddit}${route}` : route;
```
schwers: Nice, this will essentially short-circuit the `canonicalLink` check right? 

I think it would be more readable if you made the `else` of the ternary be `null` here. i.e., 

```js
return route ? `${config.reddit}${route}` : null;
```

uzi: Done.

uzi: Oh, and yes -- that's the intent.

src/server/templates/main.jsx
```diff
@@ -18,9 +18,13 @@ const { assetPath } = config;
 export default function(data, store) {
   const state = store.getState();
 
-  const canonicalLink = !state.platform.shell
-    ? createCanonicalLinkFromState(state)
-    : '';
+  let canonicalLink = null;
+  if (!state.platform.shell) {
+    const route = createCanonicalLinkFromState(state);
```
schwers: I like that this is more explicit now, but I'm not sure what this buys now that `createCanonicalLinkFromState` is returning `null`.  I think we could make this just as concise as the old version without sacrificing clarity:

```js
const canonicalLink = !state.platform.shell
  ? createCanonicalLinkFromState(state)
  : null
```

src/lib/createCanonicalLinkFromState.js
```diff
@@ -46,14 +46,20 @@ export const listingCanonical = (currentPage, state) => {
...
-  return '/';
+  // we don't know how to handle this url, probably because of an issue
+  // getting data from the server (like a banned subbreddit), so give up.
+  return null;
 };
 
 export const commentsCanonical = (currentPage, state) => {
   const { postId } = currentPage.urlParams;
-  return state.posts[`t3_${postId}`].cleanPermalink;
+  const post = state.posts[`t3_${postId}`];
```
schwers: 🔕  Not important for this patch, just a general comment: at some point we should consider using the constants exported from `node-api-client` instead of hardcoding `t3_`

src/lib/createCanonicalLinkFromState.js
schwers: 🎉  okay, great, thanks!

schwers:
💅  Just a few readability things, happy to talk more on them if you disagree

schwers:
:fish:

PR #779, opened 2016-09-28
Use one snoo on comments pages

phil303:
Problem:
We're displaying two snoos when loading comment pages via the url, one
for the post itself and one for the comments section. This doesn't occur
when we come in via a listing. This seems to be because the post is
already loaded as a prereq for the listing page.

Fix:
Check for the post's existence first. If the post doesn't exist at all -
in the case we come in via the url - the post and the comments will come
in at the same time from our api. In that case show only the loading
icon.  If the post does exist - in the case we come in via the listing -
show the post component and show the loading snoo below, where the
comments would be.

### Before:
two loading snoos

### After:
one loading snoo

:eyeglasses: @schwers

### DIFF
--- a/src/app/pages/Comments.jsx
+++ b/src/app/pages/Comments.jsx
@@ -101,35 +101,39 @@ export const CommentsPage = connect(stateProps, dispatchProps, mergeProps)(props
     post,
   } = props;
 
+  if (!postLoaded) {
+    return (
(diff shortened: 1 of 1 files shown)


schwers:
:fish:

phil303:
:haircut: updated the commit message to be more correct. thanks for that catch ryan!

PR #780, opened 2016-09-28
Make sure client can update cookies

schwers:
All of our cookies were getting set with the `httpOnly` flag.
This was intended to increase security, but had the sideeffect
of preventing the client from updating cookies. A symptom of this
is your theme or compact preference not sticking. Session
cookies are still written with the `httpOnly` flag to protect
your tokens.

👓  @uzi @phil303 @nramadas

### DIFF
--- a/src/server/initialState/permanentCookieOptions.js
+++ b/src/server/initialState/permanentCookieOptions.js
@@ -16,9 +16,10 @@ export const permanentCookieOptions = () => {
     // In theory, now that it's deprecated, we should be passing `secure: true`
     // to the `new Cookies({})` constructor, but that's called by `Koa`
     // so we'll have to fork/pull-request that in the future.
-    httpOnly: true, // This flag isn't what you think it means. It tells
-    // browsers to not expose these headers to XMLHTTPRequestss, but still
-    // set the cookies
+    httpOnly: false, // This flag isn't what you think it means. Passing true
+    // would prevent cookies from being accessible from javascript. This sounds
+    // good at first but if you ever need to update cookies on the client,
+    // this needs to be set to false.
     expires,
   };
 };
--- a/src/server/session/logoutproxy.js
+++ b/src/server/session/logoutproxy.js
@@ -1,10 +1,19 @@
+import config from 'config';
 import clearSessionCookies from './clearSessionCookies';
 
 export default (router) => {
   router.post('/logout', async (ctx/*, next*/) => {
+    // NOTE: this is only temporary for now. Desktop uses a separate cookie
+    // called `reddit_session`. On any request to mweb, the server checks for it
+    // and attempts to convert it to an oauth session. To really log out
+    // we have to clear this cookie. The `token` cookie stores mweb oauth
+    // (and amp, modmail, etc) is now available on the root reddit domain, so
+    // we can unify these sessions and only have one cookie to clear.
+    ctx.cookies.set('reddit_session', { domain: config.rootReddit });
+    ctx.cookies.set('reddit_session');
+
     clearSessionCookies(ctx);
     ctx.cookies.set('over18');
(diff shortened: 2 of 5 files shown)


src/server/session/logoutproxy.js
```diff
@@ -1,10 +1,14 @@
+import config from 'config';
 import clearSessionCookies from './clearSessionCookies';
 
 export default (router) => {
   router.post('/logout', async (ctx/*, next*/) => {
+    // explicitly log out from desktop here (we need domain to really clear it)
+    ctx.cookies.set('reddit_session', { domain: config.rootReddit });
```
phil303: In theory this is only temporary right? Once we're on unified urls, this would eventually get to go away.

src/server/initialState/permanentCookieOptions.js
```diff
@@ -16,9 +16,10 @@ export const permanentCookieOptions = () => {
     // In theory, now that it's deprecated, we should be passing `secure: true`
     // to the `new Cookies({})` constructor, but that's called by `Koa`
     // so we'll have to fork/pull-request that in the future.
-    httpOnly: true, // This flag isn't what you think it means. It tells
-    // browsers to not expose these headers to XMLHTTPRequestss, but still
-    // set the cookies
+    httpOnly: false, // This flag isn't what you think it means. Passing true
```
phil303: what is the default here? Do we even need to set this?

src/server/session/setSessionCookies.js
```diff
@@ -4,7 +4,10 @@ import { SEPERATOR, VERSION } from './constants';
 export default (ctx, session) => {
   // Set the token cookie to be on the root reddit domain if we're not
   // running on localhost
-  const options = permanentRootCookieOptions(ctx);
+  const options = {
+    ...permanentRootCookieOptions(ctx),
+    httpOnly: true, // don't let client-side javascript access session cookies
```
phil303: so it looks like desktop sets this as httpOnly as well. Should we make a note here that this needs to stay in sync with desktop?

phil303:
:nail_care: just some questions

src/server/initialState/permanentCookieOptions.js
schwers: It's default `true` unfortunately: [cookies package link](https://github.com/pillarjs/cookies/blob/master/lib/cookies.js#L134)

src/server/session/logoutproxy.js
schwers: Right, this is only temporary. I think we could do it before unified urls if desktop would start using oauth like we do -- basically any world where both desktop and mweb use one and the same cookie for sessions.

src/server/session/setSessionCookies.js
schwers: I think its okay to omit that, but I'd be happy to add a note about it if you think its helpful. My reasoning is that this is less about parity with desktop and more so a best practice for keeping this cookie safe from malicious js. However if desktop javsacript _did_ start using this cookie we'd have to make a change, so maybe that enforces leaving a note?

src/server/initialState/permanentCookieOptions.js
phil303: I thought we used js-cookie?

src/server/session/logoutproxy.js
phil303: Do you think we should leave a comment saying something to that point? If only to remind future mweb developers that we can eventually get rid of it.

schwers: Sure, that sounds good to me

src/server/initialState/permanentCookieOptions.js
schwers: We do for the client, but koa uses `cookies` for the server

phil303:
:fish:

PR #781, opened 2016-09-29
Testout fix for 'can't set readonly properly' errors.

schwers:
The redux middleware error logger adds a property to errors,
to ensure they're aren't double logged. The property name was
`.caught` and we noticed some errors in the tap for trying
to set readonly properties, and traced it back to that line.
This patch attempts to fix this by using a more verbose, and
unlikely to be already present, property name.

👓  @uzi


uzi:
🐟

PR #783, opened 2016-09-29
Don't fail on tagging errors, should reveal hidden errors

schwers:
This tries to catch errors in setting a property on errors (which is for deduping error logs). If this works we might reveal a new error that's hidden behind the error for setting this property.

👓  @uzi @phil303

### DIFF
--- a/src/app/reduxMiddleware/errorLogger.js
+++ b/src/app/reduxMiddleware/errorLogger.js
@@ -69,9 +69,19 @@ export default function errorLogger() {
           // with other .then handlers in the promise chain.
           // regardless, we set a `._SEEN_BY_REDUX_ERROR_LOGGER` property to
           // true to prevent duplicate logging
+          let possibleDuplicate = undefined;
           if (!error._SEEN_BY_REDUX_ERROR_LOGGER) {
-            error._SEEN_BY_REDUX_ERROR_LOGGER = true;
-            logErrorWithConfig(error, store.getState(), actionStack);
+            try {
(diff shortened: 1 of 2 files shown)


uzi:
💅  Would it make sense to have a boolean we set in the catch that then affects one of the args going into `logErrorWithConfig()`?  That way we know if an error has possible dups... and maybe figure out which errors are causing this to happen in the first place.

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -66,9 +66,19 @@ export default function errorLogger() {
           // with other .then handlers in the promise chain.
           // regardless, we set a `._SEEN_BY_REDUX_ERROR_LOGGER` property to
           // true to prevent duplicate logging
+          let possibleDuplicate = undefined;
```
uzi: Make this `false` instead of `undefined`?

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -106,14 +116,15 @@ const checkForSpecificErrors = action => {
   }
 };
 
-const logErrorWithConfig = (error, state, actionStack) => {
+const logErrorWithConfig = (error, state, actionStack, possibleDuplicate) => {
```
uzi: Make it `possibleDuplicate=false` in the args so it's optional?

uzi:
💅

uzi:
:fish: never mind -- had a quick discussion -- good as is

PR #784, opened 2016-09-30
Fix deletion of reddit_session cookie

schwers:
👓 @uzi @phil303 
 
the second argument is the cookie value, the third is options. :(


uzi:
:fish:

PR #786, opened 2016-09-30
Convert imgur and giphy links to use https.

uzi:
Should take care of a lot of the "Mixed content" warnings we get and be safer!


schwers:
:fish:

PR #788, opened 2016-09-30
Add autocomplete attribute to LoginInput

phil303:
Bug:
User was able to get passwords to autocomplete in plaintext by
originally typing them in with the "eyeball" enabled (makes password
input plaintext), logging out, going back to the login and enabling the
eyeball. Chrome would try to autocomplete his passwords.

Fix:
Make it possible to add the autocomplete attr onto the input dom nodes.
This works on [all mobile clients](http://caniuse.com/#search=autocomplete).

:eyeglasses: @uzi

### DIFF
--- a/src/app/components/Login/index.jsx
+++ b/src/app/components/Login/index.jsx
@@ -117,6 +117,7 @@ class Login extends React.Component {
             type={ passwordFieldType }
             placeholder='Password'
             showTopBorder={ false }
+            shouldAutocomplete={ false }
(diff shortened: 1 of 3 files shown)


uzi:
💅  Can we also put this on the password field for registration?  Thanks!

phil303:
yes! good call

phil303:
:haircut:

uzi:
:fish:

PR #789, opened 2016-09-30
Make overlays use redux state instead of query params

schwers:
I have two separate commits here (they seemed related so I grouped them, happy to split into separate PRs).

The first commit makes overlays use redux state for overlays, instead of our current usages of query params:

Proposed commit message:

Overlays were controlled by a queryParam. This caused odd, and hard-to-reproduce, issues with our back button state. It also leads to weird situations where you end up with an overlay open when you hit back, when you really wanted to go back to the page without it open. It might also be the source of a bug where we're trying to route to non-relative urls (i.e. something is causing the platform stack to try and navigate to `https://www.google.com.au`)

Second commit makes the subreddit subscription list be auto-expanded, which is pretty awesome :)

👓  @phil303 @uzi


src/app/actions/overlay.js
```diff
@@ -0,0 +1,22 @@
...
+export const COMMUNITY_MENU = 'COMMUNITY_MENU';
+export const toggleCommunityMenu = () => toggle(COMMUNITY_MENU);
+
+export const SEARCH_BAR = 'SEARCH_BAR';
+export const toggleSearchBar = () => toggle(SEARCH_BAR);
+
+export const SETTINGS_MENU = 'SETTINGS_MENU';
+export const toggleSettingsMenu = () => toggle(SETTINGS_MENU);
+
+export const POST_SUBMIT = 'POST_SUBMTI';
```
uzi: POST_SUBMTI

phil303:
Hey ryan, just starting to look at the PR - First of all, things feel so much snappier, it's like night and day. Thanks for doing this. It's a super nice upgrade.

Second, as part of this PR, can you throw together a list of what's left? I'd like to track it so we make sure we get it all eventually

src/app/components/TopNav/index.jsx
```diff
@@ -57,53 +56,80 @@ export const TopNav = (props) => {
...
+    // it's safe to do this all the time because it's cached
+    dispatch(subscribedSubredditsActions.fetchSubscribedSubreddits());
+    dispatch(overlayActions.toggleCommunityMenu());
+  },
+
+  togglePostSubmit(isLoggedIn) {
+    if (isLoggedIn) {
+      dispatch(overlayActions.togglePostSubmit());
+    } else {
+      dispatch(platformActions.navigateToUrl(METHODS.GET, '/register'));
```
phil303: should this go to `/login`? That's where the submit handler goes to right now when not logged in.

src/app/components/TopNav/index.jsx
```diff
@@ -57,53 +56,80 @@ export const TopNav = (props) => {
...
+});
+
+const mergeProps = (stateProps, dispatchProps) => {
+  const { isLoggedIn } = stateProps;
+  const { togglePostSubmit } = dispatchProps;
+
+  return {
+    ...stateProps,
+    ...dispatchProps,
+    togglePostSubmit() { togglePostSubmit(isLoggedIn); },
```
phil303: :no_bell: nit/question - I notice you're using the newer method creation syntax in this file, which feels inconsistent with the rest of the app. Do we care about consistent syntax when constructing objects?

src/app/reducers/overlay.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+createTest({ reducers: { overlay }, routes }, ({ getStore, expect }) => {
+  describe('overlay', () => {
+    describe('SET_PAGE', () => {
+      it('should reset to default state when switching pages', () => {
+        const { store } = getStore({ overlay: 'something' });
+
+        store.dispatch(platformActions.navigateToUrl(METHODS.GET, '/'));
+        const { overlay } = store.getState();
+        expect(overlay) .to.eql(DEFAULT_STATE);
```
phil303: space after your `expect` statement

src/app/reducers/overlay.test.js
```diff
@@ -0,0 +1,62 @@
...
+      it('should close the overlay if a second TOGGLE_OVERLAY action is fired', () => {
+        const OVERLAY_NAME = 'foo';
+        const { store } = getStore({ overlay: OVERLAY_NAME });
+
+        store.dispatch(overlayActions.toggle(OVERLAY_NAME));
+        const { overlay } = store.getState();
+        expect(overlay).to.equal(DEFAULT_STATE);
+      });
+
+      it('should switch overlay names if a different type is TOGGLE\'d', () => {
```
phil303: :no_bell: lol... `TOGGLE\'d'`

src/app/router/handlers/OverlayMenuThemeToggle.js
```diff
@@ -1,11 +1,11 @@
...
 
 import { toggleTheme } from 'app/actions/theme';
-import { closeOverlayMenu } from 'app/actions/overlayMenu';
+import { closeOverlay } from 'app/actions/overlay';
 
 export default class OverlayMenuThemeToggle extends BaseHandler {
   async [METHODS.POST](dispatch) {
     dispatch(toggleTheme());
-    dispatch(closeOverlayMenu());
+    dispatch(closeOverlay());
```
phil303: I have a proposal for you. What if, instead of calling this explicitly in a second dispatch, which causes an (admittedly negligible) re-render. What if your `overlay` reducer looked like this:

```js
export default function(state=DEFAULT_STATE, action={}) {
  switch (action.type) {
    case platformActions.SET_PAGE:
    case overlayActions.TOGGLE_THEME:
    case overlayActions.TOGGLE_COMPACT:
    case overlayActions.CLOSE_OVERLAY: {
      return DEFAULT_STATE;
    }

    // ....

}
```

src/app/actions/overlay.js
schwers: whoops :)

src/app/components/TopNav/index.jsx
schwers: I slightly prefer how this syntax looks, do you have a preference? I'd be happy to switch to arrow syntax for consistency, but if we like this new syntax I wouldn't mind taking a pass at converting in a separate patch.

phil303: For plain objects, I kinda like the old style syntax as a way of differentiating methods defined in a class and methods defined directly on an object. It's an arbitrary line at best though so no strong feelings, especially if we were consistent about it. There's already a lot going on in this brave new ES6 world of ours.

src/app/reducers/overlay.test.js
schwers: Double-quotes :soon: ™️

src/app/components/TopNav/index.jsx
schwers: We talked in person about this, I'm changing the submit handler to go to `/register`

src/app/components/TopNav/index.jsx
schwers: I'll 'revert' for now. I agree that the distinction is nice, we can keep thinking on this.

src/app/router/handlers/OverlayMenuThemeToggle.js
schwers: From talking in person: this is a cleaner approach that's more idiomatic and is more decoupled. One takeaway: this highlights the importance of teaching redux as part of our onboarding along with having good reducer documentation.

schwers:
:haircut:

schwers:
:haircut_man:

uzi:
💅  FOR SCIENCE

schwers:
:haircut_man:

src/app/components/TopNav/index.jsx
```diff
@@ -57,53 +56,80 @@ export const TopNav = (props) => {
...
+});
+
+const mergeProps = (stateProps, dispatchProps) => {
+  const { isLoggedIn } = stateProps;
+  const { togglePostSubmit } = dispatchProps;
+
+  return {
+    ...stateProps,
+    ...dispatchProps,
+    togglePostSubmit: () => { togglePostSubmit(isLoggedIn); },
```
phil303: :no_bell: since this is a one liner, you can remove the curlies

schwers:
💇

phil303:
had one new no_bell but otherwise :fish:

src/app/components/TopNav/index.jsx
schwers: 🔕  True, I was a little concerned it could look like the return value was needed (no brackets creates implicit return right?)

schwers:
Just as a note for @phil303, I think the only thing with urls that _shouldn't_ have them is comment reply forms

PR #790, opened 2016-10-02
Normalize z-indexes across app

phil303:
This is a prep patch/hopeful cleanup for the way we handle z-indexes.
The goal is to have a finite and easy to remember amount of stacking
levels components can live at. This is to lower the chance of
unforeseen visual bugs future devs might run into. The z-indexes are
meant to be in a stacking context that is always relative to the dom
root and therefore easier to reason about.

Based on a 50 point scale - with the majority of components having 
the implicit z-index of 0 - the z-indexes of anything above the default 
are as follows (the higher in the list, the higher the stacking precedence):

- Toasters: 50
- Modals: 40    (Reporting, CaptchaBox)
- Nav Bar and menus: 30
- Dropdowns/tooltips: 20
- Opaque overlay: 10

This is meant to handle the most general cases, which is what our
current situation is now anyway, and not meant as a strict ruleset. 

I'm also totally open to thoughts on the stacking precedences.

:eyeglasses: @nramadas

### DIFF
--- a/src/app/components/Dropdown/styles.less
+++ b/src/app/components/Dropdown/styles.less
@@ -1,6 +1,11 @@
 @import (reference) '~app/less/variables';
 @import (reference) '~app/less/themes/themeify';
 
+.DropdownWrapper {
+  position: relative;
+  z-index: @z-index-dropdown;
+}
+
 .Dropdown {
   width: 600px;
 
--- a/src/app/components/CaptchaBox/styles.less
+++ b/src/app/components/CaptchaBox/styles.less
@@ -2,7 +2,7 @@
 @import (reference) '~app/less/themes/themeify';
 
 .CaptchaBox {
-  z-index: 3;
+  z-index: @z-index-modal;
   position: fixed;
   top: 50px;
   left: 20px;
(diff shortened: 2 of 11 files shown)


src/app/components/Dropdown/styles.less
```diff
@@ -1,6 +1,11 @@
 @import (reference) '~app/less/variables';
 @import (reference) '~app/less/themes/themeify';
 
+.Dropdown__wrapper {
```
nramadas: it's a nit, but this makes it look like `__wrapper` is a subcomponent of `Dropdown`, when in fact it is its parent. can we call this `DropdownWrapper` instead?

phil303: for shizzle

nramadas:
Can we make the multiply the 5 point scale by 100? I can forsee there might be a problem in the future where we have two nav bar type items open at the same time, but we need finer control over what overlays what. Allowing for some wiggle room in the 5 point scale helps out.

phil303:
Well the idea is to not give too much wiggle room. A 500 point scale makes it way less obvious to someone developing a new modal/dropdown/whatever where that thing should live at. Wanna chat in person though? I'm having trouble picturing your use case.

phil303:
:haircut: @nramadas - Made this a 50 point scale as discussed

nramadas:
:fish:

phil303:
fyi @nramadas I've updated the variable names to be namespaced per @ajacksified's suggestion.

PR #791, opened 2016-10-03
Add autocomplete attribute to LoginInput

phil303:
Bug:
User was able to get passwords to autocomplete in plaintext by
originally typing them in with the "eyeball" enabled (makes password
input plaintext), logging out, going back to the login and enabling the
eyeball. Chrome would try to autocomplete his passwords.

Fix:
Make it possible to add the autocomplete attr onto the input dom nodes.
This works on [all mobile
clients](http://caniuse.com/#search=autocomplete).

:eyeglasses: @uzi

### DIFF
--- a/src/routes.jsx
+++ b/src/routes.jsx
@@ -1015,4 +1015,3 @@ function routes(app) {
 }
 
 export default routes;
-export { buildProps };
--- a/src/views/components/formElements/minimalinput.jsx
+++ b/src/views/components/formElements/minimalinput.jsx
@@ -4,7 +4,7 @@ const T = React.PropTypes;
 
 function Input(props) {
   const { showTopBorder, name, type, placeholder,
-          onChange, error, children, value } = props;
+          onChange, error, children, value, autocomplete } = props;
   const showTop = showTopBorder ? 'show-top' : '';
 
   let errorClass = '';
@@ -27,6 +27,7 @@ function Input(props) {
           value={ value }
           className={ `minimalInput__input ${showTop} ${errorClass}` }
           placeholder={ placeholder }
+          autoComplete={ autocomplete }
           onChange={ onChange }
         />
         { children }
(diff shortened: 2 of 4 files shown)


src/routes.jsx
```diff
@@ -1015,4 +1015,3 @@ function routes(app) {
 }
 
 export default routes;
-export { buildProps };
```
phil303: fix a dual export problem

uzi: Weird -- why is it there then, and is it needed?  This is the only part that scares me.

src/views/components/formElements/minimalinput.jsx
```diff
@@ -4,7 +4,7 @@ const T = React.PropTypes;
 
 function Input(props) {
   const { showTopBorder, name, type, placeholder,
-          onChange, error, children, value } = props;
+          onChange, error, children, value, autocomplete=null } = props;
```
uzi: 🔕  If this wasn't 1X, I'd be asking about using defaultProps to set default values.  That seems to be the convention for react components.

uzi:
💅  I just want to be sure about that routes.jsx change that it's not going to mess up something else... and the defaultProps thing is optional.

src/routes.jsx
phil303: Your fear is warranted. I have no idea but it feels like an accident. There is a direct export on the function as it was defined earlier in the file.

 I acked around a bit and found that everything was importing `buildProps` in as 
```js
import { buildProps } from 'wherever';
```

The way it is now doesn't even work that way.

phil303: @uzi, @nramadas https://github.com/reddit/reddit-mobile/blob/master/src/routes.jsx#L102

phil303:
@uzi yea the default props thing is not that big a deal since this is 1X - :haircut:

phil303:
https://github.com/reddit/reddit-mobile/commit/da56c0bf68efb51ef85453f2a4adaf55780d4fcb

PR #795, opened 2016-10-04
Fix coloring of upvote/downvote icons on :hover

ajacksified:
Themed .icon:hover classes were overriding comment upvote and downvote
colors. This makes it so that upvoted and downvoted colors are applied
appropriately by following the same method the link uses - namely,
!important.

👓 @schwers


schwers:
:fish:

PR #796, opened 2016-10-04
Keep comment reply state in state instead of qs

ajacksified:
This moves whether or not a comment reply form is collapsed to a new key in the state object - replyingComments[id] - instead of in the querystring. Querystrings still work for no-js clients, but an onClick event captures and stops propagation for js-enabled clients.

👓 @schwers


schwers:
Leaving a note from discussing in person; we don't actively support NoJS for posting so we should remove the query parameter entirely.

src/app/reducers/replyingComments.js
```diff
@@ -0,0 +1,29 @@
+import merge from '@r/platform/merge';
+
+import * as commentActions from 'app/actions/comment';
+import * as loginActions from 'app/actions/login';
+
+const DEFAULT = {};
```
schwers: One pattern we found helpful is to export the default state, then your tests don't have to be aware of what exactly the default is. Could you give that a try here?

src/app/pages/Comments.jsx
```diff
@@ -60,15 +62,19 @@ const stateProps = createSelector(
   },
 );
 
-const dispatchProps = dispatch => ({
+const dispatchProps = (dispatch) => ({
```
schwers: Just a small style nit, we generally omit parens on single arg functions

src/app/components/Comment/CommentReplyForm/styles.less
```diff
@@ -27,7 +27,7 @@
     padding: 0.5em 0 0;
     float: left;
 
-    &[class*=" icon-"] {
+    &[class*=" icon-"], &[class*=" icon-"]:visited {
```
schwers: 🔕  This might be a holdover from 1X, do we still need a class selector here? All of our icons should have the `.icon` class in addition to the `icon-${iconName}` class

schwers:
💅

ajacksified:
:haircut_man:

src/app/components/Comment/CommentReplyForm/index.jsx
```diff
@@ -4,22 +4,31 @@ import React from 'react';
...
 
-export const closeButton = (currentPage) => {
-  const href = urlFromPage(currentPage, { queryParams: { commentReply: undefined } });
+const T = React.PropTypes;
 
+export const closeButton = onToggleReply => {
   return (
-    <BackAnchor
-      href= { href }
+    <a
```
schwers: Could this be a `<span />` or `<div />` instead ?

src/app/components/Comment/CommentTools/index.jsx
```diff
@@ -63,17 +62,17 @@ CommentTools.defaultProps = {
...
 };
 
-const renderReply = ({ currentPage, id, commentReplying }) => {
-  const href = urlFromPage(currentPage, { queryParams: { commentReply: id } });
-
+const renderReply = ({ commentReplying, onToggleReply }) => {
   return (
-    <Anchor
-      href= { href }
+    <a
```
schwers: Same question on tag usage as above, I think having `<a />` makes this confusing. `onToggleReply` also now has increased complexity as it requires a call to `e.preventDefault()`, otherwise the page is going to scroll to the top when the browser looks for the `#` anchor

src/app/components/CommentsPage/CommentsPageTools/index.jsx
```diff
@@ -69,9 +77,9 @@ export default ({ replying, post, hasSingleComment, currentPage, id, onSortChang
           onSortChange={ onSortChange }
         />
         { commentingDisabled ? null :
-          <Anchor className='Button m-linkbutton' href={ replyHref }>
+          <a className='Button m-linkbutton' href={ replyHref } onClick={ onToggleReply }>
```
schwers: Same `a` question

schwers:
💅  Bunch of small nits on the usage of `<a />` tags, I think this needs to not use `<a />` at all to really remove the query param and its overhead. Could you re-look at the testing comment and let me know your thoughts?

ajacksified:
:haircut_man:

src/app/pages/Comments.jsx
```diff
@@ -64,11 +66,15 @@ const dispatchProps = dispatch => ({
   navigateToUrl(url, query) {
     dispatch(navigationActions.navigateToUrl(METHODS.GET, url, query));
   },
+  onToggleReply (e, id) {
+    e.preventDefault();
```
schwers: I think `onToggleReply` is unneeded now right?

src/app/reducers/replying.js
```diff
@@ -1,7 +1,7 @@
 import merge from '@r/platform/merge';
 import * as replyActions from 'app/actions/reply';
 
-const DEFAULT = {};
+export const DEFAULT = {};
```
schwers: I don't see where this new export is being used

ajacksified: Oh, this was cleanup from an existing file, not my new reducer - could be a separate commit instead if you'd like, or not part of this PR

src/app/components/Comment/index.jsx
```diff
@@ -229,7 +230,7 @@ const selector = createSelector(
   (state, props) => state.comments[props.commentId],
   (state, props) => state.moreCommentsRequests[props.commentId] || DEFAULT_COMMENT_REQUEST,
   (state, props) => state.collapsedComments[props.commentId] || false,
-  (state, props) => state.platform.currentPage.queryParams.commentReply === props.commentId,
+  (state, props) => state.replyingComments[props.commentId] || false,
```
ajacksified: I was following the example from a line above - should I change everything instead?

ajacksified:
:haircut_man:

src/app/reducers/replying.js
schwers: Okay, I think its fine to keep it in this patch.

schwers:
Just a few changes (namely `!!` in selector, and exporting the default from your reducer for tests). :fish:'ing optimistically since those should be easy

ajacksified:
A significant rebase, can you take a quick look, @schwers? :haircut_man:

src/app/actions/comment.js
```diff
@@ -60,6 +60,16 @@ export const toggleCollapse = (id, collapse) => async (dispatch) => {
   dispatch(toggledCollapse(id, collapse));
 };
 
+export const TOGGLE_REPLY = 'COMMENT__TOGGLE_REPLY';
+
+export const toggledReply = (id, reply) => ({ type: TOGGLE_REPLY, id, reply });
+export const toggleReply = (id, reply) => async (dispatch) => {
```
nramadas: out of curiosity, why is this needed? it seems to just dispatch `toggledReply`. Couldn't whatever is dispatching this just dispatch `toggledReply` directly?

src/app/components/Comment/index.jsx
```diff
@@ -269,8 +270,8 @@ const selector = createSelector(
   state => state.platform.currentPage,
   (state, props) => state.comments[props.commentId],
   (state, props) => state.moreCommentsRequests[props.commentId] || DEFAULT_COMMENT_REQUEST,
-  (state, props) => state.collapsedComments[props.commentId] || false,
-  (state, props) => state.platform.currentPage.queryParams.commentReply === props.commentId,
+  (state, props) => !!state.collapsedComments[props.commentId] || false,
```
nramadas: nit, but i think you don't have to `|| false` anymore since you're `!!`ing

src/app/actions/comment.js
ajacksified: I was following the pattern set by collapse and some other toggle actions. It seemed extraneous, but I was going for consistency - happy to nuke it.

ajacksified:
:haircut:

src/app/less/components/button.less
```diff
@@ -1,8 +1,10 @@
-.Button, .Button:active, .Button:visited, .Button:hover, .Button:focus {
+.Button {
   padding: .5em 1em;
   letter-spacing: 1.6;
   line-height: 1.5;
+}
 
+.Button, .Button:active, .Button:visited, .Button:hover, .Button:focus {
```
schwers: 🔕  Small nit: this could be less typing if you nested it in the `.Button` selector above and use less's & shorthand for writing a new, non-nested rule. i.e. `&, &:active, &:visited, &:hover, &:focus`

schwers:
:fish:

ajacksified:
(merged @ 4a5bd54fa4ccbdaa2fbf90e88adc57dcac0c42f3)

PR #797, opened 2016-10-04
Redirect live to desktop

ajacksified:
This sets up a url handler to redirect /live/\* urls to https://www.reddit.com/live/* urls.

Depends on https://github.com/reddit/node-platform/pull/27 to work client-side.

👓 @schwers


src/app/router/handlers/Live.js
```diff
@@ -0,0 +1,14 @@
+import { setPage } from '@r/platform/actions';
+import { BaseHandler, METHODS } from '@r/platform/router';
+
+export default class Live extends BaseHandler {
+  async [METHODS.GET](dispatch, getState) {
+    const { platform: { currentPage }} = getState();
+    const { urlParams, queryParams, hashParams, referrer } = currentPage;
+    const { url } = currentPage;
+
+    dispatch(setPage(`https://www.reddit.com${url}`, {
```
schwers: Can we use `config.reddit` here instead?

schwers:
Just a small nit, looks good otherwise :) 💅

ajacksified:
:haircut_man: and also 👓 @nramadas

src/Client.js
```diff
@@ -18,6 +18,9 @@ import reduxMiddleware from 'app/reduxMiddleware';
 import { sendTimings, onHandlerCompleteTimings } from 'lib/timing';
 import Session from 'app/models/Session';
 
+import redirectOnFullUrl from 'app/client/reduxMiddleware/redirectOnFullURL';
+reduxMiddleware.push(redirectOnFullUrl);
```
nramadas: it might be cleaner to just include this middleware in `app/reduxMiddleware/index.js`.

src/app/router/handlers/Live.js
```diff
@@ -0,0 +1,15 @@
...
+import { BaseHandler, METHODS } from '@r/platform/router';
+import config from '../../../config';
+
+export default class Live extends BaseHandler {
+  async [METHODS.GET](dispatch, getState) {
+    const { platform: { currentPage }} = getState();
+    const { urlParams, queryParams, hashParams, referrer } = currentPage;
+    const { url } = currentPage;
+
+    dispatch(setPage(`${config.reddit}${url}`, {
```
nramadas: :no_bell: so, i spent a while thinking about this last night. I'm getting a bit of a code smell with using `setPage` in this manner. We are masking the side-affect nature of what it is we're doing instead of being explicit about what it is we want to happen.

I think what worries me the most is that without the added middleware, this does not do what we think it would on the client side. But, looking at this particular bit of code, that's not evident.

Would you be open to opening up the discussion and including schwers, phil, and uzi? I actually think this particular case is a new problem we haven't encountered before, and it would be good for all of us to talk through it.

nramadas:
:nail_care:

src/Client.js
nramadas: nvm, i had the wrong idea about `reduxMiddelware`

ajacksified:
After discussion: create a `redirect` action instead of using `setPage`. also update user redirection

schwers:
Closing in favor of implementing a `redirect` action in `@r/platform`

PR #798, opened 2016-10-05
Add support for displaying message threads

uzi:
Adds another page in which we display all of the
messages in a conversation. This is still read-only
support, but now the user should be able to see
everything at least.

Also guard the message endpoints against logged
out users while in the neighborhood

👓  @schwers @phil303

### DIFF
--- a/src/app/components/MessageThread/Message.jsx
+++ b/src/app/components/MessageThread/Message.jsx
@@ -1,0 +1,39 @@
+import './Message.less';
+import React from 'react';
+import { Anchor } from '@r/platform/components';
+
+import { short } from 'lib/formatDifference';
+import RedditLinkHijacker from 'app/components/RedditLinkHijacker';
+
+const T = React.PropTypes;
+
+const SEPARATOR = ' \u2022 ';
+
+export default function MessageThreadMessage(props) {
+  const { message } = props;
+
+  return (
+    <div className='MessageThreadMessage'>
+      <div className='MessageThreadMessage__title'>
+        <Anchor
+          className='MessageThreadMessage__authorLink'
+          href={ `/user/${ message.author }` }
+        >
+          { message.author }
+        </Anchor>
+        { SEPARATOR }
+        { short(message.createdUTC) }
+      </div>
+      <RedditLinkHijacker>
+        <div
+          className='MessageThreadMessage__body'
+          dangerouslySetInnerHTML={ { __html: message.bodyHTML } }
+        />
+      </RedditLinkHijacker>
+    </div>
+  );
+}
+
+MessageThreadMessage.propTypes = {
+  message: T.object.isRequired,
+};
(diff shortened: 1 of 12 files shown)


uzi:
👓  @nramadas

src/app/components/MessageThread/Message.jsx
```diff
@@ -0,0 +1,39 @@
...
+        >
+          { message.author }
+        </Anchor>
+        { SEPARATOR }
+        { short(message.createdUTC) }
+      </div>
+      <RedditLinkHijacker>
+        <div
+          className='MessageThreadMessage__body'
+          dangerouslySetInnerHTML={ { __html: message.bodyHTML } }
```
phil303: this looks like an xss vector. just to double-check, have you made sure this is all properly escaped from the backend?

src/app/components/Messages/Message.jsx
```diff
@@ -7,7 +7,7 @@ import RedditLinkHijacker from 'app/components/RedditLinkHijacker';
 
 const T = React.PropTypes;
 
-const SEPARATOR = '\u2022';
+const SEPARATOR = ' \u2022 ';
```
phil303: :no_bell: just curious, what's with the white space changes?

src/app/router/handlers/Messages.js
```diff
@@ -1,12 +1,21 @@
...
 
 import * as mailActions from 'app/actions/mail';
 import { fetchUserBasedData } from './handlerCommon';
 import { getBasePayload, logClientScreenView } from 'lib/eventUtils';
 
 export default class Messages extends BaseHandler {
   async [METHODS.GET](dispatch, getState) {
-    dispatch(mailActions.fetchInbox(this.urlParams.mailType));
+    if (!getState().session.isValid) {
+      return dispatch(platformActions.setPage('/login'));
```
phil303: from talking to @schwers about the overlay patch, I believe this should be /register

phil303:
This looks solid. Nice work. Had some small comments/questions. :nail_care:

src/app/router/handlers/Messages.js
schwers: The argument makes sense both ways I suppose. How about we make a constant called `LOGGED_OUT_FALLBACK` or something like that. We can talk to some folks from product about what it should really be

uzi:
💇

src/app/components/MessageThread/styles.less
```diff
@@ -0,0 +1,59 @@
...
+
+    .themeify({
+      background-color: @theme-body-color;
+    });
+  }
+
+  &__item {
+    padding: 2 * @grid-size 0;
+
+    & ~ & {
```
schwers: 💯

schwers:
:fish:

PR #799, opened 2016-10-05
Prevent nsfw show button from opening link

ajacksified:
Without `e.preventDefault()`, nsfw links will open up in a new tab as
well as revealing the blurred image below.

👓 @schwers


schwers:
:fish:

PR #800, opened 2016-10-05
Titles

ajacksified:


src/app/router/handlers/CommentsPage.js
```diff
@@ -46,6 +48,17 @@ export default class CommentsPage extends BaseHandler {
...
 
+  buildTitle (state, pageId) {
+    const page = state.commentsPages[pageId];
+    const post = state.posts[page.postId];
+
+    if (post) {
+      return `Reddit - ${post.subreddit} - ${post.title}`;
+    }
+
+    return 'Reddit: the Front Page of the Internet';
```
schwers: Might be good to have a default title in `constants.js` now that is in a few places

src/app/reducers/pageMetadata.js
```diff
@@ -0,0 +1,26 @@
+import { SET_TITLE } from 'app/actions/pageMetadata';
+import * as platformActions from '@r/platform/actions';
+
+const DEFAULT = {
+  title: 'Reddit: the Front Page of the Internet',
+  canonicalPath: undefined,
```
schwers: I don't think we need to track `canonicalPath` here with the other other cannonicals code in place

src/app/router/handlers/CommentsPage.js
ajacksified: oh, yeah, that should just return undefined now, which will fall back to the default in the reducer

src/app/router/handlers/PostsFromSubreddit.js
```diff
@@ -31,6 +32,15 @@ export default class PostsFromSubreddit extends BaseHandler {
     });
   }
 
+  buildTitle (state, subredditName) {
+    if (!subredditName) { return; }
+    const subreddit = state.subreddits[subredditName];
+
+    if (subreddit && subreddit.title) {
```
prashtx: Does this guard against treating non-subreddit listings as actual subreddits? Or can we have subreddits without a title? If the latter, perhaps we should fall back to `subredditName` when `subreddit.title` is falsey.

ajacksified: The former.

ajacksified:
👓 @schwers 

Let's start with the two most important pages, and expand titles to other pages in separate PRs. This should clear the deploy blocker, at least.

src/app/reducers/pageMetadata.js
```diff
@@ -0,0 +1,25 @@
...
+    // Reset to the default on route, so we don't have sticky titles
+    case platformActions.GOTO_PAGE_INDEX:
+    case platformActions.NAVIGATE_TO_URL: {
+      return DEFAULT;
+    }
+
+    case SET_TITLE: {
+      return {
+        ...state,
+        title: action.title,
```
schwers: It looks like there's a subtle bug with these titles, handlers always call `dispatch(setTittle())` even if the post or subreddit didn't load, leading to `action.title === undefined`. Should this be returning `DEFAULT` if `action.title` is undefined?

schwers:
Should https://github.com/reddit/reddit-mobile/blob/2X/src/server/templates/main.jsx#L28 be using the title from redux state?

src/app/reducers/pageMetadata.js
```diff
@@ -0,0 +1,25 @@
+import { SET_TITLE } from 'app/actions/pageMetadata';
```
schwers: Reminder: please add tests for this reducer

schwers:
💅

ajacksified:
:haircut_man:

schwers:
:fish:

PR #801, opened 2016-10-05
Reporting Modal

phil303:
Feature:
A modal that pops up when a user tries to report a comment or a post (messages will come in a followup patch). Upon successful reporting, the modal will close and a toaster pops up to notify the user all went well. Clicking anywhere outside the modal box will also close modal.

Details:
This feature works by having a top level modal component. Inside this component, there is a switch statement checking for modal types in the `modal` redux state key. The modal key also carries with it the props intended for the modal. In this case, just the `thingId` is necessary to report someone so that's the only prop passed into `ReportingModal`.

To trigger the modal, the modal reducer is listening for the REPORT action, which will set the report modal type.

Other notes:
- I've moved `DropdownCover` into `AppMain.jsx`. Since this is a UI component, this feels like a more appropriate place for it.
- This needs to be deployed in concert with `reddit-mobile-deps` since I've updated `package.json`.

Here's what the modal looks like:
reporting modal

:eyeglasses: @nramadas @schwers


src/app/pages/PostsFromSubreddit.jsx
```diff
@@ -80,18 +82,21 @@ export const PostsFromSubredditPage = connect(mapStateToProps)(props => {
...
-      { !forFakeSubreddit && <CommunityHeader subredditName={ subredditName } /> }
-      { renderSubnav &&
-        <SubNav
-          rightLink={ subnavLink }
-          showWithoutUser={ true }
-        >
-          <SortAndTimeSelector />
-        </SubNav> }
+      { !forFakeSubreddit ? <CommunityHeader subredditName={ subredditName } /> : null }
+      { showSubnav ? renderSubNav(subnavLink) : null }
```
phil303: just some light cleanup here

src/app/components/Modal/index.jsx
```diff
@@ -5,6 +5,10 @@ import './styles.less';
 
 const T = React.PropTypes;
 
+// TODO: this was incorrectly named, and should be renamed to reflect the fact
```
schwers: 🔕 Maybe it's from working with iOS, but this still feels like a Modal that's presented fullscreen

src/app/components/ModalSwitch/index.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+
+import ReportingModal from '../ReportingModal';
+import * as reportingActions from 'app/actions/reporting';
+
+function ModalSwitch(props) {
```
schwers: 🔕  Small nit, but I have a minor preference for arrow syntax here.

src/app/components/Post/index.jsx
```diff
@@ -152,41 +155,37 @@ export function Post(props) {
...
-const singleSelector = (_, props) => props.single;
-const postModelSelector = (state, props) => state.posts[props.postId];
-const expandedSelector = (state, props) => !!state.expandedPosts[props.postId];
-const unblurredSelector = (state, props) => !!state.unblurredPosts[props.postId];
-const userSelector = state => state.user;
-const combineSelectors = (postId, compact, expanded, unblurred, single, post, user) => ({
-  postId, compact, expanded, unblurred, single, post, user,
-});
-
-const makeConnectedPostSelector = () => {
```
schwers: side-note: We're going to have some merge conflicts here from my patch

I think we should keep this as function that returns the result of `createSelector`. This change has implications for how the selectors are memoized. If you you pass a function that returns the result of `createSelector` your connected component is memoized per-component instance, as opposed to globally for any instance of the component. This is important for components that have multiple instances rendered in a list. I haven't tested this patch yet but I remember adding this gave us a nice performance boost

src/app/components/ReportingModal/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+const REASONS = {
+  HARASSMENT: 'Threatens, harasses, or bullies',
+  VIOLENCE: 'Encourages or incites violence',
+  DOXING: 'Reveals personal information',
+  IMPERSONATION: 'Impersonation',
+  ILLEGAL_CONTENT: 'Illegal content',
+  SPAM: 'Spam',
+  INVOLUNTARY_PORN: 'Involuntary pornography',
+  VOTE_MANIPULATION: 'Vote manipulation',
+  BREAKS_REDDIT: 'Breaking reddit',
```
schwers: 🔕 I'm a little confused as to what this means, is this for content  that causes the interface or backend itself to break?

src/app/components/ReportingModal/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+    return (
+      <div onClick={ onClick } className='ReportingModal__option'>
+        <div className={ className } />
+        { REASONS[reason] }
+      </div>
+    );
+  }
+}
+
+ReportingModal.propTypes = {
```
schwers: Using `static` will let you move this to inside the class defintion, and be a little more consistent with other components. You should be able to write (let's assume its moved to line 39 inside the class)

```js
static propTypes = {
  onSubmit: T.func.isRequired,
}
```

phil303: woops good catch. that's a holdover from when this was a function.

src/app/components/ReportingModal/styles.less
```diff
@@ -0,0 +1,88 @@
...
+.ReportingModal {
+  height: @modalHeight;
+
+  flex-grow: 1;
+  margin: auto 20px;
+  
+  .themeify({
+    background-color: @theme-modal-body-color;
+    border: @theme-border;
+  }, {
```
schwers: Could you add the named argument `@inDayMode`, I think that will make the override clearer

phil303: yea for sure

src/app/components/Toaster/styles.less
```diff
@@ -28,7 +28,17 @@
       background-color: @ban-red;
     }
 
+    &.success {
+      background-color: @green;
```
schwers: 🔕  I was hoping this would be `@lime`

src/app/components/ReportingModal/index.jsx
phil303: this is pulled off of iOS. I don't know what it means either tbh. I'll ping some folk and see what they were thinking on it else just remove it.

src/app/components/Toaster/styles.less
phil303: it still can be! for the low low price of another commit! (I'll see what this looks like)

src/app/components/ReportingModal/styles.less
schwers: woo, thanks

src/app/pages/PostsFromSubreddit.jsx
schwers: 🔕  Nice cleanup! I like using a function for subnav, but still prefer the && :)

src/app/reducers/modal.js
```diff
@@ -0,0 +1,23 @@
+import merge from '@r/platform/merge';
```
phil303: sure

src/app/components/ReportingModal/index.jsx
phil303: just an update on this. Asked the mobile team what is meant - just waiting for a reply. https://reddit.slack.com/archives/nativemobile/p1475868251000455#

phil303:
i believe all comments addressed :haircut:

src/app/components/ReportingModal/index.jsx
schwers: ah okay, thanks for investigating!

src/app/reducers/modal.js
schwers: danke

schwers:
:fish:

phil303:
https://github.com/reddit/reddit-mobile/commit/6f446d74ec053d35a0bd4433084381423e634f51

schwers:
:party:

PR #802, opened 2016-10-05
Replace captcha with recaptcha for mweb 2x

jackniu1:
👓 @uzi @KeyserSosa 

Looks like below for recaptcha:

screen shot 2016-10-05 at 12 08 06 pm

Post request including the recaptcha response (header name will be 'g-recaptcha-response' after this PR shipped: https://github.com/reddit/node-api-client/pull/165):

screen shot 2016-10-05 at 12 09 50 pm

Thanks.

### DIFF
--- a/src/app/components/PostSubmitModal/index.jsx
+++ b/src/app/components/PostSubmitModal/index.jsx
@@ -7,7 +7,7 @@ import { Anchor, JSForm } from '@r/platform/components';
 
 import cx from 'lib/classNames';
 import Modal from '../Modal';
-import CaptchaBox from '../CaptchaBox';
+import ReCaptchaBox from '../ReCaptchaBox';
 import * as postingActions from 'app/actions/posting';
 import './styles.less';
 
@@ -16,7 +16,6 @@ const T = React.PropTypes;
 const MODAL_TITLE_TEXT = { self: 'Text', link: 'Link' };
 const SELECT_COMMUNITY = 'Select a community';
 const TITLE_PLACEHOLDER = 'Add an interesting title';
-const CAPTCHA_URL_BASE = 'https://www.reddit.com/captcha/';
 
 class PostSubmitModal extends React.Component {
   static propTypes = {
@@ -27,9 +26,9 @@ class PostSubmitModal extends React.Component {
     showCaptcha: T.bool.isRequired,
     submissionType: T.string.isRequired,
     onFieldUpdate: T.func.isRequired,
+    onRecaptchaLoaded: T.func.isRequired,
     onSubmit: T.func.isRequired,
     onCloseCaptcha: T.func.isRequired,
-    onGetNewCaptcha: T.func.isRequired,
   };
 
   render() {
@@ -38,11 +37,9 @@ class PostSubmitModal extends React.Component {
       readyToPost,
       onSubmit,
       onFieldUpdate,
-      captchaText,
-      captchaImgUrl,
+      onRecaptchaLoaded,
       showCaptcha,
       onCloseCaptcha,
-      onGetNewCaptcha,
       title: postTitle,
     } = this.props;
 
@@ -73,12 +70,9 @@ class PostSubmitModal extends React.Component {
         </JSForm>
 
         { showCaptcha ?
-          <CaptchaBox
-            captchaText={ captchaText }
-            captchaImgUrl={ captchaImgUrl }
+          <ReCaptchaBox
             onCloseCaptcha={ onCloseCaptcha }
-            onFieldUpdate={ onFieldUpdate }
-            onGetNewCaptcha={ onGetNewCaptcha }
+            onRecaptchaLoaded={ onRecaptchaLoaded }
             onSubmit={ onSubmit }
           /> :
           null }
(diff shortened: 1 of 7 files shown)


KeyserSosa:
Seems we'll have to do a synchronized cutover on this together with the backend changes or thinks will get a little break-y?

uzi:
👓  @phil303  -- he's god good context on this as he did the posts support

phil303:
hey @jackniu1! Thanks for your work here. This is pretty close but it needs a little css help.
screen shot 2016-10-05 at 2 49 48 pm

The reason this isn't centered is due to this rule: https://github.com/reddit/reddit-mobile/pull/802/files#diff-49fbb7767e22d7ff2988143e8fe0889bR31. If you remove it, that box should center right up. You can also remove the `margin: auto` underneath since the `ReCaptcha` class has a `display: inline-block` rule set and the wrapping div is `text-align: center`. Let me know if you have any questions on this!

src/app/components/ReCaptchaBox/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+    onFieldAssign,
+    onSubmit,
+  } = props;
+
+  const { recaptchaSitekey } = config;
+
+  function setRecaptchaResponse(value) {
+    onFieldAssign('gRecaptchaResponse', value);
+  }
+  setRecaptchaResponse.bind(this);
```
phil303: this actually creates a new function that you need to assign to a variable to use. Additionally, you don't even need to do this since you're not really leveraging the context variable `this`.

src/app/components/ReCaptchaBox/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <div className='ReCaptchaBox'>
+      <div onClick={ onCloseCaptcha } className='icon icon-nav-close icon-large'></div>
+      <div className='ReCaptchaBox__question'>
+        Ok, one more thing... You're human right?
+      </div>
+      <div className='ReCaptchaBox__captcha'>
+        <ReCaptcha
+          sitekey={ recaptchaSitekey }
+          onSuccess={ setRecaptchaResponse }
```
phil303: This line can just look like this:
```jsx
onSuccess={ value => onFieldAssign('gRecaptchaResponse', value) }
```

phil303:
You should be able to safely delete CaptchaBox and its associated code now.

src/app/components/ReCaptchaBox/styles.less
```diff
@@ -0,0 +1,46 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/themes/themeify';
+
+.ReCaptchaBox {
+  z-index: 3;
```
phil303: these are now defined in variables.less, you should be able to just start using `@z-index-modal`.

phil303:
@jackniu1 How do I get this working locally?

src/app/components/PostSubmitModal/index.jsx
```diff
@@ -189,20 +181,19 @@ const mapStateToProps = createSelector(
 const dispatcher = dispatch => {
   return {
     onFieldUpdate: (name, e) => dispatch(postingActions.updateField(name, e.target.value)),
+    onFieldAssign: (name, value) => dispatch(postingActions.updateField(name, value)),
```
phil303: i'm not in love with this name. What if you called it `onRecaptchaLoaded`?

phil303:
:nail_care: Pretty solid so far! Let me know if you have any questions on any of the comments left.

src/app/components/ReCaptchaBox/index.jsx
jackniu1: Done. Thanks Phil!

src/app/components/ReCaptchaBox/index.jsx
jackniu1: Cool, happy to learn how to cook with JS :)

src/app/components/ReCaptchaBox/styles.less
jackniu1: Done.

src/app/components/PostSubmitModal/index.jsx
jackniu1: Sure, Updated.

jackniu1:
💇 

Hey Phil, I updated the config.js, and setup the start.sh file with help from Joshua.

Let me know if you need these infos from me.

Thanks.

jackniu1:
@KeyserSosa , I could add some checking to by-pass mweb when deploy the r2 patch, or if we could tolerate a little bit break-y, I could deploy all the patches together.

phil303:
Just to have record of it, the open questions on this patch are:
- [ ] How do we test this end to end if we can't submit a post while on `localhost`?
- [ ] How do we handle the sitekey?

Once those are answered, I think this should be ready to go :nail_care:

KeyserSosa:
🏃 I'm way out of my league here and you're in good hands.

jackniu1:
💇 

Replace sitekey in config.js to recaptcha key.

Tried manually to verify the g-recaptcha-response, seems did not work out.
Even register could not pass the recaptcha test (not sure how register was tested with recaptcha for localhost).

I am going to deploy r2 code today, which will by-pass mweb post submit. Maybe we could do test on prod after r2 deployed.

Thanks.

jackniu1:
So r2 was deployed out today. Right now mweb post submitting was by-passed recaptcha validation.

uzi:
To test this, we'll need @KeyserSosa to temporarily allow localhost to use the production recaptcha key.  That's how we did the registration testing.

uzi:
That answered one of @phil303's questions... I don't understand the "How do we handle the sitekey?" one.  It's in the config... and a few people have access to configure which domains can use it.  That should allow testing and verification.

But yeah -- it's looking pretty decent to me once it's properly tested.

jackniu1:
Tested locally with localhost instance. 

Also added localhost to reddit.com recaptcha domain, however, this has to be tested after this patch went out.

jackniu1:
👓 @schwers 

Could you pls review the 2nd commit to upgrade the api-client
version?

Thanks.

schwers:
:fish: second commit to upgrade api-client is spot on

PR #803, opened 2016-10-05
Send loid in ad request headers

jamesmaa:
We were sending an object string because state.loid is an object. Send the actual loid from state
Before:
image
After:
image
👓 @schwers


schwers:
:fish:

jamesmaa:
b039a97c9871787171462466e442d2e2d45602c7

PR #804, opened 2016-10-06
2 x edit posts comments

schwers:
Supports editing posts / comments. 

Depends on https://github.com/reddit/node-api-client/pull/166

Let me know if the reducer structure makes sense. I'm thinking of removing the pending text from the reducer. My original thinking was it could be useful for preventing extra submits or re-filling in text-areas as they get re-rendered, but I didn't have any troubles.

I'll squash the [wip] commits with a nice message once review is done

👓  @phil303  || @uzi  || @nramadas


phil303:
hey ryan, i've gotten in the habit of posting a quick gif or pic of my feature since I think it helps re-establish context with the code. Would you be up for doing that? If you want to do a gif, there's this nifty and simple freeware I used called [licecap](http://www.cockos.com/licecap/).

src/app/components/Post/PostContent/index.jsx
```diff
@@ -216,51 +217,69 @@ export default class PostContent extends React.Component {
...
+      <div className='PostContent__edit-wrapper'>
+        <div className='PostContent__textarea-holder'>
+          <textarea
+            className='PostContent__textarea'
+            defaultValue={ selfText }
+            disabled={ editPending }
+            ref='selfTextEditor'
+          />
+        </div>
+        { editError ? this.renderEditingError(editError) : null }
```
phil303: Hard for me to tell what this looks like but do you think it'd be nicer/possibly more consistent ux to just take advantage of the error toaster here? It'd have the added advantage of trimming down the code in here a bit.

src/app/components/Comment/index.jsx
```diff
@@ -59,18 +61,90 @@ export function Comment(props) {
...
+  const {
+    editPending,
+    editError,
+    comment,
+    onToggleEditForm,
+    onUpdateBody,
+  } = props;
+
+  // NOTE: I'm not using a class for the edit form,
+  // so to get the value out of the text area I'm using fucnction refs.
```
phil303: I feel pretty strongly that this is not the right approach. A more correct approach would be to maintain textarea state via [controlled inputs](https://facebook.github.io/react/docs/forms.html#controlled-components). 

To encapsulate that idea, this should probably be its own component.

```jsx
export default class EditUserTextForm extends React.Component {
  constructor(props) {
    super(props);
    this.state = { updatedText: props.commentMD };
  }

  render() {
    return (
      <div className='Comment__editForm'>
        <div className='Comment__edit-textarea-holder'>
          <textarea
             className='Comment__edit-textarea'
             defaultValue={ props.commentMD }
             disabled={ editPending }
             onChange={ e => this.setState({ updatedText: e.target.value }) }
          />
        </div>
      // ... other stuff ...
    );
  }
}
```

Edit: considering you track changes with your reducer, it'll probably make more sense maintain update state there.

src/app/components/Post/PostContent/index.jsx
```diff
@@ -216,51 +217,69 @@ export default class PostContent extends React.Component {
...
+      return (
+        <div className='PostContent__edit-pending'>
+          Saving
+        </div>
+      );
+    }
+
+    return (
+      <div className='PostContent__edit-controls'>
+        <div className='PostContent__edit-button'onClick={ onToggleEdit }>
```
phil303: Weird because I'd think JS would freak out - you're missing a space between the string and `onClick`.

src/app/components/Post/PostContent/index.jsx
```diff
@@ -216,51 +217,69 @@ export default class PostContent extends React.Component {
...
+          Cancel
+        </div>
+        <div className='PostContent__edit-button'onClick={ this.updateSelfText }>
```
phil303: here too

src/app/components/Post/PostContent/index.jsx
```diff
@@ -216,51 +217,69 @@ export default class PostContent extends React.Component {
...
-    //         <SquareButton
-    //           text='Save'
-    //           onClick={ this.saveText }
-    //         />
-    //       </div>
-    //     </div>
-    //   </div>
-    // );
+  updateSelfText() {
+    const newSelfText = this.refs.selfTextEditor.value.trim();
```
phil303: same comment here regarding the controlled inputs as above

src/app/components/Post/index.jsx
```diff
@@ -68,19 +68,29 @@ export function Post(props) {
...
     hideWhen,
     userActivityPage,
+    onToggleEdit,
     onToggleSavePost,
     onToggleHidePost,
+    onUpdateSelftext,
+    toggleExpanded,
+    toggleShowNSFW,
+    winWidth,
+    z,
```
phil303: what the heck is `z`?

src/app/components/Post/index.jsx
```diff
@@ -150,41 +164,41 @@ export function Post(props) {
...
   return createSelector(
     [
-      postIdSelector,
-      compactSeletor,
-      expandedSelector,
-      unblurredSelector,
-      singleSelector,
-      postModelSelector,
-      userSelector,
+      (state, props) => props.forceCompact || state.compact,
```
phil303: :no_bell: this looks very clean. thanks for the update! Instead of it being an array, for consistency across the app, would you object to the arguments passed being variadic?

src/app/components/Comment/index.jsx
schwers: I'm not actually tracking changes there, I'm just tracking the text that it's updating to

src/app/components/Post/index.jsx
schwers: Nope, that sounds fine to me. I forget why it was an array in the first place.

schwers:
Here's some gifs

editingstuff

editingerrors

src/app/components/Post/PostContent/index.jsx
schwers: I thought about it, but I'm not a fan of the toaster for form errors. I think its clearer UX when the error is presented in the form that had an error.

phil303: Mind if we pull in Derek on this? Should probably be his call anyway.

schwers: I don't mind, I'll ask him.

src/app/components/Post/PostContent/index.jsx
```diff
@@ -111,19 +117,14 @@ export default class PostContent extends React.Component {
 
     // we want things to autoplay in expanded compact mode
     this.togglePlaying = this.togglePlaying.bind(this);
-    this.saveText = this.saveText.bind(this);
+    this.updateSelfText = this.updateSelfText.bind(this);
```
uzi: Do we want to change things to

```togglePlaying = (e) => { ... }```

and kill these binds?

**Comment by uzi** [comment]
I'm with @phil303 in that I thought we were trying to avoid using refs in general by using local component state to manage the value.

File: src/app/actions/comment.js [hunk 9]
```diff
@@ -2,11 +2,56 @@ import { endpoints, models, errors } from '@r/api-client';
...
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 
-const { SavedEndpoint, CommentsEndpoint } = endpoints;
+const { SavedEndpoint, CommentsEndpoint, EditUserTextEndpoint } = endpoints;
+const { ResponseError } = errors;
 
 
-export const TOGGLE_EDIT_FORM = 'COMMENT__TOGGLE_EDIT_FORM';
-export const toggleEditForm = id => ({ type: TOGGLE_EDIT_FORM, id });
+export const TOGGLE_EDIT= 'COMMENT__TOGGLE_EDIT_FORM';
```
schwers: whoops, thanks!

src/app/components/Post/index.jsx
schwers: Just a note from our discussion: `z` is zIndex and is being used further down on line 144. I'm going to clean it up naming wise, or if possible, remove it entirely, when I touch up listings in separate patch

schwers:
Gif to show new styles and toaster:

newediterrors

:haircut_man:

src/app/components/EditForm/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+        <div className='EditForm__edit-cancel' onClick={ onCancelEdit }>
+          Cancel
+        </div>
+        <div
+          className={ cx('EditForm__edit-button', {
+            disabled: editPending,
+          }) }
+          onClick={ this.saveEdit }
+        >
+          { `Updat${editPending ? 'ing' : 'e'}` }
```
phil303: fa3e19ffd513583d5f7ae60382262d9a0505d72589cd2374af2b2a7de75e057d

src/app/components/EditForm/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+    this.state = {
+      editedText: props.startingText,
+    };
+  }
+
+  saveEdit = () => {
+    if (this.props.editPending) { return; } // don't update if we're already pending
+
+    const newText = this.state.editedText.trim();
+    if (newText !== this.props.startingText) {
```
phil303: :no_bell: because I think ultimately this is a very small UX issue. Doesn't this mean that a user can hit save and nothing will happen? What if you just called `onCancelEdit` here?

phil303:
:fish:

PR #809, opened 2016-10-13
Users have to be logged in to toggle subscription

uzi:
In the case that they're not logged in, we send
then to the registration page.

👓 @schwers @phil303


schwers:
:fish:

PR #810, opened 2016-10-13
Add a "refresh this page" message to the ErrorPage

uzi:
This is for the case of an auth token refresh issue
that can occur if you require a refresh token on a
consecutive action on a page. This makes things
slightly friendlier to the user.

👓 @schwers @phil303 

What it looks like:

screen shot 2016-10-12 at 6 20 40 pm


schwers:
:fish:

PR #811, opened 2016-10-13
Reporting modal - fix accidentally reverted changes

phil303:
I accidentally overwrote some of the changes that addressed PR
comments. I did this by force pushing without fetching my latest changes
done from another computer (hecka dumb). This undoes that mistake.

The one major change this adds back is some UI changes to the modal
recommended by Derek. Instead of a snoo icon with a green background,
it's now a flag with a yellow background.

This looks like this:

screen shot 2016-10-12 at 7 29 01 pm

:eyeglasses: @schwers


schwers:
:fish:

phil303:
Thanks Ryan!

PR #812, opened 2016-10-13
Fix missing "load more" button

phil303:
Bug:
When a comment had no replies but needed a "load more" button, we were
not showing one. This was due a `replies.length` check that was
happening ahead of rendering the "load more" button.

Fix:
Allow the "load more" button to render independent of the length of
replies. This is done by making the `replies.length` check or'd with a
`loadMoreIds.length` check.

For comparison:
Before:
screen shot 2016-10-12 at 7 48 00 pm

After:
screen shot 2016-10-12 at 7 47 18 pm

Nested comments now have "load more" button as well:
screen shot 2016-10-12 at 7 47 38 pm

:eyeglasses: @uzi || @schwers

### DIFF
--- a/src/app/components/Comment/index.jsx
+++ b/src/app/components/Comment/index.jsx
@@ -108,10 +108,13 @@ function renderFooter(props) {
     preview,
   } = props;
 
+  // it's possible to have a comment with no visible replies but a load more button
+  const showReplies = comment.replies.length || comment.loadMoreIds.length;
+
   return [
     !commentDeleted ? renderTools(props) : null,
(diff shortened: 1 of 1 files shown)


phil303:
@uzi I believe this addresses https://reddit.atlassian.net/browse/MOBILEWEB-548. Would you agree or that something else going on?

uzi:
On the surface, it sounds like it.  I'll check it out in a bit.

uzi:
:fish: Yep, this takes care of one of the problems.  We're still missing the "Load more comments" at the bottom and the ability to "continue this thread", but this is a good step.  Thanks!

PR #813, opened 2016-10-13
Fix a missing `loadMoreIds` bug

phil303:
This is somewhat of a hack for the time being just to quiet the logs.
This field should have a default value of an empty array ideally. I'm
working on that version of the fix now, which will most likely be in 
api-client, but it's taking longer than expected.

:eyeglasses: @uzi


uzi:
:fish:

PR #814, opened 2016-10-13
Call UserActivity handler on the client for u/ links

schwers:
So this seems like an unintended consequence of the u/ -> user/ redirect. By using setPage we don't call the handler on the server during a redirect, which is nice because we don't do extra work fetching anything, etc. However setPage doesn't run the handler, so OTOH(or the client) we don't run the handler, so you see broken user page. I would be open to just always using setPage and letting the handler run on the server, but I thought this was a good example for discussing platform in general, hence the large 👓  @uzi @ajacksified @nramadas @phil303 

Commit message:
We started redirecting `u/*` links to `user/*` because that's the cannonical
url. The redirect was using SET_PAGE, so on the client handlers weren't being
run. This made the user page sit idlely with a loading indicator forever
when you came there via the PostHeader link or u/ links in UGC text.
As a safety measure this patch also changes the PostHeader links to use
the canonical format.

### DIFF
--- a/src/app/components/Post/PostHeader/index.jsx
+++ b/src/app/components/Post/PostHeader/index.jsx
@@ -117,7 +117,7 @@ function renderAuthorAndTimeStamp(post, single, hideWhen) {
   } = post;
 
   const authorLink = (
-    <Anchor className='PostHeader__author-link' href={ `/u/${author}` }>
+    <Anchor className='PostHeader__author-link' href={ `/user/${author}` }>
       { `u/${author}` }
     </Anchor>
   );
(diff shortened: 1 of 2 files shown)


phil303:
do you think this will be a problem for other folks as well? maybe this should be something platfor does?

schwers:
@phil303  I think this could easily come up, but was hesitant to propose a platform change at first. I kind of think `navigateToURL` should do this by default if you're on the server, as I think in most cases you'd want to the Server to issue the redirect and move onto a new request asap. I'm fine with making a patch to platform for that, but I'm not sure if it's handling `env` variables at all. Any thoughts @uzi || @nramadas  ?

schwers:
:haircut_man:

nramadas:
i think a similar issue came up in a previous discussion of how to deal with redirects off-site and dealing with status codes. i think the best option would be to have platform have a separate action that indicates redirection. for now, this is fine, but in the future, this might look like:

```
dispatch(platformActions.redirectTo(url.replace('/u', '/user'), { permanent: true });
```

Ideally, the action would know how to handle both the server and client cases.

phil303:
:fish:

PR #815, opened 2016-10-13
Honor a post's suggestedSort for the sort dropdown

uzi:
Also honor a user's preference to ignore the suggested sort.

👓 @schwers @phil303

### DIFF
--- a/src/app/components/CommentsPage/CommentsPageTools/index.jsx
+++ b/src/app/components/CommentsPage/CommentsPageTools/index.jsx
@@ -37,7 +37,15 @@ const renderThreadNotice = (post, hasSingleComment) => {
 };
 
 
-export default ({ replying, post, hasSingleComment, currentPage, id, onSortChange }) => {
+export default ({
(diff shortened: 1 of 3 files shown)


src/app/components/CommentsPage/CommentsPageTools/index.jsx
```diff
@@ -50,6 +58,7 @@ export default ({ replying, post, hasSingleComment, currentPage, id, onSortChang
   const archived = post ? post.archived : false;
   const locked = post ? post.locked : false;
   const commentingDisabled = archived || locked;
+  const suggestedSort = !preferences.ignoreSuggestedSort && post.suggestedSort;
```
schwers: 🔕  Reading through the front controller's `GET_comments` it looks like you can set a preferred default comment sort, and ignore the suggested sort. Do you know if there's anyway we can know that's happening and what the preferred sort is?

schwers:
:fish:

PR #816, opened 2016-10-13
fixup

scarow:


scarow:
👓 @phil303 -- we also need to commit the big patch. Not sure the best way to do this.

PR #819, opened 2016-10-14
Fix z-index issue with dropdowns

phil303:
In order to create a new stacking context, `position: fixed` must be
used. This was using `position: relative`. This wasn't occurring on
desktop Chrome so I missed it unfortunately.

I tested this on ios browsers Chrome and Safari and desktop browsers
Chrome, Safari, and Firefox.

:eyeglasses: @nramadas


nramadas:
:fish:

PR #821, opened 2016-10-15
Comment and post events

phil303:
Feature:
This adds comment events via the `reply` action and submit events via
the `submitPost` action.

Details:
- With comment events, you can have parent data values that are derived
from the OP's data or just a parent comment. There's some logic around
that.
- With post events, some keys change depending on whether a post is a
link or a self, so there's some logic around that as well.
- I noticed I was missing `base_url` in the `getBasePayload` function,
so I added that in.

Testing:
For comments, tested on replies to comments and posts and for posting,
tested on link and self posts.

I think you guys are doing the other events so eyeglassing you all.
:eyeglasses: @uzi @nramadas @ajacksified

### DIFF
--- a/src/app/actions/reply.js
+++ b/src/app/actions/reply.js
@@ -2,8 +2,11 @@
 // Thus all actions related to replying, opening the form, and making the reply
 // are kept in this file.
 
+import { getEventTracker } from 'lib/eventTracker';
+import { getBasePayload, buildSubredditData, convertId } from 'lib/eventUtils';
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import modelFromThingId from 'app/reducers/helpers/modelFromThingId';
+
 
 export const TOGGLE_REPLY_FORM = 'TOGGLE_REPLY_FORM';
 export const toggleReplyForm = id => ({ type: TOGGLE_REPLY_FORM, id });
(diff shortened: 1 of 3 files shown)


phil303:
also :eyeglass: @drunken-economist in case he's curious

phil303:
also :eyeglasses: @drunken-economist in case he's curious

src/app/actions/reply.js
```diff
@@ -29,4 +32,29 @@ export const reply = (id, text) => async (dispatch, getState) => {
   // TODO the replyable mixin should just be returning a CommentModel from this...
   const reply = apiResponse.getModelFromRecord(apiResponse.results[0]);
   dispatch(replied(id, reply));
+
+  logReply(reply, getState());
 };
+
+
+function logReply(reply, state) {
+  // If the parent starts with a t1, then its parent is a comment. If not,
```
drunken-economist: This seems shaky. Is there another way to determine the type of the parent? Local and staging installs, for example, don't always use t1/t3 as comment/post prefixes

drunken-economist:
🐟  on the events

src/app/actions/reply.js
phil303: I totally agree. We've got this kind of stuff hardcoded all over though so this wouldn't be the first place it breaks unfortunately. It hasn't been a problem yet since we're pointing at prod data . I'll look through the state to see if there's another way though. Thanks for the call out.

uzi:
:fish:

PR #822, opened 2016-10-17
Increase toaster font size

phil303:
Seeing the toaster pop up on my iPhone 6, I felt like I was squinting
to read the text. Increasing the font-size up to 15px puts it at a
comfortable reading level. This is especially important with the
toaster because it auto-dismisses after 3 seconds.

I tested this down to iPhone 4 dimensions with some of the longer
messages. Sometimes the toaster text ended up taking 3 lines but still
looked pretty good even in those cases.

I removed the right padding on the text because there's already padding
around the close button to right, which was making it look like there
was too much padding in total between the two elements.

### DIFF
--- a/src/app/actions/posts.js
+++ b/src/app/actions/posts.js
@@ -87,7 +87,7 @@ export const failedUpdateSelfText = (postId, error) => ({
   type: FAILED_UPDATE_SELF_TEXT,
   thingId: postId,
   error,
-  message: 'Sorry, something went wrong with updating your post',
+  message: 'Sorry, something went wrong with updating your post.',
 });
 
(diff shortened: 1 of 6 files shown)


phil303:
:eyeglasses: @schwers

schwers:
🔕  Interestingly, [Apple's UI Guidline for Font](https://developer.apple.com/ios/human-interface-guidelines/visual-design/typography/) suggests 16px for a `Callout` and 15px for a `Subhead`. This is perhaps a moot point as we don't follow they're guidlines for Body size, but I wouldn't even be opposed to 16px here if you felt so inclined.

:fish:

phil303:
16px would be pushing the limits of what fits. Anything more than 4 lines with 15px overflowed the toaster (since it's a fixed height).

PR #823, opened 2016-10-17
Normalize all input font sizes to prevent zooming

phil303:
iOS will automatically zoom in when an input is selected but not reset
the zoom level when the user is done. This allows for horizontal panning
which is not expected behavior from a UX perspective. (Also, it annoys
the crap out of me :P)

Rationale:
I'm normalizing all inputs because I'd never expect a dev of any
experience level to know/remember this while writing an input
component. By doing this everywhere, it's just taken care of. And b/c
it's css, it's easy to overwrite should the need arise.

It's pretty annoying that iOS forces this but it still looks pretty good
so I feel more ok making a universal rule. BUT in general, this kind of
thing screams "stop and think about what you're doing here" which is why
it's an RFC.

Details:
- This applies to text, number, and password inputs and also textareas.
I specified types on the inputs to limit the scope of change (there's a
ton of input types that I don't want to mess with).

:eyeglasses: @schwers

### DIFF
--- a/src/app/components/CaptchaBox/styles.less
+++ b/src/app/components/CaptchaBox/styles.less
@@ -51,7 +51,6 @@
       width: 100%;
       padding: 14px;
       border-radius: 0;   // handle ios safari's native border rounding
-      font-size: 16px;
 
       .themeify({
         color: @theme-body-text-color;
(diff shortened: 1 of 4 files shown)


schwers:
Hey @phil303, thanks for doing this! I haven't been on iOS for a while, but just from the few times it's happened to me I agree that is a really annoying UX. I'm in favor of doing this -- it's a shame it has to be global, but I think that's the write approach in this situation. 😢  🐟

phil303:
thanks for looking at this ryan!

PR #824, opened 2016-10-17
Mount default VCL instead of copying over.

wting:
By mounting the file we can guarantee that Docker doesn't cache the
config file (default.vcl) and will always use the newest one.

:eyeglasses: @spladug 

cc: @gtaylor @telaviv

### DIFF
--- a/Dockerfile.varnish
+++ b/Dockerfile.varnish
@@ -1,5 +1,7 @@
 FROM quay.io/reddit/varnish-fastly
 
+VOLUME ["/etc/varnish"]
+
 # FIXME(wting|2016-09-13): Port overrides aren't working, Varnish is still
 # running on port 80 instead.
 ENV VARNISH_PORTS 8080
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,5 +14,7 @@ services:
       dockerfile: Dockerfile.varnish
     ports:
     - "4301:80"
+    volumes:
+    - .:/etc/varnish
     depends_on:
     - mweb


Dockerfile.varnish
```diff
@@ -1,5 +1,7 @@
 FROM quay.io/reddit/varnish-fastly
 
+VOLUME ["/etc/varnish"]
```
spladug: if this is about caching with `default.vcl`, would it work to do `COPY default.cvl ...` here? not super against doing this with the volume, but if there are any tempfiles etc. generated they clutter up the host which is kinda a pain.

spladug:
:nail_care: just one question

Dockerfile.varnish
wting: Hmm, so the current implementation uses [`COPY` anyway](https://github.com/reddit/varnish-fastly/blob/master/Dockerfile#L55) which is causing the issue.

I overrode that with an explicit `COPY` locally but the old config is still being cached:
```shell
╭─wting@nuc ~/code/reddit-mobile ‹python-2.7.12› ‹wting_mount_vcl×9c4ec87›
╰─➤  g sd                                                                                                                   2016.10.17 13:31:41 PDT 
## wting_mount_vcl...origin/wting_mount_vcl
 M Dockerfile.varnish
 M default.vcl
 M docker-compose.yml
diff --git a/Dockerfile.varnish b/Dockerfile.varnish
index 31ec7f2..359a84b 100644
--- a/Dockerfile.varnish
+++ b/Dockerfile.varnish
@@ -1,6 +1,6 @@
 FROM quay.io/reddit/varnish-fastly
 
-VOLUME ["/etc/varnish"]
+COPY ./default.vcl /etc/varnish/default.vcl
 
 # FIXME(wting|2016-09-13): Port overrides aren't working, Varnish is still
 # running on port 80 instead.
diff --git a/default.vcl b/default.vcl
index 657c4b3..d158943 100644
--- a/default.vcl
+++ b/default.vcl
@@ -15,11 +15,9 @@ sub vcl_recv {
   # easy programmatic testing.
   if (req.http.Cookie !~ "(^|;\s*)(mweb-no-redirect=1)(;|$)"
       && (req.http.X-UA-Device ~ "^mobile-" || req.http.X-UA-Device ~ "^tablet-")) {
-    # set req.backend = mweb;
-    error 444;
+    set req.backend = mweb;
   } else {
-    # set req.backend = www;
-    error 451;
+    set req.backend = mweb;
   }
 }
 
diff --git a/docker-compose.yml b/docker-compose.yml
index 1868ac4..81f9c69 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,5 @@ services:
       dockerfile: Dockerfile.varnish
     ports:
     - "4301:80"
-    volumes:
-    - .:/etc/varnish
     depends_on:
     - mweb
╭─wting@nuc ~/code/reddit-mobile ‹python-2.7.12› ‹wting_mount_vcl×9c4ec87›
╰─➤                                                                                                                         2016.10.17 13:31:47 PDT 
```

I'm working around this issue by mounting the source directory instead, but fortunately Varnish doesn't create any tempfiles in `/etc/varnish/`.

wting:
:haircut_woman:

spladug:
weird! :fish:

PR #826, opened 2016-10-17
Serve blank ads

jamesmaa:
👓 @dwick 

Manually tested:
- Checking a subreddit with blank ads
- Checking a subreddit with normal ads


dwick:
what's the purpose of this? 💅

dwick:
🐟 assuming impression pixels are fired properly

jamesmaa:
they should be. Empirically conifrmed and that's why I inserted an empty div, so `checkImpression` works properly

PR #827, opened 2016-10-17
Experiment fix: look up recommended subreddits by postId

scarow:
If there is no subredditName in the urlParams, look up the
subreddit by postId, then fetch the recommended subreddit objects

These two urls should yield the same recommended subreddits:

[http://localhost:4444/r/relationships/comments/559d7x/this_is_a_difficult_question_to_ask_because_i_35m?feature_experimentrecommendedtopplain=#](http://localhost:4444/r/relationships/comments/559d7x/this_is_a_difficult_question_to_ask_because_i_35m?feature_experimentrecommendedtopplain=#)

[http://localhost:4444/comments/559d7x/this_is_a_difficult_question_to_ask_because_i_35m?feature_experimentrecommendedtopplain=#](http://localhost:4444/comments/559d7x/this_is_a_difficult_question_to_ask_because_i_35m?feature_experimentrecommendedtopplain=#)

👓 @uzi @phil303 @schwers

### DIFF
--- a/src/app/pages/Comments.jsx
+++ b/src/app/pages/Comments.jsx
@@ -65,10 +65,15 @@ const stateProps = createSelector(
     const postLoaded = !!post;
     const replying = currentPage.queryParams.commentReply === commentsPageParams.id;
 
-    const recommendedSubredditNames = recommendedSrs[currentPage.urlParams.subredditName] || [];
-    const recommendedSubreddits = recommendedSubredditNames.map(name => subreddits[name]);
+    let recommendedSubredditNames = [];
+    if (post && post.subreddit in recommendedSrs) {
+      recommendedSubredditNames = recommendedSrs[post.subreddit];
+    }
 
-    const currentSubreddit = subreddits[currentPage.urlParams.subredditName];
+    const recommendedSubreddits = recommendedSubredditNames.map(name => subreddits[name]);
+    // since we return early in the render function if !postLoaded, it's OK
+    // for currentSubreddit to be null here
+    const currentSubreddit = post ? post.subredditDetail : null;
 
     return {
       op: postLoaded ? post.author : '',
--- a/src/app/router/handlers/CommentsPage.js
+++ b/src/app/router/handlers/CommentsPage.js
@@ -59,7 +59,7 @@ export default class CommentsPage extends BaseHandler {
 
     const commentsPageParams = CommentsPage.pageParamsToCommentsPageParams(this);
     const commentsPageId = paramsToCommentsPageId(commentsPageParams);
-    const { subredditName } = state.platform.currentPage.urlParams;
+    let { subredditName } = state.platform.currentPage.urlParams;
 
     await Promise.all([
       dispatch(commentsPageActions.fetchCommentsPage(commentsPageParams)),
@@ -67,11 +67,18 @@ export default class CommentsPage extends BaseHandler {
 
       dispatch(commentsPageActions.fetchRelevantContent()),
       dispatch(commentsPageActions.visitedCommentsPage(this.urlParams.postId)),
-      fetchRecommendedSubreddits(state, dispatch, subredditName),
-
       dispatch(subredditActions.fetchSubreddit(subredditName)),
+
     ]);
 
+    // if url does not include r/subredditName, then subredditName will be
+    // undefined, even if we are on a comments page. We can ascertain the
+    // subredditName by looking it up via postId
+    if (!subredditName) {
+      subredditName = getState().posts[`t3_${state.platform.currentPage.urlParams.postId}`].subreddit;
+    }
+
+    fetchRecommendedSubreddits(state, dispatch, subredditName);
     dispatch(setStatus(getState().commentsPages[commentsPageId].responseCode));
 
     logClientScreenView(buildScreenViewData, getState());


src/app/router/handlers/CommentsPage.js
```diff
@@ -59,19 +59,25 @@ export default class CommentsPage extends BaseHandler {
...
     await Promise.all([
       dispatch(commentsPageActions.fetchCommentsPage(commentsPageParams)),
       fetchUserBasedData(dispatch),
 
       dispatch(commentsPageActions.fetchRelevantContent()),
       dispatch(commentsPageActions.visitedCommentsPage(this.urlParams.postId)),
-      fetchRecommendedSubreddits(state, dispatch, subredditName),
-
+      // to do: subredditName is undefined when the url omits r/subredditName
+      // this will likely have weird side effects on state in the future
```
scarow: Not sure if I should leave this here, but we definitely should investigate if this is the appropriate behavior. It seems like url should NOT determine state (i.e it shouldnt matter if we have `r/relationships` in the url or not -- we should still load that subreddit's data)

schwers: This `Promise.all` was intended to guard `logClientScreenView` -- so I think it's okay to omit this comment, but I'd add a note about the url to line ~76 where you're checking `!subreddit`. In general I agree and really dislike that we have to do this kind of stuff just because of the url being a little bit different. 

Out-of-scope-sidenote: This seems like a nice example of why it would be nice to have a GraphQL or Falcor api -- your overall query for the page could ask for `recommended subreddits based on this post's subreddit` and let the backend figure out the correct subreddit to give recommendations for.

src/app/router/handlers/CommentsPage.js
```diff
@@ -59,19 +59,25 @@ export default class CommentsPage extends BaseHandler {
...
       dispatch(subredditActions.fetchSubreddit(subredditName)),
+
     ]);
 
+    if (!subredditName){
```
schwers: nit: space before bracket

src/app/router/handlers/CommentsPage.js
```diff
@@ -59,19 +59,25 @@ export default class CommentsPage extends BaseHandler {
...
+      subredditName = getState().posts['t3_' + state.platform.currentPage.urlParams.postId].subreddit;
```
schwers: nit: use a template string instead of `+`, i.e. ```js [`t3_${state.platform.currentPage.urlParams.postId`] ```

src/app/router/handlers/CommentsPage.js
```diff
@@ -118,7 +124,21 @@ const fetchRecommendedSubreddits = (state, dispatch, subredditName) => {
...
+  await dispatch(recommendedSubredditsActions.fetchRecommendedSubreddits(subredditName, max_recs));
+
+  const experimentEnabled = (
+    feature.enabled(VARIANT_RECOMMENDED_TOP_PLAIN) ||
+    feature.enabled(VARIANT_RECOMMENDED_TOP) ||
+    feature.enabled(VARIANT_RECOMMENDED_BOTTOM)
+  );
+
+  // only fetch subreddits if there are no subreddits && we are in the experiment
+  if (state.subreddit && Object.keys(state.subreddit).length < 1 && experimentEnabled){
```
schwers: What's `state.subreddit`? Should this be`state.subreddits`?

schwers:
💅

scarow:
I was actually able to remove a bunch of cruft -- it looks like the main issue was that `subredditName` was undefined.

💇

src/app/pages/Comments.jsx
```diff
@@ -53,10 +53,13 @@ const stateProps = createSelector(
     const postLoaded = !!post;
     const replying = currentPage.queryParams.commentReply === commentsPageParams.id;
 
-    const recommendedSubredditNames = recommendedSrs[currentPage.urlParams.subredditName] || [];
-    const recommendedSubreddits = recommendedSubredditNames.map(name => subreddits[name]);
+    let recommendedSubredditNames = [];
+    if (post && post.subreddit in recommendedSrs ){
```
phil303: nit: spacing - `if (post && post.subreddit in recommendedSrs) {`

src/app/pages/Comments.jsx
```diff
@@ -53,10 +53,13 @@ const stateProps = createSelector(
...
-    const recommendedSubredditNames = recommendedSrs[currentPage.urlParams.subredditName] || [];
-    const recommendedSubreddits = recommendedSubredditNames.map(name => subreddits[name]);
+    let recommendedSubredditNames = [];
+    if (post && post.subreddit in recommendedSrs) {
+      recommendedSubredditNames = recommendedSrs[post.subreddit];
+    }
 
-    const currentSubreddit = subreddits[currentPage.urlParams.subredditName];
+    const recommendedSubreddits = recommendedSubredditNames.map(name => subreddits[name]);
+    const currentSubreddit = post ? post.subredditDetail : null;
```
phil303: are you going to have a problem if this is null? it looks like [this line](https://github.com/reddit/reddit-mobile/blob/2X/src/app/components/RecommendedSubreddits/index.jsx#L40) depends on it being an object with a `url` attr

phil303:
:nail_care:

src/app/pages/Comments.jsx
phil303: ah, no that's probably fine for now. what do you think about just leaving a comment here?

phil303:
:fish:

schwers:
:fish:

PR #828, opened 2016-10-17
Support GTM and adblock events

schwers:
Hey all, 

This Pr started out as porting GTM from 1x. Including `adblock` events seemed to make sense, as it's sort of related. Happy to break it out into two separate PRs. For ease of reviewing / reading I'll leave links to the individual commits so it's easier to tell whats-what (though it's probably best to leave comments on this PR itself vs the individual commits)

GTM notes: `frames.js` was copied wholesale from 1X, `gtm.js` and `dom.js` were copied as well but they have some minor cleanup.

GTM Patch [link](https://github.com/reddit/reddit-mobile/commit/8493ea7f6ce42812d48e43e84293ac3a8d1ce3e3)
[GTM Jira Mobileweb-492](https://reddit.atlassian.net/browse/MOBILEWEB-492)

Adblock Patch [link](https://github.com/reddit/reddit-mobile/commit/8493ea7f6ce42812d48e43e84293ac3a8d1ce3e3)
[Adblock Jira Mobileweb-696](https://reddit.atlassian.net/browse/MOBILEWEB-696)

👓  @nramadas  || @phil303 
👓  @dwick

### DIFF
--- a/src/app/actions/ads.js
+++ b/src/app/actions/ads.js
@@ -5,6 +5,7 @@ import config from 'config';
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import isFakeSubreddit from 'lib/isFakeSubreddit';
 import adLocationForPostRecords from 'lib/adLocationForPostRecords';
+import { logClientAdblock } from 'lib/eventUtils';
 
 
 const { PostsEndpoint } = endpoints;
@@ -67,6 +68,14 @@ const trackAdPost = post => {
   });
 };
 
+// Used by Rendered Ads if they detect that adblock has hidden
+// they're content node. This only a thunk'd action because
+// `logClientAdblock` needs access to state
+// placementIndex -- the index in the post list
+export const trackAdHidden = placementIndex => async (dispatch, getState) => {
+  logClientAdblock('element-hidden', placementIndex, getState());
+};
+
 // For views that need to fetch ads, we need an id to track them.
 // We could use the id of the view (e.g. postsLists have a postsListId),
 // but maybe every view won't have an id like that or there might be collisions.
@@ -122,13 +131,14 @@ export const fetchAddBasedOnResults = async (dispatch, state, adId, postsList, p
   // on the page.
   const dt = postsList.results.map(record => record.uuid);
   const site = getSite(pageParams);
+  const placementIndex = adLocationForPostRecords(postsList.results);
 
   const data = {
     site,
     dt: dt.join(','),
     platform: 'mobile_web',
-    placement: `feed-${adLocationForPostRecords(postsList.results)}`,
-    raw_json: '1',
+    placement: `feed-${placementIndex}`,
+    raw_json: 1,
   };
 
   if (!state.session.accessToken) {
@@ -143,9 +153,15 @@ export const fetchAddBasedOnResults = async (dispatch, state, adId, postsList, p
     if (ad === null) {
       return dispatch(noAd(adId));
     }
-    
+
     dispatch(received(adId, ad));
   } catch (e) {
+    // A `status` of `0` means the request wasn't actually finished, likely
+    // due to adblock.
+    if (e.status === 0) {
+      logClientAdblock('endpoint-blocked', placementIndex, state);
+    }
+
     if (e instanceof ResponseError) {
       dispatch(failed(adId, e));
     } else {
--- a/src/app/components/Ad/index.jsx
+++ b/src/app/components/Ad/index.jsx
@@ -4,11 +4,27 @@ import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
 import throttle from 'lodash/throttle';
 
+import { isHidden } from 'lib/dom';
+
 import * as adActions from 'app/actions/ads';
 import { TOP_NAV_HEIGHT } from 'app/constants';
 import Post from 'app/components/Post';
 
+const T = React.PropTypes;
+
 class Ad extends React.Component {
+  static propTypes = {
+    // ownProps
+    adId: T.string.isRequired,
+    placementIndex: T.number.isRequired,
+    postProps: T.object.isRequired,
+
+    // props from connect
+    adTracked: T.bool.isRequired,
+    onAdMadeImpression: T.func.isRequired,
+    onAdblockDetected: T.func.isRequired,
+  };
+
   constructor(props) {
     super(props);
 
@@ -19,6 +35,7 @@ class Ad extends React.Component {
     if (!this.props.adTracked) {
       window.addEventListener('scroll', this.scrollListener);
       this.checkImpression(); // check manually in case the ad is visible on the first render
+      this.checkAdblock();
     }
   }
 
@@ -43,6 +60,12 @@ class Ad extends React.Component {
     }
   }
 
+  checkAdblock() {
+    if (isHidden(findDOMNode(this))) {
+      this.props.onAdblockDetected();
+    }
+  }
+
   render() {
     return <Post { ...this.props.postProps } />;
   }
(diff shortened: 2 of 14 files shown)


src/app/components/Ad/index.jsx
```diff
@@ -4,11 +4,28 @@ import { connect } from 'react-redux';
...
+import { logClientAdblock } from 'lib/eventUtils';
+
 import * as adActions from 'app/actions/ads';
 import { TOP_NAV_HEIGHT } from 'app/constants';
 import Post from 'app/components/Post';
 
+const T = React.PropTypes;
+
 class Ad extends React.Component {
+  static PropTypes = {
```
phil303: lowercase `p`!

src/app/components/Ad/index.jsx
```diff
@@ -43,18 +61,37 @@ class Ad extends React.Component {
...
+
   render() {
     return <Post { ...this.props.postProps } />;
   }
 }
 
 const selector = createSelector(
   (state, props) => state.adRequests[props.adId],
-  adRequest => ({ adTracked: adRequest.impressionTracked }),
+  state => state, // adblock events use the current state to make their event payload
```
nramadas: instead of selecting the whole state, what do you think of firing a thunk'd action? That way, the action has access to `getState` and can pass it along to `logClientAdblock`, and we don't have to worry about the ad re-rendering every time state changes.

schwers:
👓  @drunken-economist for adblock events. The biggest changes to events are:

1) the base event payload now tracks `adblock`: true | false for 'is adblock enabled?'
2) There are now adblock events for the api call being blocked and the ad div being hidden https://github.com/reddit/reddit-mobile/pull/828/files#diff-cfa1101baacc59b9f536610e1c17146cR126

nramadas:
Also, just so i have an understand of how often we track:
0. When we load up the gtm unit, we track if adblock is on
0. For each ad we render, we will also track if adblock is on
0. We also track if adblock is on when fetching the data for an ad unit

Is that correct?
:haircut_woman:

src/app/components/GoogleTagManager/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+        frame.referrer = 'no-referrer';
+        frame.id = '${GTM_JAIL_ID}';
+        frame.name = JSON.stringify({
+          subreddit: '${props.subredditName || ''}',
+          origin: location.origin,
+          pathname: location.pathname,
+        });
+        frame.src = '//${props.mediaDomain}/gtm/jail?id=${props.googleTagManagerId}';
+        document.body.appendChild(frame);
+        window.frameRef = frame;
```
phil303: is `frameRef` by gtm?

schwers: oh whoops, debugging code I left in

schwers:
@nramadas 

I think it's slightly different:

1) GTM tracks page views every time we would track a screen view
2) The base event payload now tracks if adblock is enabled
3) For each ad render, we log an event if adblock hid or removed the ad unit
4) For each ad api-call, we log an event if we think adblock blocked the api call

src/app/components/AdsTracking/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+import config from 'config';
+import GoogleTagManager from 'app/components/GoogleTagManager';
+
+// This component renders things that are needed for ads tracking.
+// Right now this is:
+// * GoogleTagManager -- used for tracking screenviews for ad targetting
+// * AdblockTester -- used to determine if adblock is enabled, this info is
+//    included on the base events payload
+export const AdTracking = props => {
```
phil303: nit this file is called Ad**s**Tracking

src/app/components/Ad/index.jsx
schwers: 🔕  I made the change but I really dislike it, it seems rather misleading that we're calling dispatch without actually needing to dispatch -- maybe that's just because we don't have other cases of this in the codebase yet. I wish we didn't really have to worry about it either as there's only one ad unit per-page, and it just renders a connected `Post`. I would hope that because the Post is connected, and does its own memoization, then we wouldn't have to really worry about the re-render, but Post's selector isn't thunk'd so this seems like a good safeguard.

src/app/components/GoogleTagManager/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+        frame.style.display = 'none';
+        frame.referrer = 'no-referrer';
+        frame.id = '${GTM_JAIL_ID}';
+        frame.name = JSON.stringify({
+          subreddit: '${props.subredditName || ''}',
+          origin: location.origin,
+          pathname: location.pathname,
+        });
+        frame.src = '//${props.mediaDomain}/gtm/jail?id=${props.googleTagManagerId}';
```
phil303: do the docs say to use protocol relative urls here? Generally you want to use https [when possible](https://www.paulirish.com/2010/the-protocol-relative-url/)

src/app/components/PostsList/index.jsx
```diff
@@ -51,7 +51,7 @@ const renderPostsList = props => {
   const { postRecords, ad, adId, forceCompact, subredditIsNSFW } = props;
   const records = ad ? recordsWithAd(postRecords, ad) : postRecords;
 
-  return map(records, postRecord => {
+  return map(records, (postRecord, index) => {
```
phil303: you can use the native `records.map` here

src/app/components/GoogleTagManager/index.jsx
schwers: There aren't really docs around this afaik, I agree but have been hesitant to change the protocol as the iframe is used for `postMessage`'ing, which can have some sensitivities to protocol mismatching. I don't see any signs of an error when I specify `https`, @dwick do you have any thoughts on the matter or know of a good way to debug this?

schwers:
💅

schwers:
:haircut_man:

phil303:
@schwers [this stuff](https://github.com/reddit/reddit-mobile/pull/828#issuecomment-254598212) would make a great addition to your commit description

src/app/components/AdblockTester/index.jsx
```diff
@@ -0,0 +1,25 @@
...
+import config from 'config';
+
+const { adblockTestClassName } = config;
+
+// Renders a component, hidden offscreen, used to detect if adblock is
+// enabled. If it's enabled, the div will either be removed entirely from
+// the dom, or hidden via css or style modifications. The checks
+// for being hidden currently live in `lib/eventUtils`
+
+export default function AdblockTester() {
```
schwers: `export default` is the one case I don't like it for because it won't turn into a named function :(

phil303:
:fish: this looks kosher to me. @dwick, can you take pass at this as well?

src/app/actions/ads.js
```diff
@@ -122,13 +131,13 @@ export const fetchAddBasedOnResults = async (dispatch, state, adId, postsList, p
...
   const dt = postsList.results.map(record => record.uuid);
   const site = getSite(pageParams);
+  const placementIndex = adLocationForPostRecords(postsList.results);
 
   const data = {
     site,
     dt: dt.join(','),
     platform: 'mobile_web',
-    placement: `feed-${adLocationForPostRecords(postsList.results)}`,
-    raw_json: '1',
```
dwick: is this not needed for the api anymore? i seem to remember weird things with escaping without this

schwers: Almost certainly needed, nice catch! I haven't noticed any issues locally but I'm hesistant to say that makes it okay, I'll add it back.

src/app/components/GoogleTagManager/index.jsx
dwick: https should work. to test it just ensure that the comscore pixel is being fired:

screen shot 2016-10-19 at 11 55 49 am

dwick:
This looks good.  Only thing missing i think is the ads holdout group: https://github.com/reddit/reddit-mobile/commit/bc938475f4bcf5c369e3df817c8e23dcdedbbfcc

I don't think that'd be a problem to add later. 🐟

drunken-economist:
🐟 on the events!

PR #831, opened 2016-10-18
Add crawl event tracking.

telaviv:
This is the event spec: https://reddit.atlassian.net/wiki/display/DA/Search+Engine+Crawl+Events

The two things I had a hard time getting were the response time, and the current server they were running on. I don't know if the concept of a server is still applicable to mweb though.

### DIFF
--- a/src/app/router/handlers/CommentsPage.js
+++ b/src/app/router/handlers/CommentsPage.js
@@ -8,10 +8,10 @@ import * as commentsPageActions from 'app/actions/commentsPage';
 import * as subredditActions from 'app/actions/subreddits';
 import * as recommendedSubredditsActions from 'app/actions/recommendedSubreddits';
 import { flags } from 'app/constants';
-import { getBasePayload, buildSubredditData, convertId, logClientScreenView } from 'lib/eventUtils';
 import features from 'app/featureFlags';
 import { paramsToCommentsPageId } from 'app/models/CommentsPage';
 import { setTitle } from 'app/actions/pageMetadata';
+import { convertId, trackPageEvents } from 'lib/eventUtils';
 
 const {
   VARIANT_RECOMMENDED_TOP_PLAIN,
@@ -96,7 +96,8 @@ export default class CommentsPage extends BaseHandler {
 
     dispatch(setTitle(this.buildTitle(getState(), commentsPageId)));
 
-    logClientScreenView(buildScreenViewData, getState());
+    const latestState = getState();
+    trackPageEvents(latestState, buildAdditionalEventData(latestState));
   }
 }
 
@@ -115,7 +116,7 @@ const fetchRecommendedSubreddits = (state, dispatch, subredditName) => {
   dispatch(recommendedSubredditsActions.fetchRecommendedSubreddits(subredditName, max_recs));
 };
 
-function buildScreenViewData(state) {
+function buildAdditionalEventData(state) {
   const { currentPage: { queryParams, urlParams } } = state.platform;
   const fullName =`t3_${urlParams.postId}`;
   const post = state.posts[fullName];
@@ -126,12 +127,12 @@ function buildScreenViewData(state) {
 
   return cleanObject({
     target_fullname: fullName,
+    post_fullname: fullName,
     target_id: convertId(post.id),
     target_type: post.isSelf ? 'self' : 'link',
     target_sort: queryParams.sort || 'confidence',
     target_url: post.cleanUrl,
     target_filter_time: queryParams.sort === 'top' ? (queryParams.time || 'all') : null,
-    ...buildSubredditData(state),
-    ...getBasePayload(state),
+    target_created_ts: post.createdUTC * 1000,
   });
 }
--- a/src/app/router/handlers/PostsFromSubreddit.js
+++ b/src/app/router/handlers/PostsFromSubreddit.js
@@ -9,7 +9,7 @@ import { FAKE_SUBS } from 'lib/isFakeSubreddit';
 import { cleanObject } from 'lib/cleanObject';
 import { listingTime } from 'lib/listingTime';
 import { fetchUserBasedData } from './handlerCommon';
-import { getBasePayload, buildSubredditData, convertId, logClientScreenView } from 'lib/eventUtils';
+import { convertId, trackPageEvents } from 'lib/eventUtils';
 
 import { setTitle } from 'app/actions/pageMetadata';
 
@@ -64,7 +64,8 @@ export default class PostsFromSubreddit extends BaseHandler {
     dispatch(setStatus(getState().postsLists[postsListId].responseCode));
     dispatch(setTitle(this.buildTitle(getState(), subredditName)));
 
-    logClientScreenView(buildScreenViewData, getState());
+    const latestState = getState();
+    trackPageEvents(latestState, buildAdditionalEventData(latestState));
   }
 }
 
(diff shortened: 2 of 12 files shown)


telaviv:
👓 @phil303 @prashtx

src/lib/eventUtils.js
```diff
@@ -13,10 +14,10 @@ export function convertId(id) {
 }
 
 
-export function buildSubredditData(state) {
+function buildSubredditData(state) {
   const { subredditName } = state.platform.currentPage.urlParams;
 
-  if (isFakeSubreddit(subredditName)) { return {}; }
+  if (!subredditName || isFakeSubreddit(subredditName)) { return {}; }
```
phil303: :no_bell: This check is actually handled inside the `isFakeSubreddit` function but I like the clarity here since that's not at all obvious from the `isFakeSubreddit` name.

src/lib/eventUtils.js
```diff
@@ -77,14 +103,19 @@ export function logClientScreenView(buildScreenViewData, state) {
   if (urlToken !== lastUrlToken) {
```
phil303: I think it'd be a little clearer if this block read like this:
```js
if (urlToken !== lastUrlToken) {
  lastUrlToken = urlToken;
  return false;
}
return true;
```

Generally speaking, side effects and return values should never mix but I can't figure out how to make that not the case here without tearing up this function. Plus it's a hack that's going away soon so whatevs for now.

src/lib/eventUtils.js
```diff
@@ -28,23 +29,30 @@ export function buildSubredditData(state) {
 }
 
 
-export function getBasePayload(state) {
+function getBasePayload(state) {
```
phil303: In other words, if it was possible to implicitly include this on every event (for instance, [comments and posts](https://github.com/reddit/reddit-mobile/pull/821) could use this) instead of having to import this in, I think that'd be an improvement.

src/lib/eventUtils.js
```diff
@@ -55,15 +63,33 @@ export function getBasePayload(state) {
     payload.loid_created = state.loid.loidCreated;
   }
 
-  return payload;
+  getEventTracker().track('screenview_events', 'cs.screenview_mweb', payload);
 }
 
 
+function trackCrawlEvent(state, additionalEventData) {
+  const {protocol, method, crawler, userAgent } = state.meta;
```
phil303: nit space between curly and protocol

src/lib/eventUtils.js
```diff
@@ -1,3 +1,4 @@
+import pick from 'lodash/pick';
```
phil303: unused

src/lib/eventUtils.js
```diff
@@ -28,23 +29,30 @@ export function buildSubredditData(state) {
...
-  const domain = window.location.host;
-
   const userAccount = user.loggedOut ? null : accounts[user.name];
 
   const payload = {
-    domain,
     geoip_country: state.meta.country,
-    user_agent: state.meta.userAgent,
-    referrer_domain: url.parse(referrer).host || domain,
+    referrer_domain: url.parse(referrer).host || state.meta.domain,
```
phil303: a lot of this stuff is used in other events and is intended to be a "base" payload. Your changing of the `getBasePayload` function for the purposes of this event is going to cause issues with the other events patches coming (which will have a higher priority since they block 100% launch of 2x)

phil303:
:nail_care: if you have questions on this stuff let me know! I'm guessing the intention of those eventUtil functions wasn't obvious since the other event patches haven't landed yet and they were only being used in screenview events.

src/app/router/handlers/CommentsPage.js
```diff
@@ -132,12 +132,12 @@ function buildScreenViewData(state) {
...
     target_fullname: fullName,
+    post_fullname: fullName,
     target_id: convertId(post.id),
     target_type: post.isSelf ? 'self' : 'link',
     target_sort: queryParams.sort || 'confidence',
     target_url: post.cleanUrl,
     target_filter_time: queryParams.sort === 'top' ? (queryParams.time || 'all') : null,
-    ...buildSubredditData(state),
-    ...getBasePayload(state),
+    target_created_ts: post.createdUTC * 1000,
```
telaviv: These are in the crawl event spec. I asked justin about putting extra data in the payload and he said it was fine as long as we documented them, which I planned to do .

telaviv:
@phil303 alright so i finally rebased and it was crazy! The overall shape of the code is the same, but I left `getBasePayload` mostly unchanged. 

@prashtx The crawler header scheme we talked about offline is implemented here.

telaviv:
💇

src/app/router/handlers/CommentsPage.js
```diff
@@ -98,7 +98,7 @@ export default class CommentsPage extends BaseHandler {
 
     dispatch(setTitle(this.buildTitle(getState(), commentsPageId)));
 
-    logClientScreenView(buildScreenViewData, getState());
+    trackPageEvents(getState(), buildAdditionalEventData(state));
```
phil303: So you have to be careful with `state` here. Every time you dispatch something it replaces the state object wholesale. Using `getState` ensures you have the latest version of the state. what you should do here is set up something like `const latestState = getState()` then pass that into your function

src/app/router/handlers/UserProfile.js
```diff
@@ -18,11 +18,11 @@ export default class UserProfilerHandler extends BaseHandler {
       fetchUserBasedData(dispatch),
     ]);
 
-    logClientScreenView(buildScreenViewData, getState());
+    trackPageEvents(getState(), buildAdditionalEventData);
```
phil303: this is no longer accepting a callback

src/app/router/handlers/PostsFromSubreddit.js
```diff
@@ -64,7 +64,7 @@ export default class PostsFromSubreddit extends BaseHandler {
     dispatch(setStatus(getState().postsLists[postsListId].responseCode));
     dispatch(setTitle(this.buildTitle(getState(), subredditName)));
 
-    logClientScreenView(buildScreenViewData, getState());
+    trackPageEvents(getState(), buildAdditionalEventData(state));
```
phil303: same comment as the one above

phil303:
:nail_care:

src/app/router/handlers/UserProfile.js
telaviv: it's so weird because the code didn't break at all. Still confused about that one.

telaviv:
💇

src/lib/eventUtils.js
```diff
@@ -82,18 +108,21 @@ export function logClientScreenView(buildScreenViewData, state) {
...
-    const data = buildScreenViewData(state);
-    if (data) {
-      getEventTracker().track('screenview_events', 'cs.screenview_mweb', data);
-    }
+export function trackPageEvents(state, additionalEventData={}) {
+  if (isDuplicatePageView(state)) {
+    return;
+  }
 
+  if (process.env.ENV == 'client') {
```
prashtx: Looks like you're hitting a linter error with the `==` here.

prashtx:
Looks good other than the small linter error

telaviv:
💇 the curse of programming in two languages!

phil303:
looks good from my vantage point :fish:

prashtx:
:fish:

PR #833, opened 2016-10-19
Add EUCookieNotice support

schwers:
This adds support for the EU Cookie Notice to 2X. It does _not_ add the general InfoBar component from 1X (which renders the EUCookie Notice in 1x) that supports app-global messages -- that seemed to be out of scope.

The biggest change was how I refactored the rendering of the AppMain and AppTopNav. Essentially AppMain now controls when the TopNav is rendered. I needed to do this so there was a consistent way to place the EUCookieNotice component above content, but after the TopNav, this is kind of of tricky because TopNav is position fixed. I didn't touch AppOverlayMenu's content, but I did move it because of a stacking issue with the SearchBarOverlay (I didn't look too deep here in interest of getting this in 2X quickly). 

General behavior:

We rely on Fastly sending a country code via headers. If that country is in the EU and we haven't shown the EUCookieNotice a certain number of times, we show it. The component is responsible for dispatching an action when it's shown on a new page. After we've shown it enough, or when the user clicks close, we stop showing it. 

Now, pictures:

Daymode:
image

Nightmode:
image

👓  @uzi || @nramadas  || @phil303  || @ajacksified


src/app/pages/AppMain.jsx
```diff
@@ -43,54 +48,71 @@ const AppMain = props => {
...
-        <Page url='/user/:userName/' component={ UserProfilePage } />
         <Page url='/login' component={ Login } />
         <Page url='/register' component={ Register } />
-        <Page url='/message/compose' component={ DirectMessage } />
-        <Page url='/message/:mailType' component={ Messages } />
         <Page url='/message/messages/:threadId' component={ MessageThread } />
+        <Page url='/r/:subredditName/submit' component={ PostSubmitModal } />
+        <Page url='/submit' component={ PostSubmitModal } />
+        <Page url='/submit/to_community' component={ PostSubmitCommunityModal } />
+        <Case url='*' exec={ () => {
```
phil303: So this is why i was hoping to talk with you about this before you made the patch. In the current infrastructure, I'm not sure this is the right way to approach this. Can we chat about it tomorrow?

src/app/reducers/euCookieNotice.js
```diff
@@ -0,0 +1,61 @@
...
+          numberOfTimesShown,
+        };
+      }
+
+      return state;
+    }
+
+    case euCookieActions.SET: {
+      const { showEUCookie, numberOfTimesShown } = action;
+      return { showEUCookie, numberOfTimesShown };
```
phil303: :no_bell: could just do `return { ...action };` but I'm guessing you didn't for explicitness

src/app/components/EUCookieNotice/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+          <div className='EUCookieNotice__close' onClick={ onClose }>
+            <span className='icon icon-x' />
+          </div>
+          <p>
+            Cookies help us deliver our Services.
+            By using our Services, you agree to our use of cookies.
+            <a
+              className='EUCookieNotice__link'
+              target='_blank'
+              href='https://www.reddit.com/help/privacypolicy'
```
phil303: Can you make this a constant?

schwers: Sure, I'll use the one from config

src/app/side-effect-components/CookieSync.js
```diff
@@ -5,10 +5,12 @@ export default makeCookieArchiver(
...
   state => state.preferences.over18,
   state => state.recentSubreddits,
-  (theme, compact, over18, recentSubreddits) => ({
+  state => state.euCookieNotice.numberOfTimesShown,
+  (theme, compact, over18, recentSubreddits, EUCookieNotice) => ({
     theme,
     compact,
     over18,
     recentSubreddits,
+    EUCookieNotice,
```
phil303: nit maybe just use lower-case `eu`? You do that in the reducer's index file and it doesn't look too weird

phil303:
in general looks pretty solid. let's just talk about the AppMain changes. To be honest, I'm not sure the way i'm thinking about it is the right way to do it either but I'd like to vet that out with you. Also tests!
:nail_care:

src/app/side-effect-components/CookieSync.js
schwers: This is one is more subtle (open to leaving a comment about it) -- the casing here determines the caseing of the cookie, which has already been set by 1X

phil303: Oh duh. :no_bell: - what if you did something like:
```js
  (theme, compact, over18, recentSubreddits, euCookieNotice) => ({
    theme,
    compact,
    over18,
    recentSubreddits,
    EUCookieNotice: euCookieNotice,
  })
```

schwers:
:haircut_man:

phil303:
just one comment about leaving a comment. otherwise :tropical_fish:

PR #834, opened 2016-10-19
Make dropdowns close on IOS when clicking outside the menu

uzi:
Thanks Apple. This is silly, but described by links like
  http://stackoverflow.com/questions/3705937/
in which setting `cursor: pointer;`to the DropdownCover
tricks mobile safari into handling the click event...
you know, because it's not a touch event.

👓 @umbrae @phil303 @schwers

### DIFF
--- a/src/app/components/DropdownCover/styles.less
+++ b/src/app/components/DropdownCover/styles.less
@@ -7,6 +7,7 @@
   bottom: 0;
   left: 0;
   right: 0;
+  cursor: pointer;
   z-index: @z-index-overlay;
 
   .themeify({


src/app/components/DropdownCover/styles.less
```diff
@@ -7,6 +7,7 @@
   bottom: 0;
   left: 0;
   right: 0;
+  cursor: pointer;
```
phil303: post-39692-little-mermaid-sebastian-jaw-d-0ov8

phil303:
i don't understand how this isn't a problem in more places?

:fish: though!

schwers:
:fish:

schwers:
@uzi can you add a note to your commit message about <TooltipShutter /> using native event listeners? I'd just like it to be clear to other devs in the future why this isn't handled by React's event normalization

PR #835, opened 2016-10-19
Add search events

phil303:
The search event types are:
- `cs.search_executed`
- `cs.search_cancelled`
- `cs.search_opened`

Caveats:
In order to track search bar closing accurately, I needed a specific
"close search" action to fire. What was available to me was a more
generic action for closing the overlays. To get around this, I pass in
the "close search" handler from the SearchBarOverlay into the generic
OverlayMenu component. If the close handler from the outside props
exists then it uses that else it uses the generic one.

:eyeglasses: @nramadas || @uzi 
:eyeglasses: @drunken-economist

### DIFF
--- a/src/app/actions/search.js
+++ b/src/app/actions/search.js
@@ -1,5 +1,8 @@
-import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import { endpoints, errors } from '@r/api-client';
+
+import { getEventTracker } from 'lib/eventTracker';
+import { getBasePayload, buildSubredditData } from 'lib/eventUtils';
+import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import { paramsToSearchRequestId } from 'app/models/SearchRequest';
 
 const { SearchEndpoint } = endpoints;
@@ -38,6 +41,19 @@ export const search = searchParams => async (dispatch, getState) => {
   try {
     const apiResponse = await SearchEndpoint.get(apiOptionsFromState(state), searchParams);
     dispatch(received(id, apiResponse));
+
+    const latestState = getState();
+    if (latestState.meta.env === 'CLIENT') {
+      getEventTracker().track('search_events', 'cs.search_executed', {
+        ...getBasePayload(latestState),
+        ...buildSubredditData(latestState),
+        query_string_length: searchParams.q.length,
+        interana_excluded: {
+          query_string: searchParams.q,
+        },
+      });
+    }
+
   } catch (e) {
     if (e instanceof ResponseError) {
       dispatch(failed(id, e));
--- a/src/app/components/OverlayMenu/index.jsx
+++ b/src/app/components/OverlayMenu/index.jsx
@@ -12,23 +12,29 @@ const stopClickPropagation = (e) => {
   e.stopPropagation();
 };
 
-export const OverlayMenu = props => (
-  <nav
-    className={ cx('OverlayMenu', { 'm-with-top-nav': !props.fullscreen }) }
-    onClick={ props.closeOverlayMenu }
-  >
-    <ul className='OverlayMenu-ul list-unstyled' onClick={ stopClickPropagation }>
-      { props.children }
-    </ul>
-  </nav>
-);
+export const OverlayMenu = props => {
+  return (
+    <nav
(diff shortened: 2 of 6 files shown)


src/app/components/SearchBarOverlay/SearchBar/index.js
```diff
@@ -47,8 +47,9 @@ export default class SearchBar extends React.Component {
         className='SearchBar'
         action={ '/search' }
       >
-        <input type='hidden' name='subreddit' value={ subreddit } />
+        <input key='search-hidden' type='hidden' name='subreddit' value={ subreddit } />
```
phil303: Was getting errors because these `Form` children didn't have unique keys.

src/app/components/OverlayMenu/index.jsx
```diff
@@ -12,23 +12,36 @@ const stopClickPropagation = (e) => {
...
-    className={ cx('OverlayMenu', { 'm-with-top-nav': !props.fullscreen }) }
-    onClick={ props.closeOverlayMenu }
-  >
-    <ul className='OverlayMenu-ul list-unstyled' onClick={ stopClickPropagation }>
-      { props.children }
-    </ul>
-  </nav>
-);
+export const OverlayMenu = props => {
+  return (
```
phil303: just moved this into an explicit return statement instead of the implicit multiline one

phil303:
:eyeglasses: @drunken-economist

src/app/components/OverlayMenu/index.jsx
schwers: 🔕  Just a follow up question, what made you want to do so?

src/app/actions/search.js
```diff
@@ -38,6 +41,16 @@ export const search = searchParams => async (dispatch, getState) => {
   try {
     const apiResponse = await SearchEndpoint.get(apiOptionsFromState(state), searchParams);
     dispatch(received(id, apiResponse));
+
+    getEventTracker().track('search_events', 'cs.search_executed', {
```
schwers: Leaving a reminder, please make sure this doesn't unintentionally fire on the server.

nramadas:
LGTM :fish: :tropical_fish: :hairy:

src/app/actions/search.js
```diff
@@ -38,6 +41,19 @@ export const search = searchParams => async (dispatch, getState) => {
...
     dispatch(received(id, apiResponse));
+
+    const latestState = getState();
+    if (latestState.meta.env === 'CLIENT') {
+      getEventTracker().track('search_events', 'cs.search_executed', {
+        ...getBasePayload(latestState),
+        ...buildSubredditData(latestState),
+        interana_excluded: {
+          query_string: searchParams.q,
+          query_string_length: searchParams.q.length,
```
drunken-economist: the `query_string_length` should be a direct child of the payload. Since we don't need to worry about its cardinality, it doesn't need to be excluded from interana

phil303: gotcha thanks justin

phil303:
:haircut:

drunken-economist:
🐟  thanks

src/app/actions/search.js
```diff
@@ -38,6 +41,19 @@ export const search = searchParams => async (dispatch, getState) => {
   try {
     const apiResponse = await SearchEndpoint.get(apiOptionsFromState(state), searchParams);
     dispatch(received(id, apiResponse));
+
+    const latestState = getState();
+    if (latestState.meta.env === 'CLIENT') {
```
schwers: 🔕  After this is out, we should think about the canonical source of 'client vs server' state. There's `process.env.ENV` which has the benefit of not being redux dependent, but OTOH relying on redux does seem nice because that's the rest of the app's source of truth.

schwers:
:fish:

uzi:
:fish:

src/app/actions/search.js
phil303: oh interesting point. I hadn't thought of that. TBH, I like the process.env.ENV approach b/c there's no real setup to be dependent on whereas with the redux state, I have slightly less confidence that this is going to be there correctly.

PR #836, opened 2016-10-19
remove superagent-retry

phil303:
Rationale:
This no longer seems to be a dependency for 2X specifically. 

Testing:
I tested this by removing it from package.json, removing node_modules, and re-installing everything just to be extra sure it was gone. Then spot checking around the app and making sure no requests were failing.


phil303:

phil303:
I'll do a follow up to reddit-mobile-deps that removes this just so no one is surprised by their diffs at a later juncture

nramadas:
:fish:

PR #837, opened 2016-10-19
move client check to process.env.ENV

phil303:
I missed this before I deployed the search events patch but for
consistency with the other client checks in the app, move this to
`process.env.ENV`.

:eyeglasses: @schwers


schwers:
:fish: Thanks for following up!

PR #838, opened 2016-10-19
2 x fixups

schwers:
Just some quick fixups:

1) GTM is proxies its event tracking through an iframe via `postMessage`, seeing some errors in the logs where `gtmIFrame.contentWindow` is undefined. I think this is either because the frame times out or the browser isn't letting us access the iframe. The latter might require more work so this is intended to be a stopgap for now so the logs are cleaner

2) The AppMain refactor made ErrorPages _not_ have a TopNav, second commit adds it back in.

👓  @uzi @phil303

### DIFF
--- a/src/app/pages/AppMain.jsx
+++ b/src/app/pages/AppMain.jsx
@@ -42,8 +42,16 @@ const AppMain = props => {
   } = props;
 
   if (statusCode !== 200) {
+    // NOTE: this manually renders TopNav, see below for an explanation
+    // of how TopNav rendering is working in general. It's manually rendered here
+    // instead of in `<ErrorPage />` as it seems easier to refactor out later
     return (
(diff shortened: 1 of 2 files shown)


src/app/pages/AppMain.jsx
```diff
@@ -42,8 +42,16 @@ const AppMain = props => {
   } = props;
 
   if (statusCode !== 200) {
+    // NOTE: this has manually renders TopNav, see below for an explanation
```
uzi: "this has"

schwers: :(

phil303:
@schwers (1) would be a good description for your commit message

phil303:
:fish: otherwise

uzi:
:fish:

PR #839, opened 2016-10-20
Add VCL rules for whitelisting valid 2X endpoints.

wting:
For all non-whitelisted 2X endpoints, we want to fallback to the desktop site.

:eyeglasses: @phil303 @schwers @uzi 

A few caveats:

- I can't test logged in behavior because login proxy isn't working for me
- We don't fall back to desktop for unsupported, deeper subreddit and user paths (e.g. `/u/wting/submitted`)
- This is a dev-only local change.

### DIFF
--- a/default.vcl
+++ b/default.vcl
@@ -7,20 +7,69 @@ backend mweb {
     .port = "80";
 } */
 
+
 sub vcl_recv {
   call device_detect;
 
-  # NOTE(wting|2016-09-26): Normally we would redirect the user to different
-  # backends, but instead we're just returning rarely used status codes for
-  # easy programmatic testing.
-  if (req.http.Cookie !~ "(^|;\s*)(mweb-no-redirect=1)(;|$)"
-      && (req.http.X-UA-Device ~ "^mobile-" || req.http.X-UA-Device ~ "^tablet-")) {
+  # Check mweb override cookie.
+  if (req.http.Cookie ~ "(^|;\s*)(mweb-no-redirect=1)(;|$)") {
+    call set_www_backend;
+  } elsif (
+    (req.http.X-UA-Device ~ "^mobile-" || req.http.X-UA-Device ~ "^tablet-")
+    && (
+      # Blacklisted endpoints
+      # This endpoint 200's for mweb, but the current implementation has issues
+      req.url !~ "^/message/compose/?$"
+    )
+    && (
+      # Whitelisted 2X endpoints from router/index.js
+      req.url == "/"
+      || req.url ~ "^/actions/"
+      || req.url ~ "^/comments/?"
+      || req.url ~ "^/login/?$"
+      || req.url ~ "^/message/"
+      || req.url ~ "^/r/"
+      || req.url ~ "^/register/?$"
+      || req.url ~ "^/report/?$"
+      || req.url ~ "^/search/?$"
+      || req.url ~ "^/submit/?$"
+      || req.url ~ "^/submit_to_community/?$"
+      || req.url ~ "^/u/"
+      || req.url ~ "^/vote/"
+      || req.url ~ "^/(help|w|wiki)/?"
+      # Whitelisted 2X Ajax endpoints
+      || req.url ~ "^/csp-report$"
+      || req.url == "/error"
+      || req.url ~ "^/(login|refresh|register)proxy$"
+      || req.url ~ "^/logout"
+      || req.url == "/routes"
+      || req.url ~ "^/timings$"
+      || req.url ~ "^/(u/)?XXX\?.*"
+      # Whitelisted dev endpoints (these will be hosted on www in production)
+      || req.url == "/apple-app-site-association"
+      || req.url ~ "^/favicon/"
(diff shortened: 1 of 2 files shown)


default.vcl
```diff
@@ -7,22 +7,64 @@ backend mweb {
...
+      || req.url ~ "^/vote/"
+      || req.url ~ "^/(help|w|wiki)/?"
+      # Framework endpoints
+      || req.url == "/apple-app-site-association"
+      || req.url ~ "^/csp-report$"
+      || req.url == "/error"
+      || req.url ~ "^/favicon/"
+      || req.url ~ "^/favicon\.ico"
+      || req.url ~ "^/fonts/"
+      || req.url == "/health"
```
uzi: Why no '^' on this one or /error?

default.vcl
```diff
@@ -7,22 +7,64 @@ backend mweb {
...
+      || req.url ~ "^/img/"
+      || req.url ~ "^/(login|refresh|register)proxy$"
+      || req.url ~ "^/logout"
+      || req.url ~ "^/ProductionClient\..*\.(css|js)$"
+      || req.url == "/robots.txt"
+      || req.url == "/routes"
+      || req.url ~ "^/timings$"
+      || req.url ~ "^/(u/)?XXX\?.*"
```
uzi: What's this one?

default.vcl
uzi: Or on some others like /routes and /robots.txt for instance

uzi:
💅

default.vcl
wting: These are exact string matches instead of regex (`==` vs `~`), which is generally more performant and easier to understand.

default.vcl
```diff
@@ -7,22 +7,64 @@ backend mweb {
...
+  # Check mweb override cookie.
+  if (req.http.Cookie ~ "(^|;\s*)(mweb-no-redirect=1)(;|$)") {
+    call set_www_backend;
+  } elsif (
+    (req.http.X-UA-Device ~ "^mobile-" || req.http.X-UA-Device ~ "^tablet-")
+    && (
+      # Normal endpoints
+      req.url == "/"
+      || req.url ~ "^/actions/"
+      || req.url ~ "^/comments/?$"
```
wting: Thanks for catching that.

default.vcl
wting: I don't know, but it's called from `ProductionClient.js` and if I don't whitelisted it causes console errors:

xxx js

wting:
:haircut_woman:

default.vcl
```diff
@@ -7,22 +7,65 @@ backend mweb {
...
+      || req.url ~ "^/r/"
+      || req.url ~ "^/register/?$"
+      || req.url ~ "^/report/?$"
+      || req.url ~ "^/search/?$"
+      || req.url ~ "^/submit/?$"
+      || req.url ~ "^/submit_to_community/?$"
+      || req.url ~ "^/u/"
+      || req.url ~ "^/vote/"
+      || req.url ~ "^/(help|w|wiki)/?"
+      # Whitelisted 2X endpoints (React, etc).
```
phil303: nit, i'm not sure what you mean by "React" here. Do you mean endpoints hit via ajax calls?

wting:
Yup. I'm not sure if it's the best name but am open to suggestions.

phil303:
Assuming you're replying to the "React" comment, it's a little weird how you're making the distinction here. Basically your first group is composed of urls that are GETs and your second group are posts, puts, etc. That seems to work out in this app b/c of how it was architected but it feels a little like a coincidence since you can make a call to that first group of endpoints with a post/put/etc.

If you must make the distinction, maybe just call the second group "Ajax endpoints".

wting:
:haircut_woman:

phil303:
:fish:

wting:
:haircut_woman:

uzi:
:fish:

default.vcl
```diff
@@ -7,22 +7,70 @@ backend mweb {
...
+      req.url !~ "^/message/compose/?$"
+    )
+    && (
+      # Whitelisted 2X endpoints from router/index.js
+      req.url == "/"
+      || req.url ~ "^/actions/"
+      || req.url ~ "^/comments/?"
+      || req.url ~ "^/login/?$"
+      || req.url ~ "^/message/"
+      || req.url ~ "^/r/"
```
schwers: Just small note, we don't support comments only listings at the moment. So this and the comments rule on line 28 look like they could match urls like `/comments` and `r/overwatch/comments` that we don't have handlers or ui for

wting: Yeah, there's actually quite a few paths that are unsupported.

We went with the idea that instead of presenting mobile users with a combination of desktop and mweb experiences, they would just see 404's. Monitoring 404s would allow us to prioritize features for 2X implementation.

wting: FYI this is ticketed as CHAN-247.

wting:
:fish: (to make Harold happy)

PR #840, opened 2016-10-20
Load GTM only after the rest of the app has loaded

schwers:
This is a quick disabling of GTM to see the impact on RUM timings after today's increase

👓  @phil303


src/lib/eventUtils.js
```diff
@@ -92,8 +92,6 @@ export function logClientScreenView(buildScreenViewData, state) {
     if (data) {
       getEventTracker().track('screenview_events', 'cs.screenview_mweb', data);
     }
-
-    gtmPageView(state);
```
uzi: eslint is mad about this being defined and not used.

schwers:
[UPDATE]:

I dug into the rum timings a bit and made a bigger change in this patch, this new version loads the GTM asynchronously. New commit message:

GTM's release triggered an increase of over a second in our RUM Timings.
Note that even though iframes load asynchronously, they still block
window.onload. Rendering the GTM script and iframe caused mount timing
to go from ~0.08s to ~0.54s. This patch moves creation of the GTM
iframe to after the page has loaded, and queues pending GTM tracking
until the iframe is properly loaded.

:haircut_man: @phil303 

👓 @uzi @dwick

src/app/components/AdsTracking/index.jsx
```diff
@@ -1,39 +1,15 @@
...
 // This component renders things that are needed for ads tracking.
 // Right now this is:
-// * GoogleTagManager -- used for tracking screenviews for ad targetting
 // * AdblockTester -- used to determine if adblock is enabled, this info is
 //    included on the base events payload
-export const AdsTracking = props => {
-  const { subredditName } = props;
-  const { mediaDomain, googleTagManagerId } = config;
-
+export default function AdsTracking() {
```
phil303: this wrapping component now seems unnecessary. I think it'd be fine to load `AdblockTester` directly.

phil303:
this patch is:
6d9cd43e00b6f4b573f635fe19f392ef0ba33ff8ed5bfbaa73f5f77efe7a39e3
:fish:

PR #841, opened 2016-10-20
Limit timings capturing down to 10% of users

uzi:
This changes both the RUM timings and the handler
timings in with both will individually get a 10%
sample rate.

👓 @schwers @umbrae


umbrae:
🐟

schwers:
:fish:

PR #845, opened 2016-10-21
Remove redirect after POST comment

ajacksified:
👓 @schwers or @nramadas


schwers:
:fish:

schwers:
:fish: to fix the bug, if possible we should add some tests for it

ajacksified:
@schwers added tests

schwers:
:fish: :tropical-fish:

PR #847, opened 2016-10-24
Replying overhaul

phil303:
This is an overhaul to replying incorporating our best practices. Which
means:

1. Real ajax requests using JSForm / removal of the reply POST handler.
2. A success and failure toaster. (thumbs up from Erica on this)
3. Simplified reducer code for handling reply state.

### Details:
This started off with a simple removal of the redirect that was
happening in the POST handler but I took it as an opportunity to clean
some of the older code (when we were still forming ideas on how this app
should work.)

There were two reducers addressing reply state. One seemed to handle the
view and one seemed to handle fetching. Nothing seemed to be taking
advantage of the latter so I just nixed it entirely. You can see
replying works fine without it. The `replying` reducer now only handles
whether or not a user is currently replying to a comment.

As mentioned above, I've also added a success toaster which triggers if
a reply was successful. The pre-existing failure toaster triggers if
not. One thing to note is that this is not optimistic. This is due to
needing the canonical reply model from the api to update the state. I
think this is fine as from a user's perspective, it's not completely
unexpected behavior.

### Caveats:
- This is dependent on this [node-api-client
  patch](https://github.com/reddit/node-api-client/pull/171). Without
  it, a user is unable to reply to a reply unless he or she refreshes.

### What this looks like:
reply overhaul

:eyeglasses: @schwers


src/app/actions/reply.js
```diff
@@ -8,32 +8,40 @@ import { apiOptionsFromState } from 'lib/apiOptionsFromState';
...
+export const FAILURE = 'REPLY__FAILURE';
 
-export const REPLYING = 'REPLYING';
-export const replying = (id, text) => ({ type: REPLYING, id, text });
+export const toggle = id => ({ type: TOGGLE, id });
+export const success = (id, reply) => ({
+  id,
+  type: SUCCESS,
+  model: reply,
+  message: 'Comment received!',
```
phil303: this is mostly a placeholder message

src/app/actions/reply.js
```diff
@@ -8,32 +8,37 @@ import { apiOptionsFromState } from 'lib/apiOptionsFromState';
...
+  type: SUCCESS,
+  model: reply,
+  message: 'Comment added!',
+});
 
-export const REPLIED = 'REPLIED';
-export const replied = (id, model) => ({ type: REPLIED, id, model });
-
-export const reply = (id, text) => async (dispatch, getState) => {
+export const submit = (id, { text }) => async (dispatch, getState) => {
```
schwers: `id` designates the parent object we're submitting a reply to right? It might be a little clearer if there was a comment or if the variable was named `parentId`

src/app/components/Comment/CommentReplyForm/index.jsx
```diff
@@ -1,31 +1,18 @@
...
-      className='CommentReplyForm__close icon icon-x Button m-linkbutton'
-      onClick={ onToggleReply }
-    />
-  );
-};
+const T = React.PropTypes;
 
 export class CommentReplyForm extends React.Component {
   static propTypes = {
-    text: T.string,
```
phil303: ah there's more cleanup to be done here. good find. currently, text is not being used as a prop anywhere in the app so really this is just internal state with a default value. i'll update accordingly

src/app/reducers/replying.js
```diff
@@ -1,21 +1,26 @@
...
+      return DEFAULT;
     }
 
-    case replyActions.REPLIED: {
-      const newState = { ...state };
-      delete newState[action.id];
-      return newState;
+    case replyActions.SUCCESS:
+    case replyActions.TOGGLE: {
+      if (state[action.id]) {
```
schwers: You could simplify this whole case to:

```js
return merge(state, { [action.id]: !state[action.id] });
```

src/app/components/Toaster/styles.less
```diff
@@ -28,21 +28,30 @@
       background-color: @ban-red;
     }
 
+    &.success {
+      background-color: @lime;
```
schwers: :excellent:

src/app/router/handlers/CommentsPage.js
```diff
@@ -100,25 +98,6 @@ export default class CommentsPage extends BaseHandler {
 
     logClientScreenView(buildScreenViewData, getState());
   }
-
-  async [METHODS.POST](dispatch, getState, { waitForState }) {
```
schwers: So much yes

schwers:
💅 A few little things but structurally everything looks good, nice job

phil303:
:haircut:

src/app/components/Comment/CommentReplyForm/index.jsx
```diff
@@ -19,16 +19,12 @@ export class CommentReplyForm extends React.Component {
...
+      disableButton: true,
     };
   }
 
   onTextChange = (e) => {
-    this.setState({
-      text: e.target.value,
-      disableButton: !e.target.value,
-    });
+    this.setState({ disableButton: !e.target.value });
```
schwers: tiny-nit: I think you want to call `.trim()` here to keep the button disabled when the textarea only has whitespace

schwers:
:fish:

PR #849, opened 2016-10-24
Update expiry on desktop redirect cookie to 1 year

phil303:
This was an oversight on the desktop redirect patch. The expiry is
currently just a session cookie. It needs to stick around for 1 year
for parity with the 1X implementation.

I also added a linter fix in `replyingComments.test`.

:eyeglasses: @uzi || @nramadas


uzi:
:fish:

PR #851, opened 2016-10-25
Add the correct path to desktop redirect

phil303:
Currently we're just redirecting the frontpage of the desktop site
instead of redirecting to the desktop version of the page the user is
on. This adds the path to the redirect to fix this.

:eyeglasses: @uzi


uzi:
:fish: Cool... I double-checked and url.parse().path preserves the query parameters as well

phil303:
Thanks for your thoroughness @uzi!

PR #852, opened 2016-10-25
Remove unused ReportHandler and its endpoint

phil303:
This was a placeholder component from the no-js days. It's going unused
now so remove it from the codebase.

:eyeglasses: @nramadas


nramadas:
:fish:

PR #854, opened 2016-10-26
Toaster - Prevent setState after unmount

phil303:
Bug:
When the toaster is dismissed via the close button the timer that runs
the toaster component's setState doesn't get cleared, which means it
will run after the toaster has unmounted.

Fix:
Using an instance variable to hold the timer reference. If the component
is about to unmount, clear the reference.

Testing:
Tested both letting the timeout animation run and hitting the dismiss
button manually and both are good.

:eyeglasses: @schwers


schwers:
:fish:

PR #855, opened 2016-10-26
Update page titles to match desktop's

phil303:
Right now, mweb titles are different than desktop's. For instance, on
the frontpage, currently we have "Reddit - the Front Page of the
Internet", but desktop has "reddit - the front page of the internet".
This updates that title accordingly.

On comment's pages, I removed the "reddit" prefix entirely and moved
the post title to the first position to differentiate tabs better. This
also matches desktop's format.

:eyeglasses: @schwers 
cc @prashtx just in case


schwers:
:fish:

schwers:
Might be good to triple check with @prashtx

prashtx:
:fish:

PR #856, opened 2016-10-27
Whitelist /user/ for 2X.

wting:
2X currently redirects /u/wting to /user/wting. Since that path isn't
whitelisted it results in showing the desktop version instead.

:eyeglasses: @schwers @prashtx

### DIFF
--- a/default.vcl
+++ b/default.vcl
@@ -35,6 +35,7 @@ sub vcl_recv {
       || req.url ~ "^/submit/?$"
       || req.url ~ "^/submit_to_community/?$"
       || req.url ~ "^/u/"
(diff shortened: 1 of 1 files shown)


uzi:
:fish:

prashtx:
:fish:

schwers:
Do we care about activity urls? It looks like mweb and desktop have slightly different urls for user activity listings. compare desktop's /user/:username/comments to mweb's /user/:username/activity?activity=comments.

Other than that, :fish:

wting:
@schwers: Good point, ticketed as CHAN-255.

PR #857, opened 2016-10-27
Fix loidcreated getting reset

nramadas:
Two issues here:

0. We were comparing the `ms` value of the loidcreated we got back from the api to the ISO string version we had stored in the cookie. They obviously would not match, and so we were continuously re-writing the cookie. This in itself isn't that bad if we were constantly re-writing the same value. But we weren't. Because....
0. When sending the `loidcreated` cookie in the request, there was an extra `}` character in there. Which meant that the `loidcreated` we were sending the api was always invalid, so it was getting re-issued. FML.

### DIFF
--- a/src/lib/apiOptionsFromState.js
+++ b/src/lib/apiOptionsFromState.js
@@ -25,7 +25,7 @@ export const apiOptionsFromState = state => {
     }
 
     if (loidCreated) {
-      cookieHeaders.push(`loidcreated=${loidCreated}}`);
+      cookieHeaders.push(`loidcreated=${loidCreated}`);
     }
 
     return merge(options, {
--- a/src/server/initialState/dispatchInitialUser.js
+++ b/src/server/initialState/dispatchInitialUser.js
@@ -25,7 +25,11 @@ export const dispatchInitialUser = async (ctx, dispatch, getState) => {
     ctx.cookies.set('loid', loid, options);
   }
 
-  if (loidCreated && (loidCreated !== oldLoidCreated)) {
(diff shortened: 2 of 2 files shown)


src/lib/apiOptionsFromState.js
```diff
@@ -25,7 +25,7 @@ export const apiOptionsFromState = state => {
     }
 
     if (loidCreated) {
-      cookieHeaders.push(`loidcreated=${loidCreated}}`);
+      cookieHeaders.push(`loidcreated=${loidCreated}`);
```
schwers: https://github.com/reddit/reddit-mobile/commit/a358ce37bd12eaafb4b7a1b79533ef3c4ce9c91e FML indeed

src/server/initialState/dispatchInitialUser.js
```diff
@@ -25,7 +25,11 @@ export const dispatchInitialUser = async (ctx, dispatch, getState) => {
     ctx.cookies.set('loid', loid, options);
   }
 
-  if (loidCreated && (loidCreated !== oldLoidCreated)) {
+  // oldLoidCreated is always an ISO string, since that's how it's stored in the
```
schwers: In case anyone else reading a long is wondering, 1X used to set `loidcreated` as an ISO string as well. So in the off-chance someone comes back who hasn't been to the site in the past say, 2 months+, this logic still holds https://github.com/reddit/reddit-mobile/blob/77dad5c963fd594955c216265882b115a6119f76/src/lib/loid.es6.js#L5

schwers:
:fish:

PR #858, opened 2016-10-27
Speed up voting

phil303:
Problem:
Voting is noticeably slow sometimes and very bad on older phones. This
is due to some sloppy selectors in our CommentsPage code - namely the
features selector and the comments selector, but in general most of the
selectors weren't being selective enough.

The features selector was listening for ANY change in state. A simple
upvote fits this criteria and therefore the whole page was re-rendering.
Listening for comments has a similar problem since a new comment, which
is created when we vote, means the whole tree changes.

Fix:
Remove these two selectors and make the rest of the selectors more
specific. This clean up should have the effect of speeding up any
actions done on the comments pages.

Additional changes
I've also cleaned up the voting component since its using the no-js
way of doing things was also a source of slowness. I've simplified it
down to the bare essentials (the amount of code to get this to work
before was ridiculous) and per Erica's recommendation, posts and
comments have the same CSS.

How it looks in action (see, still works 😄 with way less code):
voting overhaul

:eyeglasses: @schwers @scarow 
cc @umbrae


src/app/components/Post/PostFooter/styles.less
```diff
@@ -1,7 +1,8 @@
 @import (reference) '~app/less/variables';
 @import (reference) '~app/less/mixins/clearfix';
 @import (reference) '~app/less/themes/themeify';
-@import (reference) '~app/less/animations/voteBounce';
+
+@min-hit-area: 3 * @grid-size;
```
schwers: This is already defined in `variables.less`. 🔕  As a side note, I think we should turn it's definition from `3 * @grid-size` to `24px`. It's somewhat orthogonal from the size of the grid

src/app/components/VotingBox/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+import React from 'react';
+import { connect } from 'react-redux';
+
+import cx from 'lib/classNames';
+import * as votingActions from 'app/actions/vote';
+
+const T = React.PropTypes;
+
+
+function VotingBox(props) {
```
schwers: 🔕 No love for `const` ?

src/app/components/VotingBox/styles.less
```diff
@@ -0,0 +1,61 @@
...
+    .themeify({
+      color: @theme-nav-icon-color;
+    });
+  }
+
+  &__upvote,
+  &__downvote {
+    &.icon {
+      font-size: 24px;
+      line-height: 24px;
```
schwers: So in theory we want the icons to be at least `@min-hit-area` tall and wide, could we use that var instead of `24px` and `3 * @grid-size`? That has the added benefit of staying constant (i.e. right now 24px just happens to equal 3 * @grid-size -- I don't expect it to change but it seems bad that a change could affect this)

phil303: ah i didn't know that. I'll put that back in there but I agree that we don't really seem to use it (removing it was very simple)

src/app/components/VotingBox/index.jsx
phil303: I'm just not a fan of *always* using it. If I can use `function`, especially for longer functions, I'd rather use that. In a world of `const`s, to me it's a way of distinguishing my code a bit more.

src/app/pages/Comments.jsx
```diff
@@ -17,340 +15,151 @@ import GoogleCarouselMetadata from 'app/components/GoogleCarouselMetadata';
...
-    isCrawlerRequest
-  ) => {
-    const commentsPageParams = CommentsPageHandler.pageParamsToCommentsPageParams(pageProps);
-    const commentsPageId = paramsToCommentsPageId(commentsPageParams);
-    const commentsPage = commentsPages[commentsPageId];
-    const topLevelComments = (!commentsPage || commentsPage.loading)
-      ? []
-      : commentsPage.results;
-    const post = posts[commentsPageParams.id];
+  (commentsPage, post, isReplying, topLevelComments, currentPage, preferences, isCrawlerRequest) => {
```
schwers: It looks like `topLevelComments` is unused and it's selector is a duplicate of `commentsPage`. Seems safe to remove?

src/app/router/index.js
```diff
@@ -50,7 +49,6 @@ export default [
   ['/submit/to_community', PostSubmitCommunityHandler],
 
   // actions
-  ['/vote/:thingId', Vote],
```
schwers: 🔕 as it's a little out-of-scope: In this patch or a subsequent we should remove `/vote` from the whitelisted VCL

src/app/pages/Comments.jsx
phil303: nice catch! thank you

src/app/components/VotingBox/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const downvoteClass = cx('VotingBox__downvote icon icon-downvote', {
+    'downvoted m-animated': voteDirection === -1,
+  });
+
+  const scoreClass = cx('VotingBox__score', {
+    'upvoted': voteDirection === 1,
+    'downvoted': voteDirection === -1,
+  });
+
+  const upVote = voteDirection === 1 ? 0 : 1;
```
phil303: @schwers just an fyi, you were right on this. This kind of toggling is handled by [node-api-client](https://github.com/reddit/node-api-client/blob/3X/src/models2/mixins/votable.js#L17). Will revert in tomorrow morning.

phil303: upon closer inspection, this is not handled by node-api-client. notice that the `direction` passed in still determines what endpoint to use.

src/app/router/index.js
phil303: just an fyi, once this goes in, i'll make a PR with [this](https://github.com/reddit/vcl/compare/master...phil303:remove-unused-endpoints)

schwers:
:fish:

schwers:
Nice work, I'll get the gitbook for frontend guild going today or tomorrow, but we should add a write up on how the reselect perf debugging happened -- lot of good general knowledge to learned from this patch.

PR #859, opened 2016-10-27
Fix broken reducer test

phil303:
Bug:
One of the replying tests is currently broken due to my not using the
action creator. When I updated the payload in the action creator, I
forgot to update the payload in the test.

Fix:
Just use the action creator foo.

:eyeglasses: @uzi


uzi:
:fish:

PR #860, opened 2016-10-28
Add Ad Preview Component

jamesmaa:
👓 @dwick  is this a good approach?


dwick:
I think you should supply a json endpoint with https://github.com/reddit/reddit-public/pull/6047 instead of stubbing all the attributes manually. 💅

dwick:
:runner:

PR #861, opened 2016-10-28
Use r2 for all endpoints with extensions.

wting:
:eyeglasses: @prashtx @uzi

### DIFF
--- a/default.vcl
+++ b/default.vcl
@@ -16,6 +16,9 @@ sub vcl_recv {
     call set_www_backend;
   } elsif (
     (req.http.X-UA-Device ~ "^mobile-" || req.http.X-UA-Device ~ "^tablet-")
+    # We don't host endpoints with extensions in mweb.
+    # The Fastly equivalent is: && !req.url.ext
+    && !reg.url ~ "\/.+\.[a-z]{3,4}(\?.*|$)"
     && (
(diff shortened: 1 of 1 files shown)


default.vcl
```diff
@@ -20,6 +20,10 @@ sub vcl_recv {
       # Blacklisted endpoints
       # This endpoint 200's for mweb, but the current implementation has issues
       req.url !~ "^/message/compose/?$"
+      # We don't host json in mweb
+      || req.url !~ ".*\.json(\?.*|$)"
+      # We don't host XML / RSS in mweb
+      || req.url !~ "(?i).*\.(xml|rss)$"
```
prashtx: If `req.url` contains query params in addition to the path, then I think we need the same `(\?.*|$)` handling for xml/rss that you have for json (we support `?limit=50`, for example).

wting:
💇

prashtx:
:fish:

wting:
💇

wting:
Here's a handy [visual aid](https://regexper.com/#%5C%2F.%2B%5C.%5Ba-z%5D%7B3%2C4%7D(%5C%3F.*%7C%24)) for parsing the regex. Keep in mind that link has an off by one error.

prashtx:
:fish:

PR #862, opened 2016-10-29
Session refreshing on visibility changes

phil303:
Bug:
When a user has a reddit tab open and comes back to it some time later,
the session has expired. Any logged in action that user tries kicks them
to the register screen because our js doesn't know to try to update the
session. A user hitting refresh fixes that since the cookies are still
valid.

Fix:
This adds a visibility check that fires every time a tab is focused,
leveraging the [visibility api](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API), which is [well supported](http://caniuse.com/#search=visibilitychange) on mobile.
If the user had a refresh token and the token was expired, the
app will fetch a new one upon being focused again. This is a nice all
around solution since it avoids having to specially handle every logged
in action.

Caveats:
I've cleaned up the session refresh code a bit, which was handling token
refreshing for new and very long sessions. One thing I added was a
minimum to the refresh timer just for fear of ddos'ing ourselves. This
is set at 5 minutes.

Testing:
I tested this on my iPhone by dropping the expiration down to 10
seconds, opening the app in one tab, going to another, then coming back
to the first after 10 seconds. Then hitting upvote. I did a similar test
with the open app and the lock screen. Both worked. I did this in both
Chrome and Safari. This still needs testing on android.

Todo:
- [x] Test on android

:eyeglasses: @schwers || @nramadas

### DIFF
--- a/src/app/side-effect-components/SessionRefresher.jsx
+++ b/src/app/side-effect-components/SessionRefresher.jsx
@@ -2,41 +2,47 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
 
+import onVisibilityChange from 'lib/onVisibilityChange';
 import * as sessionRefreshingActions from 'app/actions/sessionRefreshing';
 
-class _SessionRefresh extends React.Component {
-  constructor(props) {
-    super(props);
+const FIVE_MIN = 5 * 60 * 1000;
 
-    this.timeoutId = 0;
-    this.mounted = false;
-  }
+class SessionRefresh extends React.Component {
+  refreshTimer = null;
 
   componentDidMount() {
-    this.mounted = true;
-    const { session } = this.props;
-    this.awaitSession(session);
+    // on initial render get the most up-to-date refresh token
+    this.refreshSession();
+
+    // when stale tabs come back into focus, attempt refresh as well
+    onVisibilityChange(() => this.refreshSession({ expiredOnly: true }));
   }
 
-  awaitSession(session) {
-    clearTimeout(this.timeoutId);
-    if (session.refreshToken) {
-      const now = new Date();
-      const when = (new Date(session.expires) - now) * 0.9;
-      this.timeoutId = setTimeout(this.props.refreshSession, when);
-    }
+  componentDidUpdate() {
+    // if the component gets a new session, it will update, and the timer needs
+    // to restart
+    this.startRefreshTimer();
   }
 
-  shouldComponentUpdate(nextProps) {
-    if (this.mounted && nextProps.session) {
-      this.awaitSession(nextProps.session);
-    }
+  startRefreshTimer() {
+    clearTimeout(this.refreshTimer);
+    // set up the refresh timer but defend against potentially too low timeouts
+    const when = Math.max((this.props.session.expires - Date.now()) * 0.9, FIVE_MIN);
+    this.refreshTimer = setTimeout(() => this.refreshSession(), when);
+  }
 
-    return false;
+  refreshSession({ expiredOnly=false }={}) {
+    if (this.props.session.refreshToken) {
+      if (expiredOnly && this.props.session.isValid) {
+        return;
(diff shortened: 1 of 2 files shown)


schwers:
Hey @phil303, thanks for doing this, it should be a really nice improvement. I tested this on android by making the following change and so I could confirm the visibility change code was working:

```diff
diff --git a/src/app/side-effect-components/SessionRefresher.jsx b/src/app/side-effect-components/SessionRefresher.jsx
index 0486d04..85fbb84 100644
--- a/src/app/side-effect-components/SessionRefresher.jsx
+++ b/src/app/side-effect-components/SessionRefresher.jsx
@@ -4,6 +4,7 @@ import { createSelector } from 'reselect';

 import onVisibilityChange from 'lib/onVisibilityChange';
 import * as sessionRefreshingActions from 'app/actions/sessionRefreshing';
+import * as reportingActions from 'app/actions/reporting';

 const FIVE_MIN = 5 * 60 * 1000;

@@ -15,7 +16,10 @@ class SessionRefresh extends React.Component {
     this.refreshSession();

     // when stale tabs come back into focus, attempt refresh as well
-    onVisibilityChange(() => this.refreshSession({ expiredOnly: true }));
+    onVisibilityChange(() => {
+      this.props.toastOnFocus();
+      this.refreshSession({ expiredOnly: true });
+    });
   }

   componentDidUpdate() {
@@ -51,6 +55,10 @@ const mapStateToProps = createSelector(

 const mapDispatchToProps = (dispatch) => ({
   refreshSession: () => dispatch(sessionRefreshingActions.refresh()),
+  toastOnFocus: () => dispatch({
+    type: reportingActions.SUCCESS,
+    message: 'focuesed, trying to refresh',
+  }),
 });

 export default connect(mapStateToProps, mapDispatchToProps)(SessionRefresh);
```

seemed to work well

src/app/side-effect-components/SessionRefresher.jsx
```diff
@@ -2,42 +2,48 @@ import React from 'react';
...
   }
 
-  shouldComponentUpdate(nextProps) {
-    if (this.mounted && nextProps.session) {
-      this.awaitSession(nextProps.session);
-    }
+  startRefreshTimer() {
+    clearTimeout(this.refreshTimer);
+    // set up the refresh timer but defend against potentially too low timeouts
+    const when = Math.max((this.props.session.expires - Date.now()) * 0.9, FIVE_MIN);
```
schwers: I'm a little worried about using `Math.max(..., FIVE_MIN)` -- we don't ensure on the backend that we've refreshed if the token would go stale in the next five minutes, so there could be a window of time when you first load that the token is stale. That could lead to some bad UX, i.e. error messages and listings not loading. I think it might be worthwhile to check with infra to see if we really need a guard like this on the client. They might prefer some random back-off as opposed to a fixed length unit (to prevent heard'ing, they'd know better if it's a good idea or not), etc.

schwers:
💅  Just a concern with the `FIVE_MIN` guard, everything else looks great

src/app/side-effect-components/SessionRefresher.jsx
phil303: ah, ok that might be the case. I had a print statement in the code leading up to the actual request but didn't look at the request itself.

schwers:
:fish:

src/app/side-effect-components/SessionRefresher.jsx
phil303: Just to leave for posterity, after investigating this, the backend only sends back 1 hours ttls. Additionally, Ryan was concerned that if refreshing wasn't happening immediately, it might encounter this check and there'd be a 5 minute dead period. The code does bypass this check though and refreshes immediately on mount per [this line](https://github.com/reddit/reddit-mobile/pull/862/files#diff-82c5a00d17ee548d11491534bd2a4ed0R15).

PR #863, opened 2016-10-30
RFC: Remove month abbreviation

phil303:
As a possible solution to the ambiguous "m" problem, this patch removes
the month abbreviation entirely. I chose this path for a few reasons.
1. The "mo" abbreviation looks weird to me (could just be me on this). 
   Ex text would be:
   - `r/foobarsub ● 5mo ● u/thephilthe`
2. The math for calculating the month was an approximation based on 30
   day months, which seems too inaccurate.
3. Finally, native apps don't show months either and in my opinion, it
   looks ok.

General notes:
I'm not entirely happy with this quick solution but I think it's better
than what we have. Looking around at how other apps handle this, most
seem to fallback to the full date text after a day or so, giving more
time info as more time goes by. Taking our example above, depending on
the "time ago" date, the text might look something like:
- `r/foobarsub ● 5 mins ● u/thephilthe`
- `r/foobarsub ● 8 hrs ● u/thephilthe`
- `r/foobarsub ● 1 day ● u/thephilthe`
- `r/foobarsub ● Aug 18 ● u/thephilthe`
- `r/foobarsub ● Dec 20, 2015 ● u/thephilthe`

I find this to be far more informative than our ambiguous "2 years ago"
format. Auditing the places where we show this type of info, I think
this would work everywhere but obviously I'd like to get a designer's
POV on it.

What this patch looks like:
**Comments**
screen shot 2016-10-31 at 10 01 36 am

**Listing**
screen shot 2016-10-31 at 10 01 02 am

:eyeglasses: @schwers

### DIFF
--- a/src/app/reducers/replying.test.js
+++ b/src/app/reducers/replying.test.js
@@ -24,7 +24,7 @@ createTest({ reducers: { replying } }, ({ getStore, expect }) => {
 
     it('should set the reply state to false when submitted successfully', () => {
       const { store } = getStore({ replying: { [ID]: true } });
-      store.dispatch({ id: '1', type: replyActions.SUCCESS });
+      store.dispatch(replyActions.success(ID, {}));
 
       const { replying } = store.getState();
       expect(replying[ID]).to.be.equal(false);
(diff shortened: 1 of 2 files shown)


umbrae:
Strong +1 from a UX perspective.

schwers:
Strong 👍  from me as well. I really like the explicit format you proposed FWIW. :fish:

phil303:
Awesome! I actually have a test branch with it to show to some designers.

PR #864, opened 2016-10-31
Add amphtml link tags for self posts

prashtx:
For self post comments pages, we add a link tag with `rel="amphtml"` to
indicate to crawlers where they can find the AMP version of the page.
Because this is a crawler-targeted feature, we only need to do this for
compelte server renders.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -63,6 +63,8 @@ const config = () => ({
   reddit,
   rootReddit,
 
+  amp: process.env.AMP,
+
   googleTagManagerId: process.env.GOOGLE_TAG_MANAGER_ID,
 
   adblockTestClassName: process.env.ADBLOCK_TEST_CLASSNAME || 'ad adsense-ad googad gemini-ad openx',
--- a/src/lib/createAmpHtmlLinkFromState.js
+++ b/src/lib/createAmpHtmlLinkFromState.js
@@ -1,0 +1,41 @@
+import { matchRoute } from '@r/platform/navigationMiddleware';
+
+import config from 'config';
+
+export const ampLink = (currentPage, state) => {
+  const { postId } = currentPage.urlParams;
+  const post = state.posts[`t3_${postId}`];
+  if (!post || !post.cleanPermalink) {
+    // We were unable to get the info from the server, which probably means
+    // the subreddit was banned or the like.
+    return null;
+  }
+
+  if (!post.isSelf) {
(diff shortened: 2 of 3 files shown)


prashtx:
👓 @phil303 @uzi

src/lib/createAmpHtmlLinkFromState.js
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const AMP_ROUTES = [
+  // comment pages
+  ['/r/:subredditName/comments/:postId/comment/:commentId', ampLink],
+  ['/r/:subredditName/comments/:postId/:postTitle/:commentId', ampLink],
+  ['/r/:subredditName/comments/:postId/:postTitle?', ampLink],
+  ['/comments/:postId/:postTitle/:commentId', ampLink],
+  ['/comments/:postId/:postTitle?', ampLink],
+  ['*', () => null],
```
phil303: is this necessary?

src/config.js
```diff
@@ -63,6 +63,8 @@ const config = () => ({
   reddit,
   rootReddit,
 
+  amp: process.env.AMP,
```
phil303: this is a domain right? What do you think of calling it something more descriptive, like `ampDomain`? (I know this is not consistent with the other domain variables)

phil303:
small small questions/nits. looks good structurally though :nail_care:

src/lib/createAmpHtmlLinkFromState.js
prashtx: Do you mean the `'*'` route? Not strictly. If there's no match, we'll get a `null` value for `handler` down below, so we'll have do some more null testing. This approach lets us stay consistent with the other situations where we provide a `null` path for the AMP'd page. I don't have a strong opinion either way.

phil303: :no_bell: Ah I see now. Moving to no_bell b/c I don't feel strongly either. A slight preference for the null check down below because it feels slightly less abusive.

src/config.js
prashtx: It's a prefix that contains both the scheme and domain. If we want to set a better precedent, how about `ampPrefix`?

prashtx:
:haircut_man:

phil303:
:fish: we chatted about the amp naming but we couldn't come up with a name we liked so I'm cool with leaving it.

PR #865, opened 2016-10-31
Reduce number of actions logged.

wting:
We currently default to 50 action steps logged. Since this exceeds our limits
per log line (4096 bytes), it truncates the log lines resulting in unparseable
JSON. As a result we can't programmatically analyze our error logs.

:eyeglasses: @schwers @uzi


uzi:
:fish:

schwers:
:fish:

PR #866, opened 2016-10-31
Correct types on the error toaster

phil303:
Bug:
This copy was hardcoded to just 'comment' but the action applies to both
comments and posts. When something went wrong on a post upvote we'd show
"failed to upvote the comment", which makes no sense.

Fix:
Use the `type` (which is either 'post' or 'comment') in the copy.

:eyeglasses: @uzi


uzi:
:fish:

PR #869, opened 2016-11-01
Fix message thread replying

uzi:
It turns out that we can only reply to a message
thread if the other side has sent us a message.
For example, if there's one message in the thread
and the current user sent it, they can't send
further messages until the other side sends a reply.

Take a look at all of the messages in a thread, find
the latest response from the other side and use that
as the message id we respond to. Unfortunately the
API appears to succeed if we reply to our own message,
so this circumvents the issue.

We no longer show the replyToggle if the user can't
yet reply to a thread.

👓 @phil303 @schwers @andytuba 

And thanks @andytuba for helping diagnose this.

### DIFF
--- a/src/app/components/MessageThread/index.jsx
+++ b/src/app/components/MessageThread/index.jsx
@@ -12,8 +12,9 @@ const T = React.PropTypes;
 
 export class MessageThread extends React.Component {
   static propTypes = {
+    headMessage: T.object.isRequired,
     messages: T.object.isRequired,
-    threadId: T.string.isRequired,
+    replyMessageId: T.string.isRequired,
   };
 
   constructor(props) {
@@ -41,13 +42,11 @@ export class MessageThread extends React.Component {
 
   render() {
     const {
+      headMessage,
       messages,
-      threadId,
+      replyMessageId,
     } = this.props;
     const { showForm } = this.state;
-
-    const headMessageId = `t4_${threadId}`;
-    const headMessage = messages[headMessageId];
 
     return (
       <div className='MessageThread'>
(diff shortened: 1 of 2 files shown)


src/app/components/MessageThread/index.jsx
```diff
@@ -43,12 +44,32 @@ export class MessageThread extends React.Component {
...
     const headMessage = messages[headMessageId];
 
+    // We can only reply to a message thread if the other side has sent
+    // us a message. For example, if there's one message in the thread and
+    // the current user sent it, they can't send further messages until the
+    // other side sends a reply.
+    //
+    // Get the latest reply from the other side and we reply to that.
+    let replyMessageId;
+    if (!user.loggedOut) {
```
phil303: I don't think this check is necessary since you can't get to this page if you're logged out right?

src/app/components/MessageThread/index.jsx
```diff
@@ -43,12 +44,32 @@ export class MessageThread extends React.Component {
...
+      // Copy and reverse the list of replies and add the head message id as well.
+      const recentMessageIds = headMessage.replies.slice().reverse();
```
phil303: This logic can be made a little simpler and should be down in `mergeProps`.
```js
const mergeProps = (stateProps, dispatchProps, ownProps) => {
  const threadId = ownProps.urlParams.threadId;   // no longer needed as prop now
  const headMessageId = `t4_${threadId}`;
  const headMessage = stateProps.messages[headMessageId];
  
  // your nice comment here
  const recentMessageIds = [ headMessage.replies.slice.reverse(), headMessageId ];
  const replyMessageId = recentMessageIds.find(id => messages[id].author !== user.name);
  
  return {
    ...stateProps,
    ...dispatchProps,
    headMessage,
    replyMessageId,
  };
});
```

The reason we move it into `mergeProps` is because generally speaking you want to keep as much logic out of components as possible. I know some of this was pre-existing but this is a good opportunity for some cleanup. 

You can actually take this one step further and close over the `replyMessageId` with your `onSubmit` handler but I'll leave that up to you.

phil303:
:nail_care:

src/app/components/MessageThread/index.jsx
uzi: You're right... I was being more careful about things, but I can remove it.

src/app/components/MessageThread/index.jsx
uzi: Ahh, good point... I'll move it.

uzi:
💇

phil303:
tested locally. Looks good! Thanks for fixing this. :fish:

PR #873, opened 2016-11-02
Update VCL routes.

wting:
:eyeglasses: @uzi @prashtx


default.vcl
```diff
@@ -45,7 +43,7 @@ sub vcl_recv {
       || req.url ~ "^/(login|refresh|register)proxy$"
       || req.url ~ "^/logout"
       || req.url == "/routes"
-      || req.url ~ "^/timings$"
+      || req.url ~ "^/timings"
```
uzi: Really? I thought it was all done with POST

uzi:
:fish: Just the one question, but good otherwise.

default.vcl
wting: Good point, I actually don't know exactly.

@prashtx thoughts?

PR #877, opened 2016-11-03
RFC - Posts and comments - different date formats

phil303:
For 7 days we show the "time ago" format with increasing increments,
minutes to hours to days, then after that we show the actual date.

How this looks:
- `r/foobarsub ● 5 mins ● u/thephilthe`
- `r/foobarsub ● 8 hrs ● u/thephilthe`
- `r/foobarsub ● 1 day ● u/thephilthe`
- `r/foobarsub ● 7 days ● u/thephilthe`
- `r/foobarsub ● Aug 18 ● u/thephilthe`
- `r/foobarsub ● Dec 20, 2015 ● u/thephilthe`

On posts on listings, this can get a little cramped (it's already pretty
cramped), so I took a page from the iOS app and removed the user. The
user will show up on the comments page version of the Post.

Various examples of what this looks like:

**Posts hours old (notice no user name)**
screen shot 2016-11-03 at 7 55 32 am

**Comments hours old**
screen shot 2016-11-03 at 7 56 39 am

**Comments days old**
screen shot 2016-11-03 at 8 16 13 am

**Mixed format (comments less than and greater than a week old)**
screen shot 2016-11-03 at 8 31 07 am

**Deeply nested comments (previous calendar year - notice the wrapped header)**
screen shot 2016-11-03 at 8 38 21 am

### DIFF
--- a/src/app/components/Comment/CommentHeader/index.jsx
+++ b/src/app/components/Comment/CommentHeader/index.jsx
@@ -3,7 +3,7 @@ import './styles.less';
 import React from 'react';
 import fill from 'lodash/fill';
 
-import { short } from 'lib/formatDifference';
+import formatElapsedTime from 'lib/formatElapsedTime';
 
 const T = React.PropTypes;
 const separator = <div className='CommentHeader__separator'> • </div>;
@@ -79,7 +79,7 @@ function renderTimestamp(created, highlight) {
   let cls = 'CommentHeader__timestamp';
   if (highlight) { cls += ' m-highlight'; }
 
-  return <div className={ cls }>{ short(created) }</div>;
+  return <div className={ cls }>{ formatElapsedTime(created) }</div>;
 }
 
 function renderGold(gildCount) {
--- a/src/app/components/MessageThread/Message.jsx
+++ b/src/app/components/MessageThread/Message.jsx
@@ -2,7 +2,7 @@ import './Message.less';
 import React from 'react';
 import { Anchor } from '@r/platform/components';
 
-import { short } from 'lib/formatDifference';
+import formatElapsedTime from 'lib/formatElapsedTime';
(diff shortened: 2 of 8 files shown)


phil303:
:eyeglasses: @schwers @dhunten

umbrae:
My 2¢ which I feel only mildly - I think I'm pro removing the username from the listing page, because I can't see much of a reason that users would care at that stage (other than I suppose mods who are looking for problem posters, or small subreddits where people are known, but don't seem like sufficient enough use cases).

I think the date format is a bit more clear but I'm not sure that clarity is worth the extra screen real estate, especially when flair and distinguishments are taken into account.

phil303:
Good feedback. I'm not sure it is either which is why I ladened this down with pics. Ultimately, what I'm trying to solve is the ambiguity in older posts/comments. When I see the every comment in a thread with "1y" in the header or big numbers like "326d", why do we bother to show it at all? These things lose any real meaning (at least to me).

schwers:
I'm 50/50 on removing the username. Admittedly I only _want_ to know it on certain subs (large and small) where the top posters are relatively consistent -- it gives me a gauge for 'how much will I like the selftext?'. This happens in so few subs that its probably okay to remove. However it might be worth considering that to a user who makes a post, seeing their name make it to the top of the frontpage of their sub, or r/all, is kinda a nice reward / warm-fuzzy-feeling for them.

Date wise, I really like the new format. I only really use dates for actively panning-out events like announcements or news, or older posts that I come across when searching for stuff. This format makes timestamps more useful in both of those situations.

dhunten:
It's better to have the platforms similar, so removal of username is fine with me. The 'time ago' stamps should also be like the others too (`3h`). I also like the context the new date format will give. Space will be a problem, and I'm not sure how to fit flairs in too. The shortest I can think of is `Nov'15`.

PR #879, opened 2016-11-03
Unified urls - Match subreddit sorts to desktop

phil303:
Bug:
Currently 2X doesn't support urls that have the format of `/r/nba/top`,
where the subreddit sort is part of the path. As part of the unified url
push, we need to support this.

Fix:
Currently we support the sort as a query parameter but we'd like to
deprecate this over time. So we're going to first look for the sort in
the url then fallback to the query param then fallback to 'hot'.

There's a few other sorts that 2X doesn't support yet at all, such as
"gilded" or "ads". Per discussion with William, we're going to 404 these
(this is the status quo).

Demo:
subreddit sorts

:eyeglasses: @nramadas @wting @schwers

### DIFF
--- a/src/app/components/SortAndTimeSelector/index.jsx
+++ b/src/app/components/SortAndTimeSelector/index.jsx
@@ -87,37 +87,23 @@ const mapDispatchToProps = dispatch => ({
 });
 
 const mergeProps = (stateProps, dispatchProps, ownProps) => {
-  const { currentPage } = stateProps;
-  const { queryParams } = currentPage;
-  const { sort = SORTS.HOT } = queryParams;
+  const { currentPage: { url, urlParams, queryParams } } = stateProps;
+
+  const sort = urlParams.sort || queryParams.sort || SORTS.HOT;
   const time = ownProps.time || listingTime(queryParams, sort);
   const { navigateToUrl } = dispatchProps;
 
-  const onSortChange = sort => {
-    navigateToUrl(currentPage.url, {
-      queryParams: {
-        ...queryParams,
-        sort,
-      },
-    });
-  };
-
-  const onTimeChange = time => {
-    navigateToUrl(currentPage.url, {
-      queryParams: {
-        ...queryParams,
-        t: time,
-      },
-    });
-  };
-
   return {
-    ...stateProps,
-    ...ownProps,
-    onSortChange,
-    onTimeChange,
-    sort,
     time,
+    sort,
+    ...ownProps,
+    ...stateProps,
+    onTimeChange: time => navigateToUrl(url, { queryParams: { ...queryParams, t: time } }),
+    onSortChange: sort => {
+      // add the sort to the url by taking just the '/r/:subredditName' part
+      const newUrl = [ ...url.split('/').slice(0, 3), sort ].join('/');
+      navigateToUrl(newUrl);
+    },
   };
 };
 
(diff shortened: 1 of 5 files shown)


wting:
The logic is right but I don't qualify to review React code. Thanks for getting this out so quickly!

:running_man:

src/app/router/index.js
```diff
@@ -20,6 +21,8 @@ import { PostSubmitHandler, PostSubmitCommunityHandler } from './handlers/PostSu
 import UserRerouteHandler from './handlers/UserReroute';
 import Status404PageHandler from './handlers/Status404Page';
 
+const sorts = SUPPORTED_SORTS.join('|');
```
schwers: This seems like a constant, might be good to capitlize

phil303:
:haircut:

src/app/components/SortAndTimeSelector/index.jsx
```diff
@@ -87,37 +87,23 @@ const mapDispatchToProps = dispatch => ({
...
-    sort,
     time,
+    sort,
+    ...ownProps,
+    ...stateProps,
+    onTimeChange: time => navigateToUrl(url, { queryParams: { ...queryParams, t: time } }),
+    onSortChange: sort => {
+      // add the sort to the url by taking just the '/r/:subredditName' part
+      const newUrl = [ ...url.split('/').slice(0, 3), sort ].join('/');
+      navigateToUrl(newUrl);
```
schwers: 🎉  🎈

src/app/components/SortAndTimeSelector/index.jsx
```diff
@@ -87,37 +87,23 @@ const mapDispatchToProps = dispatch => ({
...
-    onTimeChange,
-    sort,
     time,
+    sort,
+    ...ownProps,
+    ...stateProps,
+    onTimeChange: time => navigateToUrl(url, { queryParams: { ...queryParams, t: time } }),
+    onSortChange: sort => {
+      // add the sort to the url by taking just the '/r/:subredditName' part
+      const newUrl = [ ...url.split('/').slice(0, 3), sort ].join('/');
```
schwers: So this line breaks when you're navigating on the frontpage, it creates `//top`. Somewhere down the line platform in-turn tries to `pushState('http://top/')` and that throws an exception. Instead it might be easier to build the url with the `urlParams` from platform. you could write something like

```js
const { subredditName } = urlParams;
const newUrl = `/${subredditName ? `r/${subredditName}/` : ''}${sort}`
```

schwers:
💅  Patch looks really close! There's a small issue on frontpage pagination, everything else seems in order.

Side questions: 
* Do we want to start redirecting the queryParam versions of these urls? 
* Do we need to check the canonical urls for the queryParam versions?

src/app/components/SortAndTimeSelector/index.jsx
phil303: 👍 I think this is a much nicer solution

phil303:
@schwers regarding your questions:
- I think we're ok not redirecting or at least I'd like to consider redirecting at the CDN level instead and then eventually removing it entirely.
- I'm not sure what you mean by this one. Can you give me an example?

phil303:
:haircut: Again thanks for the find! That was a big oversight on my part.
Here's a demo of the fix. I've also tested to make sure the sort query param works on the frontpage and also that the subreddits still work.
fixed

schwers:
:fish:

PR #880, opened 2016-11-04
Fix outbound links in Safari

schwers:
The onTouchStart handler of the OutboundLinkComponent
wasn't passing the correct outboundLink. This should
have been tested more in Safari as well, but Android
Chrome was working fine, presumbably because its
events are more standardized.

👓  @nramadas


nramadas:
:fish:

PR #881, opened 2016-11-04
Add back in RelevantContent

phil303:
This is mostly just reverting an accidental deletion, but with a small
cleanup. It's possible to get a React exception if you don't return a
React element or null. This code has a code path that could possible
return undefined, which was sometimes causing an exception.

I've tested this and got it working under very select conditions, which
are:
1. User is logged out.
2. User has gotten to the comments page via a specific subreddit page
(not via the frontpage for example). The reason being that landing on
the subreddit page populates the necessary data for the relevant content
UX.
3. User has to have the "hot" sort set in the url (no other sorts work
by design looks like) when he hits the listing page. Hot just being on
by default doesn't seem to get this to show.

Talking with Andy, we're ok with this for now since this was the
previous status quo.

:eyeglasses: @schwers

### DIFF
--- a/src/app/components/RelevantContent/index.jsx
+++ b/src/app/components/RelevantContent/index.jsx
@@ -177,6 +177,8 @@ function RelevantContent(props) {
       />
     );
   }
+
+  return null;
 }
 
 const makeSelector = createSelector(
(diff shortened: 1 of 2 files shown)


phil303:
One thing to note with this PR, the `RelevantContent` component does use the feature flag code but only internally. So complete comment page re-renders won't happen.

phil303:
closing as this is most likely not the right fix

phil303:
Re-opening as a temporary measure until the real work is done

schwers:
💅  Just a small linter error, looks good for getting this re-enabled

phil303:
:haircut:

schwers:
:fish:

PR #882, opened 2016-11-04
Don't request a new ad if an old one exists already

jamesmaa:
👓 @nramadas

### DIFF
--- a/src/app/actions/ads.js
+++ b/src/app/actions/ads.js
@@ -92,16 +92,21 @@ export const fetchNewAdForPostsList = (postsListId, pageParams) =>
     const adRequest = state.adRequests[postsListId];
     if (adRequest && adRequest.loading) { return; }
 
-    const adId = nextAdId();
-
-    dispatch(fetching(adId, postsListId));
-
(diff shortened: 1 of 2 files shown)


src/app/actions/ads.js
```diff
@@ -89,27 +96,37 @@ const nextAdId = () => (uniqueId('ad_'));
...
+
+    const adId = nextAdId();
     dispatch(fetching(adId, postsListId));
+    if (adRequest && !adRequest.pending && !adRequest.failed) {
+      // Don't fetch a new ad if one exists already
+      dispatch(copyAd(adId, adRequest.ad));
+      return;
+    }
 
-    const loadedState = getState();
```
jamesmaa: `getState` was getting called twice

phil303: this is most likely done on purpose since `getState` will always get the latest state, whereas just using the original pointer risks it getting stale - basically if there are any dispatches in between when you set state and when you decide to use it.

jamesmaa: oh interesting! thanks for the insight

PR #884, opened 2016-11-08
Restrict the amphtml link tag to 2% of content

prashtx:
Create a basic page-bucketing mechanism, so we can restrict a feature to
a percentage of content. This is a temporary mechanism and not a
full-fledged page-based experimentation/flagging framework.

Restrict the amphtml link tag to just 2% of content, so that not all of
our self posts show up in search results as their AMP'd versions.

### DIFF
--- a/src/app/featureFlags.js
+++ b/src/app/featureFlags.js
@@ -1,9 +1,11 @@
 import Flags from '@r/flags';
 import omitBy from 'lodash/omitBy';
 import isNull from 'lodash/isNull';
+import sha1 from 'sha1';
 
 import getSubreddit from 'lib/getSubredditFromState';
 import getRouteMetaFromState from 'lib/getRouteMetaFromState';
+import getContentId from 'lib/getContentIdFromState';
 import url from 'url';
 import { getEventTracker } from 'lib/eventTracker';
 import { getDevice, IOS_DEVICES, ANDROID } from 'lib/getDeviceFromState';
@@ -27,6 +29,7 @@ const {
   VARIANT_XPROMO_CLICK,
   VARIANT_TITLE_EXPANDO,
   VARIANT_MIXED_VIEW,
+  SHOW_AMP_LINK,
 } = flagConstants;
 
 const config = {
@@ -170,6 +173,13 @@ const config = {
         { url: 'mixedview'},
       ] },
     ],
+  },
+  [SHOW_AMP_LINK]: {
+    url: 'showamplink',
+    pageBucketPercent: {
+      seed: 'showamplink',
+      percentage: 2,
+    },
   },
 };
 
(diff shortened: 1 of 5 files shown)


prashtx:
👓  @uzi @phil303

src/lib/getContentIdFromState.js
```diff
@@ -0,0 +1,23 @@
+import { has } from 'lodash/object';
```
phil303: this should be `import has from 'lodash/has';`.

src/lib/getContentIdFromState.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default function getContentId(state) {
```
phil303: did you mean to call this `getContentIdFromState`?

src/lib/getContentIdFromState.js
```diff
@@ -0,0 +1,23 @@
...
+  // Use post ID for comments pages and comment permalinks
+  if (state.platform &&
```
phil303: This state.platform check feels a little unnecessary

src/app/featureFlags.js
```diff
@@ -175,4 +185,35 @@ flags.addRule('seoReferrer', function (wantSEO) {
...
 });
 
+// NOTE (prashant.singh - 07 November 2016): This is interim functionality to
+// allow simple feature flagging for a percentage of pages. It should not be
+// used for true page or user experiments.
+// Bucket pages based on content ID, using granularity of 1/10th of a percent.
+flags.addRule('pageBucketPercent', function(config) {
+  const { seed, percentage } = config;
+  const contentId = getContentId(this.state);
+  //const id = convertId(contentId);
```
phil303: commented code

phil303:
:nail_care: some nitpicky stuff. Structurally looks good. 

Not that this will be a problem for this particular piece of code but just to spread the knowledge, the feature code as currently architected is fairly dangerous to use client side since it returns a new feature object on every state change. There's an mweb ticket to fix this up when we get the chance so the `addFlag` and `withContext` api may be subject to change.

src/lib/getContentIdFromState.js
prashtx: I'm following the model of `getSubredditFromState.js` with this helper, hence the shorter export name. I like the consistency and brevity, and we pass in `this.state` when we call it. But I'm fine changing this if we want to be more explicit with the name.

src/lib/getContentIdFromState.js
prashtx: No, there shouldn't be. Will rephrase the condition.

prashtx:
:haircut_man:

schwers:
@prashtx does this still need review? Happy to jump in if needed since P.hil is out

prashtx:
If you don't mind a pass, that would be great. It's unclear if we'll roll this out or not, but having it ready to deploy would be really handy in case we make the call to use it.

src/lib/getContentIdFromState.js
```diff
@@ -0,0 +1,22 @@
...
+  // Use post ID for comments pages and comment permalinks
+  if (has(state, ['platform', 'currentPage', 'urlParams', 'postId']) &&
+      has(state, ['commentsPages', 'current'])) {
+    // Vanilla comments page and comment permalink pages will have an entry in
+    // state.commentsPages
+    const commentsPage = state.commentsPages[state.commentsPages.current];
+    return commentsPage.postId;
+  }
+
+  // Use the subreddit fullname for subreddit listings
```
schwers: Technically this will be true for subreddit about and wiki pages, is that okay?

src/app/featureFlags.js
```diff
@@ -175,4 +185,34 @@ flags.addRule('seoReferrer', function (wantSEO) {
...
+
+  // hashed is a 160-bit number expressed as a hex string.
+  // We want to find (hashed % 1000), so we can map the hash to bucket sizes
+  // with 0.1% granularity. We can't work directly with 160-bit values as
+  // JavaScript Numbers, so we compute the modulo in pieces.
+  const piece3 = parseInt(hashed.slice(0,10), 16);
+  const piece2 = parseInt(hashed.slice(10,20), 16);
+  const piece1 = parseInt(hashed.slice(20,30), 16);
+  const piece0 = parseInt(hashed.slice(30,40), 16);
+  // 2^120 mod 1000 = 576
```
schwers: Might be helpful to have the math for how 120/80/40 are computed in a comment, or maybe we could add a function that given a the hash, range, and ring size could give you the overall modulo. (the latter might be overkill, but this is the trickiest part of the patch to follow)

src/lib/getContentIdFromState.js
prashtx: Yes, that's no problem right now. If we implement a page-based experimentation system for mweb, then we'll need to generalize some of the code from this PR and clean up the edge cases. But there's no immediate plan for that.

prashtx:
:haircut_man:

schwers:
Some minor rebasing is required, but :fish:!!

PR #886, opened 2016-11-15
Switch from Tune to Branch for the Mobile Native banner

prashtx:
We've switched our deep linking provider from Tune to Branch.

The first commit is @uzi's work to swap in the Branch SDK to "code complete" level. Their SDK assumes it will run in a browser, so the followup commit lets us avoid importing the SDK for code run on the server.

### DIFF
--- a/blueprints.config.js
+++ b/blueprints.config.js
@@ -21,6 +21,7 @@ module.exports = function(isProduction) {
     {
       generator: 'set-node-env',
       'process.env': {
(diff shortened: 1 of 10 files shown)


prashtx:
👓  @uzi @schwers

uzi:
:fish:

schwers:
@prashtx can we close this?

prashtx:
Yes, this is no longer necessary.

PR #889, opened 2016-11-17
Update post submit param name

jackniu1:
👓 @schwers 

Seems like the post param was not set correctly.

I tried with this param name, and could see the
header `g-recaptcha-response`:

screen shot 2016-11-17 at 12 11 40 am

Thanks


schwers:
Ah, I see now: https://github.com/reddit/node-api-client/commit/fbd92423bb1d0c8ef6ae2ef78b0ef62f9639a964#diff-a000cf73d272f08500227f2705d2e806R55

:fish:

PR #891, opened 2016-11-18
Promoted User Post events

jamesmaa:
👓 @dwick 

@drunken-economist are these event topic and names okay?


dwick:
i think you would normally have 1 event for `ad_click` and a field for where the ad was clicked. ie. title, description, etc.

dwick:
💅 i'll re-review after @drunken-economist checks the payloads.

dwick:
:runner:

PR #892, opened 2016-11-18
Branch + Cross promotion experiment

prashtx:
Switch to branch for our deep linking provider

Implement a cross-promotion interstitial experiment with 3 variants, shown only for direct visits to the front page.

We use client-generated long-form branch links, so we can avoid making 2-3 round trips to the API.

Most of the `shouldShowBanner` logic has been rolled into the feature flag system, so we can control this consistently and centrally. Because we can generate links client-side, we can create them as deterministic functions of state for a component that needs them.


prashtx:
👓  @uzi @schwers

src/app/components/InterstitialPromo/index.jsx
```diff
@@ -0,0 +1,141 @@
...
+      utm_name: 'xpromo_interstitial',
+      utm_content: 'element_1',
+    }),
+    getBranchLink(state, {
+      feature: 'interstitial',
+      campaign: 'xpromo_interstitial',
+      utm_medium: 'interstitial',
+      utm_name: 'xpromo_interstitial',
+      utm_content: 'element_2',
+    }),
```
uzi: 🔕  You could clean this up a pinch.

uzi:
Did a quick live review with @prashtx -- looks good to me.  I'll give it another pass tonight.

prashtx:
Added a couple of minor commits and split the experiment into iOS and Android experiments.

:haircut_man:

src/app/featureFlags.js
```diff
@@ -3,24 +3,36 @@ import omitBy from 'lodash/omitBy';
...
+  VARIANT_XPROMO_RATING,
 } = flagConstants;
 
 const config = {
   [BETA]: true,
-  [SMARTBANNER]: true,
+  [SMARTBANNER]: {
+    and: [
+      { allowedPages: ['index', 'listing'] },
+      { allowedDevices: IOS_DEVICES.concat(ANDROID) },
```
uzi: 🔕 You also have `RECOGNIZED_DEVICES` which you can import and use.

uzi:
🐟

PR #894, opened 2016-11-18
Record interstitial dismissal on navigation

prashtx:
When a user follows the call to action from the cross-promotion
interstitial, we don't want to show them the promotion again. To satisfy
that, we set the flag to record that they've closed the promotional UI,
and then we navigate to the Branch link via javascript.


prashtx:
👓  @uzi

uzi:
:fish:

PR #897, opened 2016-11-21
Issue an event when xpromos are dismissed

prashtx:
We need to log a user preference change event when users dismiss the cross-promotional experiences. The event originally covered cookie-based preferences, but we're overloading it to include localStorage and other non-account preferences. We also use a different name for the preference on advice of DA, so that the field is more descriptive in our data pipeline. Changing the name we use inside the app would introduce some risk when migrating the old setting to the new name, which is not ideal while we have ongoing experiments related to this.


uzi:
:fish:

PR #899, opened 2016-11-22
Log smartbanner preference change more carefully

prashtx:
Don't try to log a preference change event when the user follows the
link in the banner's call to action. We really just want to track when
users bail out of the crosspromo experience.


prashtx:
👓  @uzi @drunken-economist

uzi:
:fish:

drunken-economist:
🐟

PR #902, opened 2016-11-23
add experiment to open expando on title click

curioussavage:
https://reddit.atlassian.net/browse/WEB-2356

### DIFF
--- a/src/app/actions/posts.js
+++ b/src/app/actions/posts.js
@@ -4,11 +4,44 @@ const { SavedEndpoint, HiddenEndpoint, EditUserTextEndpoint } = endpoints;
 
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 
+import { getEventTracker } from 'lib/eventTracker';
+import { getBasePayload, buildSubredditData, getListingName, getUserInfoOrLoid } from 'lib/eventUtils';
+
 export const TOGGLE_EXPANDED = 'POSTS__TOGGLE_EXPANDED';
 export const toggleExpanded = postId => ({
   type: TOGGLE_EXPANDED,
   postId,
 });
+
+export const toggleExpandPost = (postId, clickTarget) => async (dispatch, getState) => {
+  dispatch(toggleExpanded(postId));
+  trackExpandToggle(getState(), postId, clickTarget);
+};
+
+function trackExpandToggle(state, postId, clickTarget) {
+  const eventType = state.expandedPosts[postId] ? 'cs.expand_user' : 'cs.collapse_user';
+  getEventTracker().track('expando_events', eventType, {
+    ...getBasePayload(state),
+    ...buildSubredditData(state),
+    ...getListingName(state),
+    ...getPostInfo(state, postId),
+    ...getUserInfoOrLoid(state),
+    'click_target': clickTarget,
+  });
+}
+
+function getPostInfo(state, postId) {
+  const post = state.posts[postId];
+  return {
+    'target_author_id': post.id,
+    'target_created_ts': post.createdUTC,
+    'target_id': post.id,
+    'target_url_domain': post.domain,
+    'target_url': post.cleanUrl,
+    'target_type': post.isSelf ? 'self' : 'link',
+    'nsfw': post.over18,
+  };
+}
 
 export const TOGGLE_NSFW_BLUR = 'POSTS__TOGGLE_NSFW_BLUR';
 export const toggleNSFWBlur = postId => ({
--- a/src/app/components/Post/PostHeader/index.jsx
+++ b/src/app/components/Post/PostHeader/index.jsx
@@ -65,6 +65,7 @@ PostHeader.propTypes = {
   renderMediaFullbleed: T.bool.isRequired,
   showLinksInNewTab: T.bool.isRequired,
   onElementClick: T.func.isRequired,
+  titleOpensExpando: T.bool.isRequired,
 };
 
 function postTextColorClass(distinguished) {
@@ -320,7 +321,8 @@ function renderPostHeaderLink(post, showLinksInNewTab) {
   );
 }
 
-function renderPostTitleLink(post, showLinksInNewTab, onElementClick) {
+function renderPostTitleLink(post, showLinksInNewTab, onElementClick,
+                             titleOpensExpando, onTapExpand) {
   const linkExternally = post.promoted && !post.isSelf;
   const url = linkExternally ? post.cleanUrl : cleanPostHREF(mobilify(post.cleanPermalink));
   const { title } = post;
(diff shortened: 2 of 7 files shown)


curioussavage:
👓 @schwers @nramadas

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -333,6 +334,10 @@ function renderPostTitleLink(post, showLinksInNewTab) {
     target, 
   };
 
+  if (titleOpensExpando && !linkExternally) {
+    props.onClick = (e) => { e.preventDefault(); onTapExpand(); };
```
nramadas: might be cleaner to just multiline the function

src/app/components/Post/index.jsx
```diff
@@ -189,7 +200,8 @@ const selector = createSelector(
   (state, props) => !!state.expandedPosts[props.postId],
   (state, props) => !!state.unblurredPosts[props.postId],
   (state, props) => state.editingText[props.postId],
-  (user, postId, single, compact, post, expanded, unblurred, editingState) => {
+  featuresSelector,
```
schwers: This pattern of using the `featuresSelector` should be depreciated imo. The features selector re-creates a new instance of the features collection on every state iteration, this will cause `connect` to re-render every time anything in state changes (this was the biggest issue with comments voteing performance). I would recommend having a selector that returned the boolean flag you end up using, as this will prevent unnecessary re-renders because `reselect` uses `===` comparison (hence the problem with new object instances, the downside is creating a features collection for every post every time state is updated, but this route requires less changes than say, computing the flag every componet that can render a list of posts).

In practice I'd imagine this would look like:

```js
state => features.withContenct({ state }).enabled(VARIANT_TITLE_EXPANDO),
```

Then you could pass down a `inTitleExpandoVariant` boolean to use in computing `titleOpensExpando`

src/app/components/Post/PostHeader/index.jsx
schwers: small nit: we've been omitting the parens on single arg arrow functions

schwers:
💅

curioussavage:
:haircut_man: 

Any reason why I wouldn't want to just move the selector for the feature to postsList? seems like that would be slightly better. In my next task I'm going to have to add another selector for a test so thats 2 feature objects created per rendered post already

schwers:
@curioussavage The only reason I can think of is extra complexity. I think you'll just need to make sure it works for `PostsList` and `PostAndCommentList` (in theory you'll want to make sure the Ad component passes it through, but Ads `linkExternally`? so that should be fine)

schwers:
💅  Looks good! I'll leave the call on `lifting` the bucketing test to parent components. Otherwise just a small nit.

curioussavage:
:haircut_man:  

decided not to move the selectors up. Seems to just add work/complexity for no significant benefit

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -343,7 +345,17 @@ function renderPostTitleLink(post, showLinksInNewTab, onElementClick) {
...
 
-  const anchorProps = { ...props, onClick: onElementClick };
+  const anchorProps = { ...props, 
+    onClick: e => {
+      if (titleOpensExpando) {
+        e.preventDefault();
+        onTapExpand(); 
+      }
+
+      onElementClick();
```
nramadas: do we want to do both `onTapExpand` and `onElementClick`?

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -343,7 +345,17 @@ function renderPostTitleLink(post, showLinksInNewTab, onElementClick) {
     );
   }
 
-  const anchorProps = { ...props, onClick: onElementClick };
+  const anchorProps = { ...props,
```
nramadas: `...props` should be on the next line

src/app/components/Post/PostHeader/index.jsx
schwers: We need to run `onElementClick` as the xpromo click modal variant code should still run on title click. (It will appear above and simultaneously to the expando _iff_ the xpromo conditions are met)

curioussavage: yes

curioussavage:
:haircut_man:

schwers:
:fish:

curioussavage:
👓  @umbrae

umbrae:
@curioussavage I want to test but to ask: when a user clicks the title, does it send an expando event? And if so, does that expando event have some sort of attribute that will let us know on the backend that that expando event came from a title click?

curioussavage:
Expando event is almost done. it doesn't have an attribute indicating where the click came from though...

curioussavage:
:haircut_man: 

Should I leave the tracking as a separate commit?

src/app/actions/posts.js
```diff
@@ -4,12 +4,47 @@ const { SavedEndpoint, HiddenEndpoint, EditUserTextEndpoint } = endpoints;
...
+  });
+}
+
+function getPostInfo(state, postId) {
+  const post = state.posts[postId];
+  return {
+    'target_author_id': post.id,
+    'target_created_ts': post.createdUTC,
+    'target_id': post.id,
+    'target_url_domain': url.parse(post.cleanUrl).host,
```
schwers: For self-posts, this will be null, is that okay and/or intentional?

Relatedly: could we use `post.domain` here instead?  (note: For self-posts it it will be in the form of `self.${subredditName}`)

src/lib/eventUtils.js
```diff
@@ -31,12 +31,33 @@ export function buildSubredditData(state) {
   };
 }
 
+export function getListingName(state) {
+  const subredditName = state.platform.currentPage.urlParams.subredditName;
+  return { 'listing_name': subredditName ? subredditName : 'frontpage'};
+  // we don't support multis or all yet but we will need to update this when we do.
```
schwers: r/all works, do you mean filtering?

src/app/actions/posts.js
curioussavage: nice. I didn't notice the api returns that!

curioussavage: I think the null was fine but this is better

src/lib/eventUtils.js
curioussavage: ha... I think I had a typo when I was trying out r/all on my dev instance :(

schwers: :(

src/app/featureFlags.js
```diff
@@ -151,6 +152,12 @@ const config = {
     ],
   },
   [VARIANT_XPROMO_CLICK]: false,
+  [VARIANT_TITLE_EXPANDO]: {
+    and: [
+      { compact: true},
+      { variant: 'mweb_post_title_expando:active'},
```
schwers: 🔕 It would be useful to have an `or` clause that would let us enable this with a url queryParam for testing

src/app/actions/posts.js
curioussavage: justin says post.domain is fine as is.

schwers:
:fish: :tropical_fish:

umbrae:
> it doesn't have an attribute indicating where the click came from though... is it enough to just look at the change in number of expandos for users in the experiment or do we need to do that?

Something that happened while you were on leave is that we had a hard time analyzing some experiments without information like this, although it is possible. 

If you have a spare few minutes, and want to add a `referrer_element` param to the expando event (as recommended by @drunken-economist), I think it would be helpful in analysis. If it's going to be > 2 hours of work though, we can work on speccing this better generally for next time.

PR #903, opened 2016-11-23
Add more padding to xpromo dismissal link

schwers:
👓  @uzi or @prashtx 

There's been some feedback that the dismiss button on xpromo modals is hard to click on. This patch adds more padding to the link. It adds more padding to the bottom as the link is rendered near the bottom edge of the screen, and adding too much to the top could accidentally eat clicks intended for the app link button.


uzi:
:fish:

PR #905, opened 2016-11-28
Enable logged-out outbounds

schwers:
This enables logged-out outbounds. The core feature work here is really simple, we now always send the `redditWebClient` queryParam.

This also fixes a critical bug in the OutboundLink component that prevented tracking clicks on links with nested dom elements. This slipped through QA as there's only one instance of this, but its an important one, the linkbar on comments pages.

👓  @curioussavage @nramadas

### DIFF
--- a/src/lib/findNearestLinkElement.js
+++ b/src/lib/findNearestLinkElement.js
@@ -1,0 +1,11 @@
+export default function findNearestLinkElement(el) {
+  if (el.tagName === 'A') {
+    return el;
+  }
+
+  if (el.parentNode) {
+    return findNearestLinkElement(el.parentNode);
+  }
+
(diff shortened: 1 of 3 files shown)


src/lib/findNearestLinkElement.js
```diff
@@ -0,0 +1,11 @@
...
+  if (el.tagName === 'A') {
+    return el;
+  }
+
+  if (el.parentNode) {
+    return findNearestLinkElement(el.parentNode);
+  }
+
+  return null;
+}
```
nramadas: missing newline

schwers:
:haircut_man:

nramadas:
:fish:

PR #907, opened 2016-11-30
Remove params_app from the crawl event payload

prashtx:
Requests from apps like mweb to the r2 API include an `app` parameter.
We pass that along in r2 crawl events as `params_app`, so that we can
distinguish those API requests (caused by crawls of apps) from real,
direct crawls of the desktop site. We shouldn't include that in the mweb
crawl events, since these are direct page crawls and not API requests.
The `app_name` field will still be passed along by the event tracker
module, and we should use that to distinguish between desktop and mweb
crawls.


prashtx:
👓  @telaviv

telaviv:
🐟

PR #909, opened 2016-11-30
Turn select xpromo variants to 100%

prashtx:
As a temporary measure, we are showing the list/subreddit treatments
(frontpage/listings experiments) also to users bucketed into the control
groups. We want to continue to get bucketing events, so we know when a
user has been exposed to this feature, and we want to show 100% of users
the select treatments. We can't eliminate control groups in the API's
bucketing mechanism, so we use this workaround instead.

### DIFF
--- a/src/app/featureFlags.js
+++ b/src/app/featureFlags.js
@@ -84,26 +84,12 @@ const config = {
       seoReferrer: true,
     }],
   },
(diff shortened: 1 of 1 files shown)


prashtx:
👓  @uzi

uzi:
:fish: also sounds legit

PR #910, opened 2016-11-30
Work around an API issue with the redditWebClient query param

prashtx:
The API doesn't return experiment config data when we include the
redditWebClient query param on calls to /api/me.json, so we're working
around it for shell renders, which are the only situations where we call
that endpoint and where we don't strictly need the param right now.


prashtx:
👓  @schwers

schwers:
:fish:

schwers:
closed via a6a75234d3c69fdbe3dbef6fe2c340b766289455

PR #911, opened 2016-12-02
Don't recurisvely retry to refresh access tokens forever

schwers:
👓  @curioussavage


curioussavage:
review done in person thursday.

lgtm 🐟

PR #912, opened 2016-12-02
Add post to subreddit experiment

keithblaha:
:eyeglasses: @scarow @schwers 

@schwers - i heard the experiment framework for 2X was/is being rewritten, am i using it right?

note: requires https://github.com/reddit/node-api-client/pull/179


src/app/reducers/subredditsByPost.js
```diff
@@ -0,0 +1,19 @@
...
+import * as subredditsByPostActions from 'app/actions/subredditsByPost';
+
+const DEFAULT = {};
+
+export default (state=DEFAULT, action={}) => {
+  switch (action.type) {
+    case subredditsByPostActions.RECEIVED_SUBREDDITS_BY_POST: {
+
+      return merge(state, {
+        [action.postId]: Object.values(action.apiResponse.subreddits),
```
schwers: We should track the subreddits by id here, and merge the models in reducers/subreddits

src/app/featureFlags.js
```diff
@@ -74,6 +75,14 @@ const config = {
       seoReferrer: true,
     }],
   },
+  [VARIANT_RECOMMENDED_BY_POST]: {
+    url: 'experimentrecommendedbypostsrbypost',
+    and: [{
+      variant: 'subreddits_by_post:sr_by_post',
```
schwers: The logged in check should be first so we don't fire bucketing events for users that are inelgible

keithblaha:
:haircut_woman:

keithblaha:
bump and adding :eyeglasses: @nramadas

keithblaha:
FYI i'll be adding 1 more thing (some click event tracking)

src/app/actions/subredditsToPostsByPost.js
```diff
@@ -0,0 +1,75 @@
...
+    if (some(LEGAL_VARIANTS, variant => feature.enabled(variant))) {
+      dispatch(fetching(post.id));
+      const apiOptions = apiOptionsFromState(state);
+
+      try {
+        const response = await SubredditsToPostsByPost.get(
+          apiOptions,
+          {
+            link: post.name,
+            variant: 'nb',
```
scarow: What is 'nb'?

src/app/components/RecommendedSubreddits/index.jsx
```diff
@@ -37,7 +37,7 @@ export default function RecommendedSubreddits(props) {
     subredditListing = (
       <div className={ cssClass }>
         <Anchor
-          href={ addUtmTracking(currentSubreddit.url, 0, variant) }
+          href={ currentSubreddit.url }
```
scarow: Data team doesn't want utm tracking?

src/app/pages/Comments.jsx
```diff
@@ -98,6 +119,8 @@ CommentsPage.propTypes = {
   currentPage: T.object.isRequired,
   preferences: T.object.isRequired,
   isCrawlerRequest: T.bool.isRequired,
+  recommendedSubreddits: T.arrayOf(T.object).isRequired,
+  recommendedPosts: T.arrayOf(T.object).isRequired,
```
scarow: Should these be required for all `CommentsPage`? Not sure

src/app/pages/Comments.jsx
```diff
@@ -121,8 +144,30 @@ const stateProps = createSelector(
...
+  },
+  (state, props) => {
+    const pageParams = CommentsPageHandler.pageParamsToCommentsPageParams(props);
+    const post = state.posts[pageParams.id];
+    const postLoaded = !!post;
+    let recommendedPosts = null;
+    if (postLoaded && post.id in state.subredditsToPostsByPost) {
+      const postNames = state.subredditsToPostsByPost[post.id];
+      recommendedPosts = Object.values(state.posts).filter(post => postNames.includes(post.uuid));
+    }
```
scarow: Can `postNames` ever be `null`? If so, what happens? Does this break?

src/app/pages/Comments.jsx
```diff
@@ -133,6 +178,8 @@ const stateProps = createSelector(
       currentPage,
       preferences,
       isCrawlerRequest,
+      recommendedSubreddits: !recommendedSubreddits ? [] : recommendedSubreddits,
+      recommendedPosts: !recommendedPosts ? [] : recommendedPosts,
```
scarow: IMO it's easier to read this without the `!`, thereby flipping the logic. Like:

`recommendedPosts: recommendedPosts ? recommendedPosts : [],`

Up to you

src/app/pages/Comments.jsx
keithblaha: the check inside if statement should prevent that from happening

src/app/pages/Comments.jsx
keithblaha: i agree but i was following the general style throughout the file (better to be locally correct than globally correct)

keithblaha:
bump- please review, tracking is now in and this is functionally complete. we are trying to deploy this tomorrow

keithblaha:
adding :eyeglasses: @prashtx @nramadas

src/app/pages/Comments.jsx
keithblaha: it's expected that itll be [] if none

src/app/components/RecommendedSubreddits/styles.less
```diff
@@ -68,10 +73,6 @@
     margin: 0px 40px;
     padding: 10px 10px 0px 10px;
     text-align: center;
-
-    .themeify({
-      color: @theme-toast-color;
-    });
```
scarow: Can;t remember what this is for -- can you attach a screen shot of what this looks like in night mode?

src/app/router/handlers/CommentsPage.js
```diff
@@ -88,8 +94,9 @@ export default class CommentsPage extends BaseHandler {
       subredditName = post.subreddit;
     }
 
-    if (subredditName) {
-      fetchRecommendedSubreddits(state, dispatch, subredditName);
+    if (post) {
+      fetchRecommendedSubredditsByPost(state, dispatch, post);
+      fetchRecommendedSubredditsToPostsByPost(state, dispatch, post);
```
scarow: Why are we fetching both of these? Isn't this dependent on variant?

src/app/components/RecommendedSubreddits/styles.less
keithblaha: been through review with design team and they're happy with how it looks:

subreddits_nightv5

src/app/router/handlers/CommentsPage.js
keithblaha: the bucketing check is inside the fetch functions

src/lib/eventUtils.js
```diff
@@ -85,6 +85,17 @@ function trackScreenViewEvent(state, additionalEventData) {
...
 
+export function trackExperimentClickEvent(state, experimentName, targetThing) {
+  const payload = {
+    ...getBasePayload(state),
+    'experiment_name': experimentName,
+    'target_fullname': targetThing.name,
+    'target_url': targetThing.url,
+    'target_type': targetThing.type,
+  };
+  getEventTracker().track('internal_click_events', 'cs.experiment_clicks', payload);
```
scarow: has this been approved by data? Not sure what event names should look like.

src/app/router/handlers/CommentsPage.js
```diff
@@ -116,6 +123,35 @@ const fetchRecommendedSubreddits = (state, dispatch, subredditName) => {
...
 };
 
+const fetchRecommendedSubredditsByPost = (state, dispatch, post) => {
+  let experimentId = 105;
+  dispatch(subredditsByPostActions.fetchSubredditsByPost(post, experimentId));
+};
+
+const fetchRecommendedSubredditsToPostsByPost = (state, dispatch, post) => {
+  const feature = features.withContext({ state });
+  let experimentId = 105;
```
scarow: Should we store `experimentId` so we don't have to dupe it in `fetchRecommendedSubredditsByPost` and `fetchRecommendedSubredditsToPostsByPost`?

src/lib/eventUtils.js
keithblaha: it's already speced out https://reddit.atlassian.net/wiki/display/DA/Internal+Click+Events

src/app/router/handlers/CommentsPage.js
keithblaha: ultimately thisll be ripped out when its featurized so probably not worth the effort

src/app/components/Post/index.jsx
```diff
@@ -182,7 +185,7 @@ export function Post(props) {
           showingLink={ !!(displayCompact && !hasExpandedCompact && externalDomain) }
           renderMediaFullbleed={ renderMediaFullbleed }
           showLinksInNewTab={ showLinksInNewTab }
-          onElementClick={ onElementClick }
+          onElementClick={ () => { onPostClick(post); onElementClick() } }
```
keithblaha: there should be full internal click tracking across the board.... but, note that this is an experiment click event, not a generic click event. so we should only fire these types of events for experiments

scarow:
🐟  lgtm -- I would wait on the review of an mweb person for any subtleties I might be missing!

keithblaha:
thank you @scarow!

we are about a week late pushing this experiment out, so anyone who'd like to review it please try to today! aiming to deploy this tomorrow

prashtx:
It looks like there are some linter errors that travis-ci is catching.

src/app/reducers/subreddits.js
```diff
@@ -17,6 +18,7 @@ export default function(state=DEFAULT, action={}) {
 
     case recommendedSubredditsActions.RECEIVED_RECOMMENDED_SUBREDDITS:
     case searchActions.RECEIVED_SEARCH_REQUEST:
+    case subredditsByPostActions.RECEIVED_SUBREDDITS_BY_POST:
```
prashtx: Can you add this to the tests for this reducer?

prashtx:
💅

src/app/reducers/subreddits.js
keithblaha: is that necessary? it is tested in its own file (included in this PR)

src/app/reducers/posts.js
```diff
@@ -45,6 +46,7 @@ export default function(state=DEFAULT, action={}) {
     case hiddenActions.RECEIVED_HIDDEN:
     case savedActions.RECEIVED_SAVED:
     case searchActions.RECEIVED_SEARCH_REQUEST:
+    case subredditsToPostsByPostActions.RECEIVED_SUBREDDITS_TO_POSTS_BY_POST:
```
keithblaha: is that necessary? it is tested in its own file (included in this PR)

keithblaha:
:haircut_woman:

src/app/pages/Comments.jsx
```diff
@@ -121,8 +152,30 @@ const stateProps = createSelector(
...
   state => state.platform.currentPage,
   state => state.preferences,
+  (state, props) => {
+    const pageParams = CommentsPageHandler.pageParamsToCommentsPageParams(props);
+    const post = state.posts[pageParams.id];
+    const postLoaded = !!post;
+    let recommendedSubreddits = null;
+    if (postLoaded && post.id in state.subredditsByPost) {
+      const subredditNames = state.subredditsByPost[post.id];
+      recommendedSubreddits = Object.values(state.subreddits).filter(subreddit => subredditNames.includes(subreddit.uuid));
```
schwers: I want to do a quick test, but I think using `Object.values` like this will end up creating a new object everytime state changes, which could cause re-renders for any state change. The first pass I usually do for testing this page is to log in and go to  heavily nested comments page (usually anything on the logged out frontpage or askreddit), go to timeline in dev tools and turn on 5x device throttling, start measuring, and upvote a nested comment and then stop measuring. It usually takes somewhere between 100ms and 200ms to handle the click event in prod with redux dev tools enabled.

src/app/reducers/subreddits.js
schwers: +1 for test

src/app/pages/Comments.jsx
prashtx: `filter` will create a new object either way. If this causes unnecessary rerenders and a related performance hit, would it be better to move the filtering logic into the combiner, so that the selectors produce the exact same objects for unrelated state changes?

schwers: Good point, and from testing `values` does a create a new object as well. 

From my tests there was a performance hit to voting and comment collapse, I'd suspect there's more than those two interactions given how this is tied to any state change. Ideally we should only be returning new object references from a selector like this if the data we need has changed. In this case I think we should be able to return `state.subredditsByPost[post.id]` to avoid new object references. At that point we'd either have to pass `state.subreddits` into the combiner to extract the subreddit objects, or connect the `RecommendedSubreddits` component to the store and pass it the list of recommended subreddit ids that it would extract in its own selector.

In the combiner approach: `state.subreddits` shouldn't change that often, so this would minimize the re-renders. The next selector for `subredditsToPostsByPost` would need to do something similar, but would require passing `state.posts` into the combiner as well. Given that would make for two 'hacks', I think it would be cleanest to connect the `RecommendedPosts` and `RecommendedSubreddits` components and pass them a list of id's they should be rendering. For this selector that would mean returning `state.subredditsByPost[post.id]` and in the next selector, `state.subredditsToPostsByPost[post.id]`.

schwers:
💅

src/app/components/RecommendedPosts/styles.less
```diff
@@ -0,0 +1,72 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/themes/themeify';
+
+.RecommendedPosts__container {
+  .themeify({
+    background-color: @theme-comment-body-color;
+  });
+  .nightMode & {
```
keithblaha: well thats handy!

src/app/pages/Comments.jsx
keithblaha: thank you for investigating, definitely wanna minimize performance impact. ill refactor as suggested

keithblaha:
partial :haircut_woman: 

i still need to add those reducer tests (will do tomorrow), but i did the refactoring for the recommendation components to be connected which i think will be the harder part to review

keithblaha:
added reducer tests. FYI i'll be adding one more endpoint to this to support another variant in the experiment... a direct post to post recommendation

src/app/pages/Comments.jsx
```diff
@@ -51,6 +53,14 @@ function CommentsPage(props) {
     <div className='CommentsPage'>
       <SubNav />
       <Post postId={ pageParams.id } single={ true } key='post' />
+      <RecommendedPosts
+        postId={ pageParams.id.substring(3) }
```
schwers: This is a really nice refactor, thanks for doing that! Just a small nit but I have a pretty strong preference for using the full post id so we're as opaque about ids as possible.

schwers:
Just as general note, @keithblah, @curioussavage is taking over a CONT ticket PR, https://github.com/reddit/reddit-mobile/pull/920, and these do sort of similiar things. Can you coordinate and make sure we don't show two of these relevancy experiments on the same comments page?

src/app/pages/Comments.jsx
keithblaha: Good point- and I agree, it's my bad for using ids instead of fullnames in the actions!

keithblaha:
Thanks for letting me know, I'll talk with @curioussavage about it tomorrow

This PR is finally in a final state now- no more functionality to add. We're aiming to push live monday! I'll bug you all for fish tomorrow :)

keithblaha:
Friendly bump- would love to have this fished today so we can launch on monday!

src/app/components/RecommendedPosts/styles.less
```diff
@@ -0,0 +1,70 @@
...
+    padding: 0;
+    width: 90%;
+
+    .themeify({
+      border-top: @theme-border;
+    });
+  }
+
+  .Post__header-wrapper {
+    padding-left: 16px;
```
schwers: small nit: I would try to use a multiple of `@grid-size` whenever possible for alignment and spacing like this. It will be more consistent across the other less files in the repo and keep things uniform visually

src/app/components/RecommendedSubreddits/index.jsx
```diff
@@ -1,104 +1,113 @@
...
-                    : {}
-                  }
-          />
+  const subredditListing = recommendedSubreddits.map((sr) => {
+    return (
+      <div className='RecommendedSubreddits__top'>
+        <div
+          className='subreddit-icon-image'
+          style={ sr.iconImage
+                  ? { 'backgroundImage': `url(${sr.iconImage})`,
```
schwers: Sorry for missing this, but you should use the keyColor here as well. There are some subreddits that put transparency in their icon and rely on the keycolor showing through for branding / themeing

src/app/components/RecommendedSubreddits/styles.less
```diff
@@ -23,6 +22,10 @@
   }
 }
 
+.RecommendedSubreddits__SubredditInfo {
+  margin-top: 8px;
```
schwers: Looks like a good place to use `@grid-size` too

src/app/components/RecommendedSubreddits/styles.less
```diff
@@ -34,19 +37,20 @@
...
+   }, @inNightMode: {
+      background-color: #222;
   });
 
   .subreddit-icon-image {
     width: @community-icon-size;
     height: @community-icon-size;
     float: left;
-    margin-right: 25px;
+    margin-right: 16px;
```
schwers: Another use case for `@grid-size`

src/app/components/RecommendedPosts/styles.less
keithblaha: so i mostly based the css off the old PR for 1X for similar subreddits experiment... sorry its not up to snuff. the good news is that design has been working on new designs for recommendations, so all of this CSS is throwaway!

src/app/router/handlers/CommentsPage.js
```diff
@@ -6,16 +6,23 @@ import { cleanObject } from 'lib/cleanObject';
 import { fetchUserBasedData } from './handlerCommon';
 import * as commentsPageActions from 'app/actions/commentsPage';
 import * as subredditActions from 'app/actions/subreddits';
-import * as recommendedSubredditsActions from 'app/actions/recommendedSubreddits';
+//import * as recommendedSubredditsActions from 'app/actions/recommendedSubreddits';
```
keithblaha: so were still working on the similar subreddits model (which goes subreddit->subreddit, as opposed to post->subreddit which were testing this PR)... so this might be something we use again. just wanted to silence lint errors as its not used

src/app/components/RecommendedSubreddits/index.jsx
keithblaha: good call, will do

PR #913, opened 2016-12-05
Support xpromo interstitials for in-app navigation

prashtx:
We need to support display of the cross-promo interstitials for the
appropriate pages for in-app navigation, but we also don't want to show
multiple interstitials per session (and in particular not for sequential
navigations). We track if we have already shown an xpromo experience,
and then we can avoid showing it for later pages.

### DIFF
--- a/src/app/actions/smartBanner.js
+++ b/src/app/actions/smartBanner.js
@@ -7,6 +7,12 @@ export const show = () => ({ type: SHOW });
 export const HIDE = 'SMARTBANNER__HIDE';
 export const hide = () => ({ type: HIDE });
 
+export const RECORD_SHOWN = 'SMARTBANNER__RECORD_SHOWN';
+export const recordShown = url => ({
+  type: RECORD_SHOWN,
+  url,
+});
+
 const EXTERNAL_PREF_NAME = 'hide_mweb_xpromo_banner';
 
 // element is the interface element through which the user dismissed the
--- a/src/app/components/InterstitialListing/index.jsx
+++ b/src/app/components/InterstitialListing/index.jsx
@@ -7,6 +7,8 @@ import { redirect } from '@r/platform/actions';
 import { markBannerClosed } from 'lib/smartBannerState';
 import * as smartBannerActions from 'app/actions/smartBanner';
 
+import XPromoWrapper from 'app/components/XPromoWrapper';
+
 import {
   InterstitialListingCommon,
   selector,
@@ -14,11 +16,13 @@ import {
 
 export function InterstitialListing(props) {
   return (
-    <div className='InterstitialListing'>
-      <div className='InterstitialListing__content'>
-        <InterstitialListingCommon { ... props } />
+    <XPromoWrapper>
+      <div className='InterstitialListing'>
+        <div className='InterstitialListing__content'>
+          <InterstitialListingCommon { ... props } />
+        </div>
       </div>
-    </div>
+    </XPromoWrapper>
   );
 }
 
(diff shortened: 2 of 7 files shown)


prashtx:
👓 @schwers @telaviv 

/cc @uzi

src/app/components/InterstitialListing/index.jsx
```diff
@@ -7,18 +7,22 @@ import { redirect } from '@r/platform/actions';
 import { markBannerClosed } from 'lib/smartBannerState';
 import * as smartBannerActions from 'app/actions/smartBanner';
 
+import XPromoWrapper from '../XPromoWrapper';
```
schwers: Nit: prefer absolute imports, you can write this as `app/components/XPromoWrapper`

src/app/actions/smartBanner.js
```diff
@@ -7,6 +7,12 @@ export const show = () => ({ type: SHOW });
 export const HIDE = 'SMARTBANNER__HIDE';
 export const hide = () => ({ type: HIDE });
 
+export const RECORD_SHOWN = 'SMARTBANNER__RECORD_SHOWN';
+export const recordShown = (url) => ({
```
schwers: Nit: preference for no parens in single arg arrow functions

src/app/components/InterstitialListing/modal.jsx
```diff
@@ -7,6 +7,7 @@ import { redirect } from '@r/platform/actions';
 import { markBannerClosed } from 'lib/smartBannerState';
 import * as smartBannerActions from 'app/actions/smartBanner';
 import * as modalActions from 'app/actions/modal';
+import XPromoWrapper from '../XPromoWrapper';
```
schwers: Nit: preference for absolute imports

src/app/reducers/smartBanner.js
```diff
@@ -1,10 +1,12 @@
 import merge from '@r/platform/merge';
-import * as platformActions from '@r/platform/actions';
+import platformActions from '@r/platform/actions';
```
schwers: We use the `import * as...` here as there is no default exports from `@r/platform/actions`, was there something broken with that in this case? If this is needed we probably need to make this change globally in the app.

src/app/reducers/smartBanner.js
```diff
@@ -20,8 +22,21 @@ export default function(state=DEFAULT, action={}) {
       return DEFAULT;
     }
 
+    case smartBannerActions.RECORD_SHOWN: {
```
schwers: Please add tests for this and the new `NAVIGATE_TO_URL` logic

schwers:
💅  (for harold)

src/app/reducers/smartBanner.js
prashtx: No, this was an artifact of removing `platformActions` and then later adding them back during development. Good catch.

src/app/reducers/smartBanner.js
prashtx: :+1:

prashtx:
:haircut_man:

src/app/reducers/smartBanner.test.js
```diff
@@ -32,5 +36,47 @@ createTest({ reducers: { smartBanner } }, ({ getStore, expect }) => {
...
+        // Indicate desire to show, record that we showed, navigate.
+        store.dispatch(smartBannerActions.show());
+        store.dispatch(smartBannerActions.recordShown('/foo/bar'));
+        store.dispatch(platformActions.navigateToUrl(METHODS.GET, '/user/foobar'));
+        const { smartBanner } = store.getState();
+        const { showBanner } = smartBanner;
+        expect(showBanner).to.eql(false);
+      });
+
+      it('should not remove desire to show xpromo if we not shown it yet', () => {
```
schwers: nit: 'if we not shown it yet' -> 'if we have not shown it yet'

schwers:
Small typo to address in the tests, but :fish:

PR #916, opened 2016-12-06
Add mixed view experiment

curioussavage:
this displays posts with expando as cards in contact view

[CONT-18](https://reddit.atlassian.net/browse/CONT-18)

👓 @schwers @madbook


src/lib/gifToHTML5Sources.js
```diff
@@ -41,6 +41,19 @@ function posterForHrefIfGiphyCat(url) {
 
 export { posterForHrefIfGiphyCat };
 
+const IMAGE_DOMAINS = [
```
schwers: This looks like a good place to use a `Set`.

src/app/components/Post/index.jsx
```diff
@@ -102,13 +106,19 @@ export function Post(props) {
     z,
   } = props;
 
+  const canExpand = post.preview && post.preview.images.length > 0 || !!post.oembed;
+  const inImageDomain = isImageDomain(post.cleanUrl);
+  const mixedExpActive = inMixedViewExp &&
+    (['link', 'self'].includes(post.postHint) && !inImageDomain);
```
schwers: I would make this array a `Set`, and define it once as a constant for the file. You could use `set.has` afterwards instead of `includes`

src/app/featureFlags.js
```diff
@@ -161,6 +162,15 @@ const config = {
       ] },
     ],
   },
+  [VARIANT_MIXED_VIEW]: {
+    and: [
+      { compact: false },
+      {or: [
```
schwers: tiny nit: space before `or`

src/lib/gifToHTML5Sources.js
```diff
@@ -41,6 +41,19 @@ function posterForHrefIfGiphyCat(url) {
...
+  'imgur.com', 'redd.it', 'reddituploads.com', 'flickr.com',
```
schwers: Should we add 'giphy.com' and 'gfycat.com' too? (they're in `gifToHTML5Sources`)

schwers:
💅

src/lib/gifToHTML5Sources.js
madbook: We have a standard [list of image host domains in the wiki](https://reddit.atlassian.net/wiki/display/WEB/Image+Host+Domains), we should probably use that.

src/app/components/Post/index.jsx
```diff
@@ -102,13 +106,19 @@ export function Post(props) {
     z,
   } = props;
 
+  const canExpand = post.preview && post.preview.images.length > 0 || !!post.oembed;
+  const inImageDomain = isImageDomain(post.cleanUrl);
```
madbook: nit – `inImageDomain` vs `isImageDomain` – using really similar names like this is error-prone.  Perhaps this could just be `isImage` or `isPostImage`

madbook:
💅

src/lib/gifToHTML5Sources.js
curioussavage: thanks @madbook ill take a look

src/lib/gifToHTML5Sources.js
curioussavage: It looks like this is not available in IE 10 and under. Too early for this maybe?

curioussavage: I went ahead and changed it for now (for the new file) because I saw we are already using it anyways..

curioussavage:
:haircut_man:

curioussavage:
:haircut_man: looks like I just had to squash stuff and it fixed it

schwers:
:fish:

madbook:
🐟

curioussavage:
👓  @umbrae

umbrae:
Added feedback in the ticket FYI!

🐟  presuming the event on view switching thing I brought up is figured out

PR #917, opened 2016-12-07
Fix layout for smaller screens.

telaviv:
* A larger margin is added to the top
* To accomodate this we also shrink the images a little bit to simplify the logic
* For very small phones, we also completely hide the 3rd row of images.  

The iphone 4 layout

iphone 4

The iphone 6 layout

iphone 6

👓  @prashtx @schwers


schwers:
hey @telaviv, really like the idea of using a flex column!

I'm seeing a little bit of a mismatch in the screenshots of your patch and running it locally. It looks like thumbnail grid images are being cut-off on the iPhone 4 and iPhone 5 presets:

image

image

Maybe there's a mismatch with our preset dimensions? 💅

schwers:
Note from our in-person debugging: longer subreddit names make the text content of the bottom container take up more height, cutting off the grid

telaviv:
Here's the new iphone 4 layout with smaller fonts

screen shot 2016-12-07 at 12 12 19 pm

💇

src/app/components/InterstitialListing/styles.less
```diff
@@ -164,9 +164,9 @@
     text-align: center;
     max-width: 320px;
 
-    @media (max-width: 270px) {
-      font-size: 18px;
-      line-height: 20px;
+    @media (max-width: 320px) {
```
schwers: We might want to increase this a little bit, playing around with the responsive mode in devtools it feels like there are some in-between layouts where we still get cutoff.

schwers:
Looking good for iPhones! it seems like there are still some layouts between iphone 4 and iphone 5 size ranges (i.e. some small android devices) that might still have problems. I'll optimistically :fish: to fix iPhones for now

PR #918, opened 2016-12-08
Playing state gif 

curioussavage:
[CONT-4](https://reddit.atlassian.net/browse/CONT-4)

**issue:**
This pull request resolves the issue caused by gifs playing when users go back to the previous page. example: from a comment page back to a listing page where they had previously expanded several posts. The problem is all the gifs playing at once use up a lot of data and slow things down.

**solution:**
playing state was moved out of the postContent component into redux and also decoupled from expanded state.

**bonus:**
 this lets us keep playing the gifs seamlessly when a user taps on the comments page link after playing the gif (only relevant platforms like android that play video/gifs inline).

**concerns:**
1. since moved out the only state in postContent it was expedient to convert it to a pure functional component. This was a major refactor of the component. Several pre-existing bugs were discovered while refactoring. There is the possibility of new bugs being introduced so please test the various behaviors required by the postContent component other than playing. This component may be a good place to add some ui testing in the future too.
2. the logic to decide when to play or not turned out to be a teeny bit convoluted, there could be states where we get undesired behavior still. The reducer was tested and does handle the back button and links going back to the listing correctly ( removes the playing state on all posts )

### DIFF
--- a/src/app/reducers/playingPosts.js
+++ b/src/app/reducers/playingPosts.js
@@ -1,0 +1,54 @@
+import * as postActions from 'app/actions/posts';
+import { SET_PAGE, GOTO_PAGE_INDEX } from '@r/platform/actions';
+import { removePrefix } from 'lib/eventUtils';
+
+export const DEFAULT = {};
+
+export default function(state=DEFAULT, action={}) {
+  switch (action.type) {
+    case GOTO_PAGE_INDEX: {
+      return DEFAULT;
+    }
+
+    case SET_PAGE: {
+      const postId = action.payload.urlParams.postId;
+      // Adam 12/2016: only do this if not on ios since it will autoexpand.
+      // We can remove this for ios10 and up when we are on react >= v15.3.2
+      // the navigator check is done here to avoid changing the SET_PAGE action
+      if (postId && typeof window === 'object' && window.navigator) {
+        const iOS = /iPad|iPhone|iPod/i.test(navigator.userAgent) && !window.MSStream;
+        if (state[postId] && !iOS) {
+          return { [postId]: true };
+        }
+      }
+      return DEFAULT;
+    }
+
+    case postActions.START_PLAYING: {
+      const postId = removePrefix(action.thingId);
+      const currentlyPlaying = state[postId];
+
+      if (currentlyPlaying) {
+        return state;
+      }
+
+      return {
+        ...state,
+        [postId]: true,
+      };
+    }
+
+    case postActions.STOP_PLAYING: {
+      const postId = removePrefix(action.thingId);
+      if (state[postId]) {
+        const nextState = { ...state };
+        delete nextState[postId];
+        return nextState;
+      }
+
+      return state;
+    }
+
+    default: return state;
+  }
+}
--- a/src/app/reducers/playingPosts.test.js
+++ b/src/app/reducers/playingPosts.test.js
@@ -1,0 +1,64 @@
+import createTest from '@r/platform/createTest';
+import * as platformActions from '@r/platform/actions';
+import * as postActions from 'app/actions/posts';
+
+import routes from 'app/router';
+
+import playingPosts, { DEFAULT } from './playingPosts';
+
+createTest({ reducers: { playingPosts }, routes }, ({ getStore, expect }) => {
+  describe('playingPosts', () => {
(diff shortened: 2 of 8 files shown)


src/app/reducers/playingPosts.js
```diff
@@ -0,0 +1,36 @@
...
+import { SET_PAGE } from '@r/platform/actions';
+
+const DEFAULT = {};
+export default function(state=DEFAULT, action={}) {
+  switch (action.type) {
+    case SET_PAGE: {
+      const postId = action.payload.urlParams.postId;
+      // Adam 12/2016: only do this if not on ios since it will autoexpand.
+      // We can remove this for ios10 and up when we are on react >= v15.3.2
+      var iOS = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
```
schwers: These lines might not work on the server as `navigator` and `window` won't be defined. I'm seeing some error on the server testing locally, might be this or something in PostContent (haven't tracked it down)

src/app/reducers/playingPosts.js
```diff
@@ -0,0 +1,36 @@
+import * as postActions from 'app/actions/posts';
+import { SET_PAGE } from '@r/platform/actions';
+
+const DEFAULT = {};
```
schwers: I would add a space between these lines for readability. 

We've been trying to export the default state of reducers for testing, I think this is a good reducer to test as well.

src/app/reducers/playingPosts.js
schwers: 🔕  I wish we had more control over how the apps reducer is created, if we weren't using `combineReducers` within platform you could look at (the overall state's) `state.meta.userAgent` and make this client/server agnostic

curioussavage:
👓 @schwers @nramadas

curioussavage:
👓 also @fruchtose fyi this moved everything in `postContent` around.

fruchtose:
@curioussavage Believe it or not, even with all the changes this doesn't affect #927 all that much. I can't give this a proper review though since I'm going to be traveling until Tuesday. But from a cursory glance it looks good.

curioussavage:
@fruchtose all good. Just wanted to check. I wish there was a CC like option other than eyeglasses

curioussavage:
👓  @andytuba

curioussavage:
argh.. I didn't notice those new conflicts. will rebase

fruchtose:
💅  Please solve merge conflicts introduced by my deploy today (sorry)

curioussavage:
:haircut_man:

src/app/reducers/playingPosts.js
```diff
@@ -0,0 +1,54 @@
...
+    case GOTO_PAGE_INDEX: {
+      return DEFAULT;
+    }
+
+    case SET_PAGE: {
+      const postId = action.payload.urlParams.postId;
+      // Adam 12/2016: only do this if not on ios since it will autoexpand.
+      // We can remove this for ios10 and up when we are on react >= v15.3.2
+      // the navigator check is done here to avoid changing the SET_PAGE action
+      if (postId && typeof window === 'object' && window.navigator) {
```
fruchtose: I'm not a fan of doing this detection in a reducer because it makes the reducer impure (dependency on `window`). Plus, we already have `lib/getDeviceFromState`&mdash;would it make sense to use that?

fruchtose: Also, it looks like we use `global` instead of `window` where we need to use the `window` object.

fruchtose:
🐟

src/app/reducers/playingPosts.test.js
```diff
@@ -0,0 +1,64 @@
...
+        const { playingPosts } = store.getState();
+        expect(playingPosts).to.eql(DEFAULT);
+      });
+
+      it('should leave post playing state if user is going to comment page', () => {
+        const postId = 'id23e';
+        const playingState = { [postId]: true };
+        const { store } = getStore({ playingPosts: playingState });
+
+        store.dispatch(platformActions.setPage('/', { urlParams: { postId: postId }}));
```
andytuba: Not `/comments/${postId}`?

curioussavage: Ill add that for clarity - my reducer only looks at the urlParams key

PR #919, opened 2016-12-08
Replace user activity routes with r2 equivalents.

wting:
This changes the user comment and submission paths from `/user/wting/activity?activity=comments` to `/users/wting/comments`. Visitors hitting the old activity URLs will be forwarded to the replacements, and defaulting to the user's profile page if the activity is not understood.

:eyeglasses: @uzi @schwers @prashtx


src/app/actions/activities.js
```diff
@@ -28,7 +28,7 @@ export const failed = (id, error) => ({
   error,
 });
 
-export const fetch = activiesParams => async (dispatch, getState) => {
+export const fetch = (activiesParams) => async (dispatch, getState) => {
```
uzi: Convention is not to put parenthesis when there's only one argument.

src/app/components/UserProfileHeader/index.jsx
```diff
@@ -41,13 +40,13 @@ const UserProfileTabs = props => {
...
       <UserProfileTab
-        href={ UserActivityHandler.activityURL(userName, POSTS_ACTIVITY) }
+        href={ `/user/${userName}/submitted` }
         icon='icon-posts'
         text='POSTS'
         selected={ currentActivity === POSTS_ACTIVITY }
       />
       <UserProfileTab
-        href={ UserActivityHandler.activityURL(userName, COMMENTS_ACTIVITY) }
+        href={ `/user/${userName}/comments` }
```
uzi: Why not change `UserActivityHandler.activityURL()`?  This looks to be the only place that uses it.

src/app/router/handlers/UserSubmitted.js
```diff
@@ -0,0 +1,35 @@
+import { BaseHandler, METHODS } from '@r/platform/router';
+
+import { cleanObject } from 'lib/cleanObject';
+import { SORTS } from 'app/sortValues';
+import { POSTS_ACTIVITY } from 'app/actions/activities';
+import * as activitiesActions from 'app/actions/activities';
+import { fetchUserBasedData } from './handlerCommon';
+import { listingTime } from 'lib/listingTime';
+
+export default class UserCommentsHandler extends BaseHandler {
```
uzi: You didn't rename this to UserSubmittedHandler ... but even then, this seems the same as the other with the only difference being POSTS_ACTIVITY instead of COMMENTS_ACTIVITY.  We should consolidate.

uzi:
💅  Looking good ... just a few cleanups if possible.

src/app/router/handlers/UserActivity.js
```diff
@@ -1,18 +1,11 @@
...
 
 import { cleanObject } from 'lib/cleanObject';
 import { SORTS } from 'app/sortValues';
 import { POSTS_ACTIVITY } from 'app/actions/activities';
-import * as activitiesActions from 'app/actions/activities';
-import { fetchUserBasedData } from './handlerCommon';
 import { listingTime } from 'lib/listingTime';
-import { urlWith } from 'lib/urlWith';
 
 export default class UserActivityHandler extends BaseHandler {
```
schwers: Should this be removed? Can't tell what this is doing if the `GET` is removed. It could be a useful place to store the common code but I think having plain functions would better fill that purpose

wting:
Travis CI is picking a unpinned dependency that's breaking `npm run build` (e.g. [noop change fails](https://github.com/reddit/reddit-mobile/pull/921)), which is orthogonal to the changes in this branch.

:haircut_woman:

wting:
:haircut_woman:

uzi:
:fish:

PR #920, opened 2016-12-08
Seo top 3 posts

schwers:
This re-implements the top 3 posts on comments pages for mweb with a holdout group of 1% 

See: https://github.com/reddit/reddit-private/pull/2173 for the experiment config

One issue with this is that due to how we render comments, we can't really limit the number of comments we show above the recommendations. We could either manually set some max-height and add a fade to the list of comments above the recommendations, or invest in changing our comment rendering strategy

👓  @curioussavage


schwers:
Here's a screenshot of the posts:

image

There's a lot of comments in front of the recommendations about 1200pixels in height worth. For reference this was a quick prototype of limiting the height of the first comment section based on the vertical height of the screen:

image

src/app/actions/relevantActions.js
```diff
@@ -0,0 +1,39 @@
+import { navigateToUrl } from '@r/platform/actions';
+import { METHODS } from '@r/platform/router';
+
+import {
+  trackGotoRelevantSubreddit,
+  trackGotoRelevantPost,
+} from 'lib/eventUtils';
+
+/**
+ * @param {object} - payload object with all the data needed for tracking and navigating
```
curioussavage: doesn't the name of the param go on the left of the `- <description>`?

src/app/actions/relevantActions.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {string} - payload.subredditName name of the subreddit we're navigating to
+ * @param {string} - payload.url the url we're navigating to
+ * @param {string} - payload.linkName name of the link that caused us to navigage
+ * @param {int}    - payload.linkIndex number of links in the listing that led to this navigation
+ * @param {bool}   - payload.isSelfPost bool indicating the source post is a self-post
+ * @returns {undefined} no return value, used for tracking and navigation side-effects
+ */
+export const gotoSubreddit = payload => async (dispatch, getState) => {
+  trackGotoRelevantSubreddit(getState(), payload);
```
curioussavage: should tracking be after dispatch in case it fails?

src/lib/eventUtils.js
```diff
@@ -189,3 +189,69 @@ export const logClientAdblock = (method, placementIndex, state) => {
...
 };
+
+
+const INTERNAL_CLICK_EVENTS = 'internal_click_events';
+const MWEB_RELEVANCE_CLICK_TYPE = 'cs.relevant_posts_mweb_click';
+
+const referrerPageType = isSelfPost => isSelfPost ? 'self' : 'link';
+
+/**
+ * @param {object} - payload object with all the data needed for tracking and navigating
```
curioussavage: nit: should the `state` argument be included? seems weird to only include one function arg

src/app/actions/relevantActions.js
schwers: Yup, my mistake, I'll swap them everywhere

src/app/actions/relevantActions.js
schwers: Good question, I think we may want to track the click either way and can rely on monitoring plus screenview events to track errors. I'll followup with data

src/app/pages/Comments/index.jsx
```diff
@@ -0,0 +1,316 @@
...
+    let comments;
+    let hiddenComments;
+    if (commentsPage && !commentsPage.loading) {
+      // If we have received comments, decide if we need to do comment abbreviation.
+      // We do this based on the selectors decision, which will handle abbreviating
+      // we're showing top posts. For simplicity and reselect immutability reasons
+      // we're passed a bool indicating we should abbreviate and int showing us where
+      // to do so. For SEO reasons we'll render everything after the abbreviation
+      // point as hidden comments.
+      // NOTE(schwers|Dec-25-2016): to un-hide the abbreviated comments, we won't put them back into
```
schwers: Whoops, typo turned into a gift from christmas future 🎉  

image

keithblaha:
When is this scheduled/planned for launch? We're going to be launching a big experiment showing recommendations on post comment pages monday, and something like this could contaminate the analysis for us

schwers:
I don't think this will go out  next week, but correct me if I'm wrong @curioussavage. We're waiting on re-doing comment tree rendering so it will be possible to render a specific number of comments before recommendations (moving from recursive tree to single list of comments). One thing that might be good to reconcile is that this patch and yours, @keithblaha, both show recommendations in different places.

keithblaha:
So my PR will just be for a short lived experiment, we will just be running it for 1-2 weeks before pulling it. Overall, our aim is more to test what types of recommendations perform well so we know which models to hone. Ultimately I think the mweb team itself will do the best job actually deciding where to show recommendations and integrate them well into the experience :)

curioussavage:
🏃

PR #922, opened 2016-12-09
Replace user activity routes with r2 equivalents.

wting:
This changes the user comment and submission paths from `/user/wting/activity?activity=comments` to `/users/wting/comments`. Visitors hitting the old activity URLs will be forwarded to the replacements, and defaulting to the user's profile page if the activity is not understood.


wting:
Just testing the Travis CI pipeline.

PR #923, opened 2016-12-09
AppMain.jsx refactor to use a nav frame.

telaviv:
The refactor idea came from this comment:

```
          // In Essence this is a nested router to facilitate adding UI that sits
          // above all page's content. At the moment that's TopNav and EUCookieNotice
          // In the interest of time this was implemented manually using the
          // `Case` statement. A better solution would be to have an
          //  a component that renders the App's 'frame' (TopNav, EUCookieNotice, etc)
          // and each page could manually use that component to wrap its content.
          // Another solution would be to have first class support for nested
          // routes in r/platform. Both of these will be investigated
```

This is kind of a mushing of both ideas. We have a `NavFrame` wrapper component. In addition we have a `FramedPage` that wraps stuff in the `NavFrame`. 

👓 : @schwers @prashtx @uzi


telaviv:
fixed the conflict 💇

src/app/selectors/crossPromoSelector.js
```diff
@@ -0,0 +1,42 @@
...
+  VARIANT_XPROMO_BASE,
+  VARIANT_XPROMO_LIST,
+  VARIANT_XPROMO_RATING,
+  VARIANT_XPROMO_LISTING,
+  VARIANT_XPROMO_SUBREDDIT,
+  VARIANT_XPROMO_CLICK,
+} = flagConstants;
+
+
+export function crossPromoSelector(state) {
```
schwers: Do you anticipate that we'll need to reference this selector from components other than `NavFrame`?  I think it would be more readable if this was written in `NavFrame` for now, it took a little back and forth to see where the `showX` props were coming from.

schwers:
Hey @telaviv thanks for looking into this, I really like this solution and thinks its a good example of how flexible react can be. I just have a question about the `crossPromoSelector` but otherwise think the patch is looking great and is a much needed refactor! 💅

telaviv:
💇  yeah that was the part i was least sure about. I think your right though that it's way easier to follow if they are both in the same file.  i've moved the cross promo selector in to `NavFrame`

schwers:
LGTM :fish:

PR #924, opened 2016-12-12
Change disable_comments to disableComments

jamesmaa:
👓 @schwers 

Change all instances of `disable_comments` to `disableComments`


schwers:
:fish:

PR #926, opened 2016-12-13
Update readme quickstart guide

andytuba:
👓 @uzi, @schwers


start.sh.example
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Configuration
+PORT=4444
+LIVERELOAD=true
```
schwers: should be safe to remove this

start.sh.example
```diff
@@ -0,0 +1,27 @@
...
+DEBUG_LEVEL='info'
+PROCESSES=2
+API_PASS_THROUGH_HEADERS='accept-language'
+LOGIN_PATH="/login"
```
schwers: we should remove this

start.sh.example
```diff
@@ -0,0 +1,27 @@
...
+MINIFY_ASSETS="false"
```
schwers: we should 🔪  this too, now covered by dev or prod mode

start.sh.example
```diff
@@ -0,0 +1,27 @@
...
+STATSD_DEBUG="true"
+API_PASS_THROUGH_HEADERS='accept-language'
+OAUTH_APP_ORIGIN="https://m.reddit.com"
+TRACKER_KEY="MwebTest1"
+LOGIN_PATH='/login'
```
schwers: duplicated, but should also be removed

start.sh.example
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Configuration
+PORT=4444
+DEBUG_LEVEL='info'
```
schwers: Whoops, I missed this one, we should nix this too

schwers:
Just one more quick :knife:, optimistic :fish:'ing :)

andytuba:
:haircut:

start.sh.example
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Configuration
+PORT=4444
+PROCESSES=2
+API_PASS_THROUGH_HEADERS='accept-language'
+STATSD_DEBUG="true"
+API_PASS_THROUGH_HEADERS='accept-language'
```
nramadas: this is a duplicate from line 6

start.sh.example
```diff
@@ -0,0 +1,22 @@
...
+OAUTH_APP_ORIGIN="https://m.reddit.com"
+TRACKER_KEY="MwebTest1"
```
nramadas: i would prefer to leave this an empty string as well

start.sh.example
```diff
@@ -0,0 +1,22 @@
...
+TRACKER_ENDPOINT="https://events-test.redditmedia.com/v1"
+TRACKER_CLIENT_NAME="reddit-mobile-test"
+
+# Secrets shhh
+GOOGLE_ANALYTICS_ID="UA-XXXXXX-1"
```
nramadas: and these as well :) if the keys aren't real, i think just making these empty strings will be a nice indication that they need to be filled out. otherwise, someone might mistake these as valid keys (since you have to really examine to realize that "abcd" is unlikely to be the actual key).

andytuba:
:haircut_man:

schwers:
:fish:

PR #927, opened 2016-12-13
GIF/video previews expand on click in compact mode

fruchtose:
👓  @nramadas @schwers

Previously, playable content previews in compact mode would start playing on click where the preview is. For compact mode, this means that video would play in the teeny-tiny space allotted for preview. Now, a click will expand the playable content for playback, like we do with inline images.


nramadas:
LGTM! :fish:

schwers:
:fish: !

PR #928, opened 2016-12-13
Change Post tool display

scarow:
Previously, we were using a tooltip to show
actions like save, report, etc. We are moving
towards a new design that utilizes modals
instead, to make it easier to implement mod
tools with updated design in the future.

screen shot 2016-12-13 at 2 24 51 pm
screen shot 2016-12-13 at 2 24 37 pm

https://reddit.atlassian.net/browse/CE-353

### DIFF
--- a/src/app/components/Comment/CommentDropdown/index.jsx
+++ b/src/app/components/Comment/CommentDropdown/index.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { Dropdown, DropdownRow, DropdownLinkRow } from 'app/components/Dropdown';
+import { DropdownModal, DropdownRow, DropdownLinkRow } from 'app/components/Dropdown';
 
 const T = React.PropTypes;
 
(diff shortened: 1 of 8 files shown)


scarow:
👓 @birakattack @nramadas

powerlanguage:
@scarow can we remove the horizontal lines per the designs attached to https://reddit.atlassian.net/browse/CE-353 ?

scarow:
@powerlanguage oh yeah! good eye

scarow:
For comments:
screen shot 2016-12-13 at 3 02 03 pm

scarow:
@birakattack in subsequent patches, I'll probably move this into its own component so we can dispatch API calls from the component itself.

nramadas:
this is an annoying question: should there be an explicit "close" call to action in the modal somewhere? on small screens (and especially when we add a few more options for mod actions), there might not be enough of the black overlay visible for the modal to be easily closed.

powerlanguage:
@nramadas I think this is a good suggestion, but outside the scope of the current work.  I'll file a separate ticket for this task.

birakattack:
:fish: lgtm

PR #930, opened 2016-12-14
Revert "Add back the desktop redirect on mdot servers"

wting:
This reverts commit e4b89894979b40c45f6acdc6abe0a352aa373db0.

Since the redirect cookie checking now happens at the CDN layer, this removes the redundant redirect from m.reddit.com to www.reddit.com. It doesn't solve the issue of not properly deleting the redirect cookie in r2, but helps alleviate the issue of mobile users stuck on the desktop site since they can at least use http://m.reddit.com in the interim.

:eyeglasses: @telaviv @prashtx @phil303


telaviv:
this seems right but i would feel more comfortable with one more fish
:fish:

PR #931, opened 2016-12-15
Try to replace Reddit-hosted GIFs with MP4s during preview playback

fruchtose:
👓  @nramadas @andytuba @schwers

This can save on bandwidth significantly, due to MP4s's much more efficient compression (compared to GIFs at least). We may show users smaller or lower quality content as a result, but the vast savings on bandwidth and load times present a huge upside which will be helpful to users.


src/app/components/Post/PostContent/index.jsx
```diff
@@ -73,6 +73,36 @@ function postToPlayableType(post) {
...
+ * and tries to swap it for an MP4 preview. The MP4
+ * may have smaller dimensions, but considering the
+ * filesize of a long GIF, this is worth it.
+ * @returns {Object|null}
+ */
+function gifToMp4Content(post) {
+  const url = post.cleanUrl;
+  // if the url is for a gif hosted by reddit,
+  // try to show a video in its place
+  if (post.domain === 'i.redd.it' && url.indexOf('.gif') > -1) {
```
andytuba: This feels like it belongs inside `gifToHTML5Sources`, but certainly would mean passing more properties of `post` into that function.

On a similar note, it would be nice to consolidate `gifToHtml5Sources` and `mediaUtils` into `lib/media`.

fruchtose: > This feels like it belongs inside `gifToHTML5Sources`, but certainly would mean passing more properties of `post` into that function.

For Reddit-hosted MP4s, we already need other `post` properties if we want to detect both an MP4 *and* its preview still.

> On a similar note, it would be nice to consolidate `gifToHtml5Sources` and `mediaUtils` into `lib/media`.

True. At the same time, that would significantly expand the scope of this change. That's something we could tackle after this.

src/app/components/Post/mediaUtils.js
```diff
@@ -67,6 +68,41 @@ export function findPreviewImage(isCompact, preview, thumbnail, oembed, width, n
...
+export function findPreviewVideo(preview) {
+  if (isEmpty(preview) || isEmpty(preview.images)) {
+    return null;
+  }
+
+  const img = preview.images[0];
+  // If image has both gif and mp4 variants, we must be looking at a gif.
+  // Let's choose the MP4 variant to save on bandwidth.
+  const hasVars = has(img, 'variants.gif.resolutions') && has(img, 'variants.mp4.resolutions');
+  if (!hasVars) {
```
andytuba: Could consolidate the `const foo = bar; if (!foo) return` into `if (!(bar))`.

src/app/components/Post/mediaUtils.js
```diff
@@ -67,6 +68,41 @@ export function findPreviewImage(isCompact, preview, thumbnail, oembed, width, n
...
+    return null;
+  }
+
+  const variant = img.variants.mp4;
+  if (!variant) {
```
andytuba: This looks superfluous after `hasVars` check.

fruchtose: Good catch.

src/app/components/Post/mediaUtils.js
fruchtose: I considered that, but it looked like it was hard to read because of the long condition. But if that's the preferred style, I'm fine with that.

src/app/components/Post/mediaUtils.js
```diff
@@ -67,6 +68,41 @@ export function findPreviewImage(isCompact, preview, thumbnail, oembed, width, n
...
+    return null;
+  }
+
+  if (!variant.resolutions) {
+    return variant.source;
+  }
+
+  const largestFirst = variant.resolutions.sort((a, b) => {
```
andytuba: Is it worth refactoring the "find best fit logic" from `findBestFitPreviewImage` into a `findBestFit`?

I also think the `needsNSFWBlur` logic should be handled where `findBestFitPreviewImage` is called from, potentially from a function like `findAppropriatePreviewImages`.

fruchtose: There is a slight difference in logic here. We want the largest MP4 file available, but we cannot _require_ that that it be GTE the width of the original file. So that logic cannot be reused.

I will look into the NSFW blur part. I'm not clear yet if my behavior is what already exists.

fruchtose: It looks like when a GIFV is expanded, it starts playing without showing a blur first. So I'm going to keep that behavior here as well. This code is not used for thumbnails; only for showing expanded media.

fruchtose:
💇

src/app/components/Post/mediaUtils.js
andytuba: The blur would be nice to add, but that's definitely a separate ticket.

andytuba:
:fish: and thanks for added test!

andytuba:
belay that, sorting out poster weirdness on the comments page view 💅

andytuba:
belay the belay, I had caching issues with my copy :fish:

PR #933, opened 2016-12-16
Check redirect cookie's value.

wting:
In the past we only checked the presence of the cookie rather than the value. As a result, when we zero out the cookie it still redirects users.

:eyeglasses: @andytuba @schwers @telaviv

### DIFF
--- a/src/Server.js
+++ b/src/Server.js
@@ -117,8 +117,8 @@ export function startServer() {
       },
       async (ctx, next) => {
         await next();
-        const redirectToDesktop = !!ctx.cookies.get('mweb-no-redirect');
-        if (redirectToDesktop) {
+        const desktopPreferredCookie = ctx.cookies.get('mweb-no-redirect');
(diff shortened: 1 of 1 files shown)


telaviv:
🐟

src/Server.js
```diff
@@ -117,8 +117,8 @@ export function startServer() {
       },
       async (ctx, next) => {
         await next();
-        const redirectToDesktop = !!ctx.cookies.get('mweb-no-redirect');
-        if (redirectToDesktop) {
+        const desktopPreferredCookie = ctx.cookies.get('mweb-no-redirect');
+        if (desktopPreferredCookie == 1) {
```
andytuba: How about commenting that you deliberately mean `==`, or `parseInt(desktopPreferredCookie, 10) === 1`?

wting:
:haircut_woman:

andytuba:
:fish:

schwers:
:fish:

PR #934, opened 2016-12-16
Mod modal

scarow:
https://reddit.atlassian.net/browse/CE-355
https://reddit.atlassian.net/browse/CE-443

### DIFF
--- a/src/app/actions/modTools.js
+++ b/src/app/actions/modTools.js
@@ -1,0 +1,47 @@
+import { collections, errors } from '@r/api-client';
+const { ModeratingSubreddits } = collections;
+const { ResponseError } = errors;
+
+import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+
+export const FETCHING_MODERATING_SUBREDDITS = 'FETCHING_MODERATING_SUBREDDITS';
+export const fetchingSubs = () => ({
+  type: FETCHING_MODERATING_SUBREDDITS,
+});
+
+export const RECEIVED_MODERATING_SUBREDDITS = 'RECEIVED_MODERATING_SUBREDDITS';
+export const receivedSubs = apiResponse => ({
+  type: RECEIVED_MODERATING_SUBREDDITS,
+  apiResponse,
+});
+
+export const FETCH_FAILED_MODERATING_SUBREDDITS = 'FETCH_FAILED_MODERATING_SUBREDDITS';
+export const fetchSubsFailed = error => ({
+  type: FETCH_FAILED_MODERATING_SUBREDDITS,
+  error,
+});
+
+export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+
+  const state = getState();
+  const subredditsAlreadyFetched = (
+    state.moderatingSubreddits.loading === true ||
+    state.moderatingSubreddits.responseCode === 200);
+
+  // don't make API call again
+  if (subredditsAlreadyFetched) { return; }
+
+  dispatch(fetchingSubs());
+
+  try {
+    const moderatingSubs = await ModeratingSubreddits.fetch(apiOptions);
+    dispatch(receivedSubs(moderatingSubs.apiResponse));
+  } catch (e) {
+    if (e instanceof ResponseError) {
+      dispatch(fetchSubsFailed(e));
+    } else {
+      throw e;
+    }
+  }
+};
--- a/src/app/components/Comment/index.jsx
+++ b/src/app/components/Comment/index.jsx
@@ -6,6 +6,7 @@ import { createSelector } from 'reselect';
 import { models } from '@r/api-client';
 import { Anchor } from '@r/platform/components';
 import { toggleModal } from '@r/widgets/modal';
+import includes from 'lodash/includes';
 
 import mobilify from 'lib/mobilify';
 import * as replyActions from 'app/actions/reply';
@@ -136,11 +137,14 @@ function renderTools(props) {
     commentingDisabled,
     votingDisabled,
     onToggleModal,
+    moderatingSubreddits,
   } = props;
 
   const className = cx('Comment__toolsContainer', 'clearfix', {
     'm-hidden': commentCollapsed,
   });
+
+  const isSubredditModerator = includes(moderatingSubreddits.names, comment.subreddit);
 
   return (
     <div className={ className }>
@@ -163,6 +167,7 @@ function renderTools(props) {
           commentingDisabled={ commentingDisabled }
           votingDisabled={ votingDisabled }
           onToggleModal={ onToggleModal }
+          isSubredditModerator={ isSubredditModerator }
         />
       </div>
     </div>
@@ -294,8 +299,17 @@ const selector = createSelector(
   (state, props) => !!state.collapsedComments[props.commentId],
   (state, props) => !!state.replying[props.commentId],
   (state, props) => state.editingText[props.commentId],
-
-  (user, currentPage, comment, moreCommentStatus, commentCollapsed, commentReplying, editingState) => {
+  state => state.moderatingSubreddits,
+  (
+    user,
+    currentPage,
+    comment,
+    moreCommentStatus,
+    commentCollapsed,
+    commentReplying,
+    editingState,
+    moderatingSubreddits
+  ) => {
     const editing = !!editingState;
     const editPending = editing && editingState.pending;
     const editError = editing ? editingState.error : null;
@@ -311,6 +325,7 @@ const selector = createSelector(
       editError,
       moreCommentStatus,
       highlightedComment: currentPage.urlParams.commentId,
+      moderatingSubreddits,
     };
   },
 );
--- a/src/app/components/ModeratorModal/index.jsx
+++ b/src/app/components/ModeratorModal/index.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import { Modal } from '@r/widgets/modal';
+
+const T = React.PropTypes;
+
+export function ModeratorModal(props) {
+
+  return (
+    <div className='ModeratorModalWrapper'>
+      <Modal
+        id={ props.id }
+        className='DropdownModal ModeratorModal'
+      >
+        <div onClick={ props.onClick }>
+          { props.children }
+        </div>
+      </Modal>
+    </div>
+  );
+}
+
+ModeratorModal.propTypes = {
+  id: T.string.isRequired,
+  onClick: T.func,
+};
--- a/src/app/components/Post/index.jsx
+++ b/src/app/components/Post/index.jsx
@@ -5,6 +5,7 @@ import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
 import { models } from '@r/api-client';
 import { toggleModal } from '@r/widgets/modal';
+import includes from 'lodash/includes';
 const { PostModel } = models;
 
 import * as modalActions from 'app/actions/modal';
@@ -117,6 +118,7 @@ export function Post(props) {
     winWidth,
     z,
     onToggleModal,
+    moderatingSubreddits,
   } = props;
 
   const canExpand = post.preview && post.preview.images.length > 0 || !!post.oembed;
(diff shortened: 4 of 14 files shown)


scarow:
👓  @birakattack @schwers 

This PR does two things:
1. Makes it possible to tell which subreddits a user moderates
2. Shows the `ModeratorModal` on posts and comments where a user is a moderator of the subreddit they belong to.

Right now, the `ModeratorModal` does not have any extra functionality, but it WILL have extra functionality added to it in the near future.

src/app/reducers/moderatingSubreddits.js
```diff
@@ -0,0 +1,17 @@
...
+
+const DEFAULT = {};
+
+export default function(state=DEFAULT, action={}) {
+  switch (action.type) {
+    case modToolActions.RECEIVED_MODERATING_SUBREDDITS: {
+      const moderatingSubredditIds = action.apiResponse.results.map(sr => sr.uuid);
+      // returns an array of subreddit names that the user moderates
+      // example: { moderatingSubreddits: ['pics', 'reddit'] }
+      return merge(state, moderatingSubredditIds);
```
scarow: I can change the format of this to prevent re-rendering, if necessary. This value should hardly ever change (only when someone becomes a moderator of a new sr).

scarow:
:haircut_man:

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
...
+  const apiOptions = apiOptionsFromState(getState());
+
+  try {
+    await Modtools.approve(apiOptions, postId);
+  } catch (e) {
+    throw e;
+  }
+};
+
+export const FETCHING_MODERATING_SUBREDDITS = 'FETCHING_MODERATING_SUBREDDITS';
```
birakattack: Please move all constants to the top of the file.

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
...
+export const fetchingSubs = () => ({
+  type: FETCHING_MODERATING_SUBREDDITS,
+});
+
+export const RECEIVED_MODERATING_SUBREDDITS = 'RECEIVED_MODERATING_SUBREDDITS';
```
birakattack: Move to top

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
...
+export const receivedSubs = apiResponse => ({
+  type: RECEIVED_MODERATING_SUBREDDITS,
+  apiResponse,
+});
+
+export const FETCH_FAILED_MODERATING_SUBREDDITS = 'FETCH_FAILED_MODERATING_SUBREDDITS';
```
birakattack: Move to top

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
...
+  } catch (e) {
+    throw e;
+  }
+};
+
+export const approve = (postId) => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+
+  try {
+    await Modtools.approve(apiOptions, postId);
```
birakattack: Do we not want to dispatch events before or after to record the fact that we are removing a post/comment?

Similar to how you are calling `dispatch(receivedSubs(moderatingSubs.apiResponse));` in `fetchModeratingSubreddits`?

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
...
+
+import { apiOptionsFromState } from 'lib/apiOptionsFromState';
+
+export const remove = (postId, spam) => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+
+  try {
+    await Modtools.remove(apiOptions, postId, spam);
+  } catch (e) {
+    throw e;
```
birakattack: Do we not want to handle a ResponseError if it is raised?

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+};
+
+export const approve = (postId) => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+
+  try {
+    await Modtools.approve(apiOptions, postId);
+  } catch (e) {
+    throw e;
```
birakattack: Should we handle a `ResponseError` here if it is raise like we do in `fetchModeratingSubreddits`?

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
...
+});
+
+export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+
+  dispatch(fetchingSubs());
+
+  try {
+    const moderatingSubs = await ModeratingSubreddits.fetch(apiOptions);
+    dispatch(receivedSubs(moderatingSubs.apiResponse));
```
birakattack: Can we parse values from the apiResponse here and dispatch to the appropriate action rather than tying up all of this logic in the reducer?

src/app/reducers/moderatingSubreddits.js
birakattack: Should it be called `moderatingSubredditIds` if we are returning the display name of the subreddit based on the comment?

src/app/actions/modTools.js
```diff
@@ -0,0 +1,60 @@
+import { endpoints, collections } from '@r/api-client';
+const { Modtools } = endpoints;
```
birakattack: Can we rename this to `ModTools`?

src/app/components/ModeratorModal/index.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { Modal } from '@r/widgets/modal';
+
+export function ModeratorModal(props) {
```
birakattack: Are we using this anywhere at the moment? Just want to check that we haven't forgot to check in something.

scarow:
:haircut_man:

src/app/components/Comment/index.jsx
```diff
@@ -294,8 +299,9 @@ const selector = createSelector(
   (state, props) => !!state.collapsedComments[props.commentId],
   (state, props) => !!state.replying[props.commentId],
   (state, props) => state.editingText[props.commentId],
-
-  (user, currentPage, comment, moreCommentStatus, commentCollapsed, commentReplying, editingState) => {
+  state => state.moderatingSubreddits,
+  (user, currentPage, comment, moreCommentStatus, commentCollapsed,
+                    commentReplying, editingState, moderatingSubreddits) => {
```
birakattack: Any reason for the whitespace here?

src/app/components/Post/index.jsx
```diff
@@ -132,6 +134,9 @@ export function Post(props) {
   const hasExpandedCompact = compact && expanded;
   const shouldPlay = !isPlaying && ((compact && !hasExpandedCompact) || !compact);  
   const onTogglePlaying = shouldPlay ? onStartPlaying : onStopPlaying;
+  const isSubredditModerator = includes(moderatingSubreddits, post.subreddit)
+                               ? true
+                               : false;
```
birakattack: We can simplify this to just `includes` no need for the `true|false` as `includes` returns a boolean

src/app/components/Post/index.jsx
```diff
@@ -237,8 +243,10 @@ const selector = createSelector(
   state => features.withContext({ state }).enabled(VARIANT_TITLE_EXPANDO),
   state => features.withContext({ state }).enabled(VARIANT_MIXED_VIEW),
   (state, props) => state.playingPosts[removePrefix(props.postId)],
+  state => state.moderatingSubreddits,
   (user, postId, single, compact, post, expanded, unblurred, editingState,
-   inTitleExpandoExp, inMixedViewExp, isPlaying) => {
+                inTitleExpandoExp, inMixedViewExp, isPlaying,
+                moderatingSubreddits) => {
```
birakattack: Any specific reason for the added whitespace?

src/app/router/handlers/handlerCommon.js
```diff
@@ -1,9 +1,11 @@
 import * as preferenceActions from 'app/actions/preferences';
 import * as userActions from 'app/actions/user';
+import * as modActions from 'app/actions/modTools';
 
 export const fetchUserBasedData = (dispatch) => {
   return Promise.all([
     dispatch(userActions.fetchMyUser()),
     dispatch(preferenceActions.fetch()),
+    dispatch(modActions.fetchModeratingSubreddits()),
```
birakattack: :+1:

src/app/components/Post/index.jsx
scarow: ahh. good point

scarow:
:haircut_man:

scarow:
:haircut_man:

src/app/actions/modTools.js
```diff
@@ -0,0 +1,39 @@
...
+export const FETCH_FAILED_MODERATING_SUBREDDITS = 'FETCH_FAILED_MODERATING_SUBREDDITS';
+export const fetchSubsFailed = error => ({
+  type: FETCH_FAILED_MODERATING_SUBREDDITS,
+  error,
+});
+
+export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+
+  dispatch(fetchingSubs());
```
schwers: It looks like this is getting called in `handlerCommon`, which will run on every page change. I think we should probably add a check to prevent this from from re-doing the request if its currently pending or we've already made it

src/app/components/Comment/index.jsx
```diff
@@ -6,6 +6,7 @@ import { createSelector } from 'reselect';
 import { models } from '@r/api-client';
 import { Anchor } from '@r/platform/components';
 import { toggleModal } from '@r/widgets/modal';
+import { includes } from 'lodash/collection';
```
schwers: Its better to import this via `import includes from 'lodash/includes';`

src/app/reducers/moderatingSubreddits.js
```diff
@@ -0,0 +1,17 @@
+import merge from '@r/platform/merge';
+import * as modToolActions from 'app/actions/modTools';
+
+const DEFAULT = {};
+
+export default function(state=DEFAULT, action={}) {
+  switch (action.type) {
```
schwers: There should be a case to clear this on log in / log out

src/app/reducers/moderatingSubreddits.js
```diff
@@ -0,0 +1,17 @@
...
+    case modToolActions.RECEIVED_MODERATING_SUBREDDITS: {
```
schwers: Could you add support for this action to the subreddits reducer as well? If we're getting the subreddit models back we might as well merge them in

src/app/reducers/moderatingSubreddits.js
```diff
@@ -0,0 +1,17 @@
+import merge from '@r/platform/merge';
+import * as modToolActions from 'app/actions/modTools';
+
+const DEFAULT = {};
```
schwers: Please add tests for this reducer :)

src/app/reducers/moderatingSubreddits.js
scarow: We are actually just getting a list of subreddit names here -- not full subreddit objects!

scarow:
:haircut_man:

src/app/reducers/moderatingSubreddits.test.js
```diff
@@ -0,0 +1,145 @@
...
+import * as loginActions from 'app/actions/login';
+
+
+createTest({ reducers: { moderatingSubreddits } }, ({ getStore, expect }) => {
+  describe('moderatingSubreddits', () => {
+    describe('LOGGED_IN and LOGGED_OUT', () => {
+      it('should return default on log in', () => {
+        const { store } = getStore({
+          moderatingSubreddits: {
+            'names': ['test'],
```
schwers: 🔕 small nit, you shouldn't need to add single quotes around the keys

src/app/actions/modTools.js
```diff
@@ -0,0 +1,46 @@
...
+
+export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
+  const apiOptions = apiOptionsFromState(getState());
+
+  const state = getState();
+  const subredditsAlreadyFetched = (
+    state.moderatingSubreddits.loading === true ||
+    state.moderatingSubreddits.responseCode === 200);
+
+  if (!subredditsAlreadyFetched) {
```
schwers: 🔕  Stye-wise I think its a little cleaner to bail out here. e.g. `if (subredditsAlreadyFetched) { return; }`

PR #937, opened 2016-12-20
Handle new login error types from the API

prashtx:
We're adding new login error types to the API, so we need to recognize
them and issue appropriate user messages. We should also make sure the
client can indicate unknown error types.

### DIFF
--- a/src/app/components/Login/index.jsx
+++ b/src/app/components/Login/index.jsx
@@ -6,7 +6,7 @@ import { createSelector } from 'reselect';
 import { METHODS } from '@r/platform/router';
 import { Form, Anchor, BackAnchor } from '@r/platform/components';
 
-import { loginErrors } from 'app/constants';
+import { loginErrors, genericErrors } from 'app/constants';
 
 import * as sessionActions from 'app/actions/session';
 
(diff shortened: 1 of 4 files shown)


prashtx:
👓  @birakattack @manishapme @andytuba

src/app/components/Login/index.jsx
```diff
@@ -83,8 +85,18 @@ class Login extends React.Component {
     const errorType = session ? session.error : null;
 
     const error = { username: '', password: '' };
-    if (errorType) {
+
+    if (errorType === loginErrors.WRONG_PASSWORD) {
```
andytuba: 🔕 Seems a perfect use case for `switch`

src/server/session/loginproxy.js
```diff
@@ -19,12 +21,15 @@ export default (router, apiOptions) => {
...
-      if (error === 'WRONG_PASSWORD') {
+      if (error === loginErrors.WRONG_PASSWORD ||
+          error === loginErrors.BAD_USERNAME ||
+          error === loginErrors.INCORRECT_USERNAME_PASSWORD) {
         ctx.body = { error };
-      } else { // we're not sure what this error is, log it for now
+      } else {
+        // If we don't know the error type, we still want to convey to the
+        // client that an error occurred.
+        ctx.body = { error: 'UNKNOWN_ERROR' };
```
andytuba: For consistency, how about adding a `const genericErrors = { UNKNOWN_ERROR: 'UNKNOWN_ERROR' }` to `app/constants.js` and referencing that here/elsewhere instead of a plain string?

src/app/components/Login/index.jsx
```diff
@@ -83,8 +85,18 @@ class Login extends React.Component {
...
       error.password = 'Sorry, that’s not the right password';
```
andytuba: 🔕 Out of scope for this change, but gosh i18n would be nice to have.

src/app/components/Login/index.jsx
```diff
@@ -83,8 +85,18 @@ class Login extends React.Component {
...
+    } else if (errorType === loginErrors.BAD_USERNAME) {
+      error.username = 'Sorry, that’s not a valid username';
+    } else if (errorType === loginErrors.INCORRECT_USERNAME_PASSWORD) {
+      // Use a zero-width space, so we have something truthy to trigger the
+      // clear-username button without having to show two messages.
+      error.username = '\u200B';
```
andytuba: Bit of a weird hack, how about simply passing `true` and changing [`LoginInput`'s `errorMessage`](https://github.com/prashtx/reddit-mobile/blob/7c746916c8ff27ed76f0290e90056e4f41cff7c6/src/app/components/LoginRegistrationForm/Input.jsx#L25) to verify `error` is a string?

andytuba:
💅

prashtx:
:haircut_man:

src/app/components/Login/index.jsx
```diff
@@ -83,8 +85,28 @@ class Login extends React.Component {
     const errorType = session ? session.error : null;
 
     const error = { username: '', password: '' };
-    if (errorType) {
-      error.password = 'Sorry, that’s not the right password';
+
+    switch (errorType) {
+      case loginErrors.WRONG_PASSWORD: {
+        error.password = 'Sorry, that’s not the right password';
```
andytuba: I'd prefer `\'` over `’` but I think it'll be okay for mweb.

andytuba:
:fish:

prashtx:
Merged 6d47126ea79b02a255db7ac7b328fcc74717155f

PR #938, opened 2016-12-21
Send correct referrer with events

curioussavage:
Platform will now send a value indicating when the
user navigated directly to the site. We also now handle
internal navigation by building the referrer from the
current url/domain

**Solution**
I modified the function that initializes the currentPage for the 'first' page (the second on the stack technically) which sends a string indicating there was no referrer, we can then look for that in eventUtils and distinguish between empty referrer on first load vs on single page app routing changes.

**issues**
I get errors but I think they are related to running off a local version of @r/platform. if you have a local copy of platform set up please test. Otherwise we need to get someone who has it working correctly to test.

for [CONT-5](https://reddit.atlassian.net/browse/CONT-5) together with https://github.com/reddit/node-platform/pull/30

### DIFF
--- a/src/lib/eventUtils.js
+++ b/src/lib/eventUtils.js
@@ -57,21 +57,26 @@ export function getUserInfoOrLoid(state) {
   };
 }
 
+function getDomain(referrer, meta) {
+  const x = url.parse(referrer);
+  return x.host || meta.domain;
+}
 
 export function getBasePayload(state) {
   // NOTE: this is only for usage on the client since it has references to window
-  const referrer = state.platform.currentPage.referrer || '';
+  const { platform, meta, compact, preferences } = state;
+  const referrer = platform.currentPage.referrer;
 
   const payload = {
-    domain: state.meta.domain,
-    geoip_country: state.meta.country,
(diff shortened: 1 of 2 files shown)


curioussavage:
👓  @andytuba @fruchtose

src/lib/eventUtils.js
```diff
@@ -54,17 +56,34 @@ export function getUserInfoOrLoid(state) {
   };
 }
 
+function getRealReferrer(platform, meta) {
+  const referrer = platform.currentPage.referrer;
+  if (referrer === '') {
```
curioussavage: Your are correct above though. I can just look at the history and determine if I use the last page or `referrer`. One caveat though is the fix working will be dependent on the wonky extra page in history  behavior staying the same.  Any opinions @prashtx or @andytuba ?

curioussavage:
per in person conversation. we will go with your solution, in fact I don't think we will even have to look at history once I fix the referrer in reducer.js in platform

curioussavage:
:haircut_man:

fruchtose:
🐟

PR #939, opened 2016-12-21
Hide table of contents for wikis

fruchtose:
👓  @schwers @curioussavage

This is a short term bugfix. Currently, our link router cannot handle links to anchors on the current page. This causes users who click on the table of contents to see 404 errors. Until the problem is resolved, we can hide the table of contents so users stop seeing such errors.


curioussavage:
🐟

PR #940, opened 2016-12-21
Update usage of moderator and admin color text

fruchtose:
👓  @andytuba @curioussavage 

[MOBILEWEB-598](https://reddit.atlassian.net/browse/MOBILEWEB-598)

The `@admin-text` variable was not used anywhere, so I made it useful. `@mod-text` is now defined and matches the intended color.


src/app/less/variables.less
```diff
@@ -101,7 +101,8 @@
 @dark-red: #BE1337;
 @pink-red: #FF575B;
 
-@admin-text: @red;
+@admin-text: @orangered;
```
andytuba: 🔕 It will be nice to clarify if this should be @orangered or #ff0011, but out of scope for this PR.

andytuba:
:fish:

src/app/less/variables.less
fruchtose: Yes, but at least this matches what users currently see.

PR #946, opened 2017-01-01
Chan 304 instrument interstitials

telaviv:
The goal of this PR to include two new fields for the screenview payload. One about if an interstitial has been closed and one about if an interstitial was scrolled passed. I chose 50% of the viewport to be considered a 'scroll passed event'. Not sure if this the best idea but it seems like at 50% they really meant to just get out of that interstitial. 

Justin requested simply adding the value of true to these properties. Given that, my implementation of the scroll passed storage is simply idempotently setting true whenever the event happens. On desktop these events happen very very very often. On mobile though, they happen only when a user's finger releases from the screen.

### DIFF
--- a/src/app/components/XPromoWrapper/index.jsx
+++ b/src/app/components/XPromoWrapper/index.jsx
@@ -17,11 +17,11 @@ class XPromoWrapper extends React.Component {
     // Indicate that we've displayed a crosspromotional UI, so we don't keep
     // showing them during this browsing session.
     this.props.recordXPromoShown();
-    window.addEventListener('scroll', this.onScroll);
+    window.addEventListener('scroll', this.onScroll.bind(this));
   }
 
   componentWillUnmount() {
-    window.removeEventListener('scroll', this.onScroll);
+    window.removeEventListener('scroll', this.onScroll.bind(this));
   }
 
   onScroll() {
@@ -30,7 +30,7 @@ class XPromoWrapper extends React.Component {
     // note the referencing of window
     if (window.pageYOffset > window.innerHeight / 2) {
       markXPromoScrolledPassed();
-      window.removeEventListener('scroll', this.onScroll);
+      window.removeEventListener('scroll', this.onScroll.bind(this));
     }
   }
 
--- a/src/app/components/XPromoWrapper/index.jsx
+++ b/src/app/components/XPromoWrapper/index.jsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { createStructuredSelector } from 'reselect';
 
 import * as smartBannerActions from 'app/actions/smartBanner';
-import { markXPromoScrolledPassed } from 'lib/smartBannerState';
+import { markXPromoScrolledPast } from 'lib/smartBannerState';
 
 
 const T = React.PropTypes;
@@ -26,10 +26,10 @@ class XPromoWrapper extends React.Component {
 
   onScroll() {
     // For now we will consider scrolling half the viewport
-    // "scrolling passed" the interstitial.
+    // "scrolling past" the interstitial.
     // note the referencing of window
     if (window.pageYOffset > window.innerHeight / 2) {
-      markXPromoScrolledPassed();
+      markXPromoScrolledPast();
       window.removeEventListener('scroll', this.onScroll.bind(this));
     }
   }
--- a/src/app/components/XPromoWrapper/index.jsx
+++ b/src/app/components/XPromoWrapper/index.jsx
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import { createStructuredSelector } from 'reselect';
 
 import * as smartBannerActions from 'app/actions/smartBanner';
+import { markBannerScrolledPassed } from 'lib/smartBannerState';
 
 
 const T = React.PropTypes;
@@ -16,6 +17,20 @@ class XPromoWrapper extends React.Component {
     // Indicate that we've displayed a crosspromotional UI, so we don't keep
     // showing them during this browsing session.
     this.props.recordXPromoShown();
+    window.addEventListener('scroll', this.onScroll);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.onScroll);
+  }
+
+  onScroll() {
+    // For now we will consider scrolling half the viewport
+    // "scrolling passed" the interstitial.
+    // note the referencing of window
+    if (window.pageYOffset > window.innerHeight / 2) {
+      markBannerScrolledPassed();
+    }
   }
 
   render() {
--- a/src/app/components/XPromoWrapper/index.jsx
+++ b/src/app/components/XPromoWrapper/index.jsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import { createStructuredSelector } from 'reselect';
 
 import * as smartBannerActions from 'app/actions/smartBanner';
-import { markBannerScrolledPassed } from 'lib/smartBannerState';
+import { markXPromoScrolledPassed } from 'lib/smartBannerState';
 
 
 const T = React.PropTypes;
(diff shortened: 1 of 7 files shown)


telaviv:
👓 @prashtx @schwers

src/lib/eventUtils.js
```diff
@@ -47,13 +48,25 @@ export function getUserInfoOrLoid(state) {
...
+  const loid = state.loid;
   return {
     'loid': loid.loid,
     'loid_created': loid.loidCreated,
   };
 }
 
+export function getBannerInfo() {
+  const info = {};
+  console.log('scrolled passed', BANNER_SCROLLED_PASSED);
```
schwers: small nit: need to remove this log.

We should test if localstorage is available here to prevent potential exceptions from breaking tracking.

src/lib/eventUtils.js
```diff
@@ -47,13 +48,25 @@ export function getUserInfoOrLoid(state) {
...
+  if (localStorage.getItem(BANNER_LAST_CLOSED)) {
```
prashtx: Do we need to check `localStorageAvailable()` before calling `localStorage.getItem`?

src/app/components/XPromoWrapper/index.jsx
```diff
@@ -16,6 +17,20 @@ class XPromoWrapper extends React.Component {
...
+  componentWillUnmount() {
+    window.removeEventListener('scroll', this.onScroll);
+  }
+
+  onScroll() {
+    // For now we will consider scrolling half the viewport
+    // "scrolling passed" the interstitial.
+    // note the referencing of window
+    if (window.pageYOffset > window.innerHeight / 2) {
+      markBannerScrolledPassed();
```
schwers: I think it's a good idea to remove the scroll listener here as well.

src/lib/smartBannerState.js
```diff
@@ -114,5 +117,11 @@ export function markBannerClosed() {
...
 
   // note that we dismissed the banner
-  localStorage.setItem('bannerLastClosed', new Date());
+  localStorage.setItem(BANNER_LAST_CLOSED, new Date());
+}
+
+export function markBannerScrolledPassed() {
+  if (!localStorageAvailable()) { return false; }
+
+  localStorage.setItem(BANNER_SCROLLED_PASSED, true);
```
schwers: 🔕 As it sounds like you've talked with Justin about this, but I'm a little worried that we should be firing an event here as well. I have the impression that screen views will very likely fire before this is marked in local storage, making it hard for us to tell the difference between scrolling past the xpromo and not returning, vs seeing the xpromo and not realizing they can scroll and thus not returning. Is the scrolled past flag intended to be used for measuring something else?

telaviv: @schwers I just double checked with justin and even given that limitation he still wants this to be the behavior.

src/lib/smartBannerState.js
```diff
@@ -6,6 +6,9 @@ import { getDevice, IOS_DEVICES, ANDROID } from 'lib/getDeviceFromState';
 import * as constants from 'app/constants';
 import features from 'app/featureFlags';
 
+export const BANNER_LAST_CLOSED = 'bannerLastClosed';
```
telaviv: yeah i was thinking that, but I obviously don't want to change to the field that literally goes into local storage. I think I have a good solution with just proper commentating.

src/lib/eventUtils.js
```diff
@@ -47,13 +48,25 @@ export function getUserInfoOrLoid(state) {
...
+    info.banner_closed = 'true';
```
telaviv: it's sent and interpreted as a string  ... but you're right I don't think it's a normal pattern to do this.

telaviv:
💇

src/app/components/XPromoWrapper/index.jsx
```diff
@@ -16,6 +17,21 @@ class XPromoWrapper extends React.Component {
...
+    window.removeEventListener('scroll', this.onScroll);
+  }
+
+  onScroll() {
+    // For now we will consider scrolling half the viewport
+    // "scrolling passed" the interstitial.
+    // note the referencing of window
+    if (window.pageYOffset > window.innerHeight / 2) {
+      markXPromoScrolledPassed();
+      window.removeEventListener('scroll', this.onScroll);
```
schwers: Will `onScroll` be undefined here? I think this function may need a binding

telaviv: @schwers I'm not sure why think it would be undefined. Are you suggesting that `removeEventListener` would make `onScroll` undefined?

telaviv:
💇

src/app/components/XPromoWrapper/index.jsx
prashtx: When `onScroll` is called, it may not be called with a context object, so `this` might actually be undefined. If `onScroll` is bound to the object, then we'll have access to `this` and can reference `this.onScroll` without issue.

telaviv: ah yeah woops!

telaviv:
💇

telaviv:
👓 : @uzi

src/lib/eventUtils.js
```diff
@@ -62,6 +64,19 @@ function getDomain(referrer, meta) {
...
 }
 
+export function getBannerInfo() {
+  if (!localStorageAvailable()) { return; }
+
+  const info = {};
+  if (localStorage.getItem(XPROMO_LAST_CLOSED)) {
+    info.banner_closed = true;
+  }
+  if (localStorage.getItem(XPROMO_SCROLLED_PASSED)) {
```
uzi: Should it be `XPROMO_SCROLLED_PAST`?

src/lib/eventUtils.js
```diff
@@ -62,6 +64,19 @@ function getDomain(referrer, meta) {
...
+    info.banner_scrolled_passed = true;
```
uzi: And here.

src/lib/eventUtils.js
telaviv: Yeah i've thought about this a lot and I think no. "scrolled passed" means that we scrolled passed some point. "scrolled past" implies scrolling some time in the past. Im not an english expert though.

src/lib/smartBannerState.js
```diff
@@ -110,9 +114,15 @@ export function shouldShowBanner() {
...
-  if (!localStorageAvailable()) { return false; }
+export function markXPromoClosed() {
+  if (!localStorageAvailable()) { return; }
 
   // note that we dismissed the banner
-  localStorage.setItem('bannerLastClosed', new Date());
+  localStorage.setItem(XPROMO_LAST_CLOSED, new Date());
+}
+
+export function markXPromoScrolledPassed() {
```
uzi: Silly question, but do we actually need to track this?  If they scroll past, then we don't want to show the banner anymore, right?  Wouldn't we just want to mark the timestamp of it being closed like the other case?

uzi:
💅  Asked a question about the overall patch ... we can still send the event when it's scrolled past, but then mark it as closed in the usual way, right?

telaviv:
Yeah this is what I would personally prefer. I talked to Jeremy about this and he wants to first test how many people are scrolling past the banner. There's a theory that part of the reason we have such a high install rate is because it's kind of annoying to click that "continue to mobile site" link. 

As far as "scroll passed" and "scroll past" i've thought about this a lot and I think no. "scrolled passed" means that we scrolled passed some point. "scrolled past" implies scrolling some time in the past. Im not an english expert though. I only know codez

telaviv:
💇

src/lib/eventUtils.js
uzi: http://blog.oxforddictionaries.com/2015/01/passed-past/

Look at the prepositional form.

uzi:
Hmm... but then we can just send one event when they scroll past the banner, and another when they click to close... and mark it as closed in the same way.  Why do we need to store it?

uzi:
💅

telaviv:
💇 

@uzi The feature request from justin is that we include this data with all screenviews. This means that we can't simply fire an event and be done with it. We have to include the data with all subsequent screenviews. I think long term it makes sense to have an "interstitial event" but we don't yet have plans for that.

uzi:
💅  Almost there... just a few more things that are "passed" and not "past".  I think we should be consistent.

telaviv:
💇

uzi:
:fish: Have to put this on all screenview events?! @drunken-economist is why we can't have nice things...

PR #947, opened 2017-01-03
Fix duplicate images being rendered behind gifs while playing

curioussavage:
problem: while in playing state gif posts had the image rendered behind the playing gif. Also images were being rendered with both the image tag and the css background image. 

solution: We now don't change playing state for things that are not playable. We also check to make sure we are not playing before rendering the preview image as a background image.

👓 @schwers


fruchtose:
This should also be looked over by @curioussavage since this is code he touched very recently.

schwers:
:fish:

PR #950, opened 2017-01-11
Spoilers and NSFW Refactor

birakattack:
JIRA Ticket: https://reddit.atlassian.net/browse/CE-152

As part of integrating spoilers into mweb we decided to refactor the NSFW flow to leverage the new obfuscated image previews returned by the r2 api (especially since we are looking to deprecate the `nsfw` key from the api response). Spoilers will follow the exact same pattern as NSFW images and posts on mweb.

**Screenshots**

*Listing View*
screen shot 2017-01-11 at 11 48 33 am

*Listing Expanded View*
screen shot 2017-01-11 at 11 49 21 am

*Comments Page*
screen shot 2017-01-10 at 5 58 41 pm

*NSFW and Spoiler*
screen shot 2017-01-10 at 7 06 39 pm

### DIFF
--- a/src/app/components/Post/PostContent/index.jsx
+++ b/src/app/components/Post/PostContent/index.jsx
@@ -50,6 +50,7 @@ PostContent.propTypes = {
   toggleShowNSFW: T.func.isRequired,
   togglePlaying: T.func.isRequired,
   showNSFW: T.bool.isRequired,
+  showSpoilers: T.bool.isRequired,
   editing: T.bool,
   editPending: T.bool,
   onToggleEdit: T.func.isRequired,
@@ -149,7 +150,7 @@ function renderTextEditor(rawMarkdown, props) {
 
 function buildMediaContent(post, linkDescriptor, props) {
   const { isPlaying, isThumbnail, single, width, showNSFW, onTapExpand,
-          togglePlaying } = props;
+          togglePlaying, showSpoilers } = props;
   const oembed = post.media ? post.media.oembed : null;
 
   if (isThumbnail && !(post.preview || oembed)) {
@@ -157,14 +158,15 @@ function buildMediaContent(post, linkDescriptor, props) {
   }
 
   const isNSFW = isPostNSFW(post);
-  const needsNSFWBlur = showNSFWBlur(isNSFW, showNSFW);
+  const isSpoiler = post.spoiler;
+  const needsObfuscating = showObfuscated(isNSFW, isSpoiler, showNSFW, showSpoilers);
   const playableType = postToPlayableType(post);
   const sourceURL = optimizeContent(post);
 
   let previewImage;
   if (post.preview || oembed) {
     previewImage = findPreviewImage(
-      isThumbnail, post.preview, post.thumbnail, oembed, width, needsNSFWBlur);
+      isThumbnail, post.preview, post.thumbnail, oembed, width, needsObfuscating);
   }
 
   // if we got a preview image that's a gif, convert it to a still image if possible
@@ -192,7 +194,7 @@ function buildMediaContent(post, linkDescriptor, props) {
 
   if (isThumbnail && previewImage && previewImage.url) {
     return renderImage(previewImage, sourceURL, linkDescriptor, callOnTapExpand,
-                       needsNSFWBlur, playableType, props);
+                       needsObfuscating, playableType, props);
   }
 
   // handles:
@@ -205,7 +207,7 @@ function buildMediaContent(post, linkDescriptor, props) {
       togglePlaying();
     };
     return buildImagePreview(previewImage, sourceURL, linkDescriptor,
-                             togglePlay, needsNSFWBlur, playableType, props);
+                             togglePlay, needsObfuscating, playableType, props);
   }
 
   if (oembed) {
@@ -213,18 +215,18 @@ function buildMediaContent(post, linkDescriptor, props) {
   }
 
   if (previewImage) {
-    const callback = isNSFW ? e => {
+    const callback = (isNSFW || isSpoiler) ? e => {
       e.preventDefault();
       props.toggleShowNSFW();
     } : null;
     return buildImagePreview(previewImage, sourceURL, linkDescriptor, callback,
-                             needsNSFWBlur, playableType, props);
+                             needsObfuscating, playableType, props);
   }
 }
 
 
 function buildImagePreview(previewImage, imageURL, linkDescriptor, callback,
-                           needsNSFWBlur, playableType, props) {
+                           needsObfuscating, playableType, props) {
   const html5sources = gifToHTML5Sources(imageURL, previewImage.url);
   const { single, isPlaying } = props;
 
@@ -247,7 +249,7 @@ function buildImagePreview(previewImage, imageURL, linkDescriptor, callback,
   }
 
   return renderImage(previewImage, imageURL, linkDescriptor, callback,
-    needsNSFWBlur, playableType, props);
+    needsObfuscating, playableType, props);
 }
 
 function buildMediaPreview(post, sourceURL, oembed, single) {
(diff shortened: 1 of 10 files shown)


birakattack:
:eyeglasses: @schwers @scarow

src/app/components/Post/PostHeader/styles.less
```diff
@@ -174,6 +174,23 @@
       color: @nsfw-salmon !important;
       text-decoration: none;
     }
+
+    + .PostHeader__spoiler-text {
```
scarow: wonky css here

scarow: (meaning the spacing)

src/app/components/Post/index.jsx
```diff
@@ -89,6 +91,7 @@ export function Post(props) {
   const isAndroid = userAgent && /android/i.test(userAgent);
   const showLinksInNewTab = externalDomain && isAndroid;
   const showNSFW = props.subredditIsNSFW || props.unblurred;
+  const showSpoilers = props.subredditShowSpoilers && !props.unblurred;
```
scarow: Can you explain this?

src/app/components/Post/mediaUtils.js
```diff
@@ -100,12 +100,12 @@ export function findPreviewVideo(preview) {
...
-function findBestFitPreviewImage(isThumbnail, previewImage, imageWidth, needsNSFWBlur) {
-  if (needsNSFWBlur) {
+function findBestFitPreviewImage(isThumbnail, previewImage, imageWidth, needsObfuscating) {
+  if (needsObfuscating) {
     // for logged out users and users who have the 'make safer for work'
     // option enabled there will be no nsfw variants returned.
-    if (has(previewImage, 'variants.nsfw.resolutions')) {
-      previewImage = previewImage.variants.nsfw;
+    if (has(previewImage, 'variants.obfuscated.resolutions')) {
+      previewImage = previewImage.variants.obfuscated;
```
scarow: I'm not sure I understand this -- where do `variants.obfuscated.resolutions` come from?

birakattack: We added a new image variant when we added spoilers (PR: https://github.com/reddit/reddit-public/pull/6113) to prevent us creating both a `nsfw` and `spoiler` image variant, especially since they leverage the exact same workflow. So instead we refactored the preview image key to return an `obfuscated` variant that can be used with both `nsfw` and `spoiler` posts. We hope to eventually deprecate the `nsfw` variant once third party apps have updated to the new `obfuscated` image variant.

scarow: Sounds good!

src/app/components/Post/index.jsx
birakattack: The `subredditShowSpoilers` is the subreddit preference to either display or not display the spoiler. The `unblurred` prop is used to determine whether or not the image has been revealed. So in order to show a spoiler image without obfuscation the preference has to be enabled for a subreddit and the image should not have been revealed yet.

scarow: Ah, I see. So `unblurred == True` when you have already clicked `show spoiler` on the image?

birakattack:
:haircut:

src/app/components/Post/PostContent/index.jsx
```diff
@@ -495,8 +502,8 @@ function useOurOwnPlayingOrPreviewing(oembed, url, isPlaying) {
   return (provider === 'gfycat' || provider === 'imgur' || /\.gif$/.test(url));
 }
 
-function showNSFWBlur(isNSFW, showNSFW) {
-  return isNSFW && !showNSFW;
+function showObfuscated(isNSFW, isSpoiler, showNSFW, showSpoilers) {
+  return (isNSFW && !showNSFW) || (isSpoiler && showSpoilers);
```
scarow: Why are we checking `!showNSFW` for spoilters, and `showSpoilers` (without the `!`) for spoilers?

src/app/pages/Comments.jsx
```diff
@@ -129,10 +130,11 @@ const stateProps = createSelector(
     const pageParams = CommentsPageHandler.pageParamsToCommentsPageParams(props);
     return !!state.replying[pageParams.id];
   },
+  (state, props) => state.subreddits[props.urlParams.subredditName].spoilersEnabled,
```
schwers: Is there a race condition here? I think this might try to access `spoilersEnabled` of undefined if the subreddit model hasn't loaded yet. (If it does that would throw an exception and might disrupt rendering).

schwers: Side-note: the url technically won't always have the subreddit in the name, so this might fail in that case as well. (It might be nice to have a way of updating the pageParams at the platform level for cases like this). 

Here's a quick link to how the commentsPage handles this https://github.com/reddit/reddit-mobile/blob/2X/src/app/router/handlers/CommentsPage.js#L93

src/app/components/Post/PostContent/index.jsx
```diff
@@ -464,19 +466,24 @@ function renderLinkBar(displayText, href, showLinksInNewTab, outboundLink) {
...
-      <div className='PostContent__nsfw-warning'>
-        <p className='PostContent__nsfw-warning-text'>NSFW</p>
-      </div>
-    );
+    return;
+  }
+
+  let obfuscationText;
+  if (isPostNSFW(post) && post.spoiler) {
+    obfuscationText = 'a spoiler and NSFW';
```
schwers: 🔕 Probably doesn't matter, but this is the backwards order of how the flairs show up

src/app/components/Post/PostContent/index.jsx
birakattack: This is because the semantics of `showNSFW` vs `showSpoilers` is slightly different. If a subreddit checks the option that they are a `NSFW` subreddit then we will show a gate to the subreddit before a user enters the community stating that the subreddit is 18+. However, once the user has accepted that the community is 18+ we do not show any NSFW treatment as the subreddit is already marked as 18+. If the subreddit is not marked as 18+ and a post is marked `NSFW` then we will show the  `NSFW` treatment. Whereas if spoilers are disabled for a community we will not show the spoilers at all in that community.

src/app/components/Post/PostContent/index.jsx
```diff
@@ -205,26 +207,26 @@ function buildMediaContent(post, linkDescriptor, props) {
...
 
   if (oembed) {
     return buildMediaPreview(post, sourceURL, oembed, single);
   }
 
   if (previewImage) {
-    const callback = isNSFW ? e => {
+    const callback = (isNSFW || isSpoiler) ? e => {
       e.preventDefault();
       props.toggleShowNSFW();
```
schwers: We should change NSFW -> Obfuscated for consistency

birakattack: Technically `isObfuscated` would map to whether the post is `NSFW` and the sr is not `over18` OR the post is a `spoiler` and the sr has `spoilersEnabled`. For this I do not believe that we should rename `isNSFW` to `isObfuscated` as they have different meanings.

birakattack:
:haircut:

scarow:
lgtm 🐟

src/app/pages/Comments.jsx
```diff
@@ -130,7 +131,10 @@ const stateProps = createSelector(
     const pageParams = CommentsPageHandler.pageParamsToCommentsPageParams(props);
     return !!state.replying[pageParams.id];
   },
-  (state, props) => state.subreddits[props.urlParams.subredditName].spoilersEnabled,
+  state => {
+    const sr = getSubreddit(state);
```
schwers: `getSubreddit` has a bit misleading of a name, its just returning the subreddit name as a string as opposed to the whole subreddit.
 
Naming wise, we've tried to avoid abbreviations for clarity, `subreddit` please :)

src/app/components/Post/PostContent/index.jsx
schwers: Oh, whoops, just to clarify I was thinking we want to change `toggleShowNSFW` to `toggleShowObfuscated` as it's handling NSFW and spoiler obfuscation. Sorry if I'm missing something

src/app/pages/Comments.jsx
birakattack: Whoops!

birakattack:
:haircut:

PR #953, opened 2017-01-12
Checkin yarn offline packages

schwers:
Prerequisite for using yarn, not for immediate deploy.

👓  @griffinmichl @nramadas @curioussavage


griffinmichl:
LGTM.

@schwers, I'm trying to figure out the best DX to help ensure people don't accidentally `npm install`. A call to `yarn integrity --check` will throw an error if a package was not installed correctly.

I think this could belong in either the test or lint script. Thoughts?

curioussavage:
or maybe a pre push git hook?

we could have it log a message reminding that any new deps should be added with yarn.

Can there be an accompanying commit adding to the readme that yarn is now being used?

griffinmichl:
Only thing about pre-commit hooks is that it won't let the CI catch errors.

I guess as long as the CI runs w/ the `yarn --offline` flag that would also catch improperly installed packages.

schwers:
I think a pre-commit hook and adding it to the test script would be good.
That should make it easy to catch early in dev, and prevent things from
breaking on deploy.

schwers:
Replaced with https://github.com/reddit/reddit-mobile/pull/1073

PR #955, opened 2017-01-13
Add tracking pixel

madbook:
for [CE-391](https://reddit.atlassian.net/browse/CE-391)

This tracking pixel was accidentally during the transition from mweb classic (1x) to 2x.  [Here's the original implementation from the 1x branch](https://github.com/reddit/reddit-mobile/blob/ecf7b3a45a11e48139f88e392bfd9ebae19cf2ac/src/lib/TrackingPixel.js), and [here's where it got used in 1x](https://github.com/reddit/reddit-mobile/blob/master/src/views/pages/BasePage.jsx).  The way it generally works:

1. Some api requests (generally GET requests that correspond to traditional page loads, like getting a listing) include a response header containing the tracking pixel URL, with verification param attached.
2. We add some additional contextual info from the clientside as url params.
3. We create a dummy `Image` and assign the modified URL as it's href, trigging the request.

This is for the most part a direct port of that implementation into 2x.  The main differences are:

1. `TrackingPixel` is implemented as a react component, located in the `side-effect-components` directory and loaded with the other side-effect components.
2. The current tracking pixel is added to the state tree with a new reducer.
3. I've dropped support for including experiments in the URL.  In the 1x branch, it looks like there is support for passing in feature flag states and having them included in the URL, but it wasn't actually being used anywhere.  🔪 

**Example url**

```
https://pixel.redditmedia.com/pixel/of_destiny.png?v=%2BZnsE%2F%2BFa9CctRvQGRnrpgKCKMPyyqw7ft9PfARrWMp%2B2V57Tx3irWInLi5eBixh2%2Fmv5TUD2MU%3D&r=0.8323848260048954&loid=dguOn3yHomm6WWF47U&loidcreated=1483571120000&view_type=compact&dnt=false
```

👓 @birakattack @schwers


src/app/side-effect-components/TrackingPixel.jsx
```diff
@@ -0,0 +1,100 @@
...
+);
+
+class TrackingPixel extends React.Component {
+  componentDidMount() {
+    this.fire();
+  }
+
+  shouldComponentUpdate(nextProps) {
+    // prevent re-firing if platform updates but we don't get a new tracking pixel
+    return nextProps.pixelTrackerUrl !== this.props.pixelTrackerUrl;
```
birakattack: Oh nvm looks like in the `fire` method we just return so this would not cause and issue.

src/app/side-effect-components/TrackingPixel.jsx
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { createSelector } from 'reselect';
+import url from 'url';
+
+const T = React.PropTypes;
+
+const selector = createSelector(
```
birakattack: Can we move the selector to the bottom of the page to match the style of other components?

src/app/side-effect-components/TrackingPixel.jsx
```diff
@@ -0,0 +1,100 @@
...
+  ({compact}) => compact,
+  ({user}) => user,
+  ({loid}) => loid,
+  ({platform}) => platform,
+  ({tracking}) => tracking,
```
madbook: yeah, it's the same as

```
(state) => state.tracking,
```

though looking back now, maybe we aren't using this style anywhere in `reddit-mobile`.  Probably worth switching to match existing styles

madbook:
Addressed those style concerns, and also realized that the old 1x TrackingPixel class was still in there :O 

👓 @birakattack

src/app/reducers/tracking.js
```diff
@@ -0,0 +1,31 @@
...
+    case commentsPageActions.RECEIVED_COMMENTS_PAGE:
+    case mailActions.RECEIVED:
+    case postsListActions.RECEIVED_POSTS_LIST:
+    case savedActions.RECEIVED_SAVED:
+    case searchActions.RECEIVED_SEARCH_REQUEST:
+    case similarPostsActions.RECEIVED_SIMILAR_POSTS:
+    case wikiActions.RECEIVED_WIKI: {
+      const pixel = action.apiResponse.response.headers['x-reddit-tracking'] || state.pixel;
+      return {
+        ...state,
```
schwers: This line seems unnecessary, the only key will always be 'pixel' and we're pulling it from state anyway?

madbook:
8bd0d4285392a7f0db4e68dc66b593df342262f2

src/app/reducers/tracking.js
madbook: whoops, I missed this comment somehow.  Yeah, probably could have removed that

PR #956, opened 2017-01-16
Remove unused 'count' param from urls

alsuren:
This makes it easier to glance up and see how much time you've
wasted browsing /r/all, as you can now see the page number in the
URL bar.

I had a grep for count and it doesn't look like anything actually
honours it.  On the desktop site, it is used to add numbers to the
left-hand-side, and is incremented whenever you page forwards or
back. On the mobile version, you have no number beside the posts,
and the 'count' param stays pointlessly set to whatever you type in
(including non-numeric values).

### DIFF
--- a/src/app/components/PaginationButtons/index.jsx
+++ b/src/app/components/PaginationButtons/index.jsx
@@ -19,7 +19,6 @@ export const buildPrevUrl = (records, pagingPrefix='', currentQueryParams={}) =>
 
   if (page > 0) {
     const query = {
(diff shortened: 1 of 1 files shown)


alsuren:
My original PR (https://github.com/reddit/reddit-mobile/pull/703) just moved the 'count' param to the right in the query parameters, but the way the current code uses the spread operator makes it more difficult to move. I had a look, and it turns nothing cares about the param, so it's easier to simply remove it.

PRs with negative diffstat are the best kind of PRs, right?

PR #959, opened 2017-01-19
Include experiment owner in bucketing event if present

madbook:
Another piece of **[CE-463](https://reddit.atlassian.net/browse/CE-463)**

If the experiment data includes an `owner` field, include that in the `bucketing_events`

👓 @prashtx @birakattack

### DIFF
--- a/src/app/featureFlags.js
+++ b/src/app/featureFlags.js
@@ -347,7 +347,7 @@ flags.addRule('variant', function (name) {
   const user = extractUser(this);
 
   if (user && user.features && user.features[experiment_name]) {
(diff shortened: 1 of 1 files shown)


prashtx:
Do we need to add `owner` to the `cs.experiment_click` events, too, or just bucketing events?

madbook:
@prashtx Good question, I'm not sure.  Technically the scope of the work is just adding to bucketing events, so I think this could proceed even if we also want to add them to that event as well.

madbook:
1f9b8d7dfa7bde1e6cc3d1cf73ae0e925f4d1593

PR #944, opened 2016-12-23
Nightmode fixes

ajacksified:
All the interstitial changes that went out Wednesday broke a bunch of stuff; here's fixes for the things that were annoying me.

**Before**

screen shot 2016-12-23 at 9 14 53 am

**After**

screen shot 2016-12-23 at 9 17 27 am

btw: NSFW images are still broken because of #931.

Also, that interstitial is the worst thing I have literally ever seen. I'd feel bad for the person who had to write the code to say "my work sucks, use this instead," but good choice having one of the non-core devs do it, eh? But now you've got bugs. _tradeoffs_

PR #962, opened 2017-01-23
Fix PR #944

ajacksified:
Fixes a typo on pr #944, which turns the background instead of the text color. Whoops, sorry >.<

👓  @fruchtose


fruchtose:
Thanks for the fix. I noticed the bug and aborted deploying it.

ajacksified:
👍

ajacksified:
While I've got your ear, @fruchtose, do you know if the non-sticky header is intentional or if it's also a bug? I know it was changed at the same time as the new interstitials, but wasn't sure if it was to fix a bug with interstitials or removed for other reasons. I miss it, because navigating between subs is just a little more annoying now; I typically want to switch subreddits at the bottom of a page, which means hiking back up to the top.

fruchtose:
Hmmm, I'll look into it.

fruchtose:
@ajacksified unfortunately I found out that the sticky nav isn't ready at this point due to what I believe is an interstitial compatibility problem.

ajacksified:
Figured as much. Thanks for looking into it!

PR #965, opened 2017-01-23
Chan 360 require login to skip interstitial

telaviv:
I recommend reading this PR commit by commit.

1st commit - Remove a bunch of the dead code from past xpromo experiments.

2nd commit - The actual feature. The basic idea being, that you are required to log in before actually viewing the content on mweb. 

You'll notice that there are some css changes to the login page. This is a slight refresh by Jonathan and is intentional. 

You'll also see that the behavior for after you login has changed. In the past we sent you to the home page. I've changed it to return you to the same location as if you had clicked the 'x' at the top left. I think this is a nice quality of life change for everyone.

👓 : @schwers @prashtx


src/app/components/Login/index.jsx
```diff
@@ -77,8 +81,25 @@ class Login extends React.Component {
     );
   }
 
+  onAppPromoClick() {
```
schwers: You can bind this to the component by writing this as `onAppPromoClick = () => { ... }`, and avoid callind bind in `renderAppPromo`

src/app/components/LoginRegistrationForm/styles.less
```diff
@@ -79,10 +86,58 @@
...
+    font-size: 1.4em;
+    font-weight: bold;
+    line-height: 2 * @grid-size;
+    margin: @grid-size auto;
+
+    @media (max-height: 600px) {
+      margin: 2 * @grid-size 0;
+    }
+
+    .SquareButton {
```
schwers: :nit: In the spirit of keeping things BEM, I think it would be better to pass a modifier class to the button component instead of relying on nesting

src/app/components/LoginRegistrationForm/styles.less
```diff
@@ -79,10 +86,58 @@
...
+    }
+  }
+
   &__terms {
     color: @grey-text;
     text-align: center;
     margin: 10px auto 0;
     max-width: 450px;
   }
+
```
schwers: :nit: don't need the extra newline

src/app/components/LoginRegistrationForm/styles.less
```diff
@@ -79,10 +86,58 @@
...
+    @media (max-height: 600px) {
+      margin: 2 * @grid-size 0;
+    }
+
+    .SquareButton {
+      background-color: @orangered;
+    }
+
+    .or {
```
schwers: :nit: I would pull `.or` and `.subtext` out of being nested under `&__app-promo`, and renamed to something like `.Login__appPromo-or` (`&__appPromo-or`) to keep things BEM.

src/app/components/Post/index.jsx
```diff
@@ -297,7 +296,6 @@ const mapDispatchToProps = (dispatch, { postId }) => ({
   onStopPlaying: () => dispatch(postActions.stopPlaying(postId)),
   onStartPlaying: () => dispatch(postActions.startPlaying(postId)),
   onReportPost: () => dispatch(reportingActions.report(postId)),
-  onElementClick: () => dispatch(modalActions.showXpromoModal()),
```
schwers: `onElementClick` still needs to be removed from lines 127, 213, and 234. That should fix the error if you click the title or comments link (the error is breaking the prevent default and causing a real page navigation)

src/app/components/LoginRegistrationForm/styles.less
telaviv: ah I'm more used to the `.Login__app-promo--or` style of bem.

schwers: I should probably say 'bem-like' :) I think your example is good too

telaviv:
💇

telaviv:
👓 : @prashtx. 3 more feature requests. Just look at the last two commits.

1) instead of saying `or login to the mobile site` it says `or login`
2) We only serve the interstitial to iphone + android. no more ipad.
3) separate experiments for ios and android.

prashtx:
Last two commits look good. :fish: for those.

telaviv:
💇

prashtx:
:fish: assuming the build passes

telaviv:
💇

prashtx:
:fish:

src/app/components/DualPartInterstitial/Footer/index.jsx
```diff
@@ -3,109 +3,155 @@ import './styles.less';
...
+    utm_medium: medium,
+  });
+}
 
 const selector = createSelector(
   getSubreddit,
-  interstitialExperimentSelector,
-  (subredditName, experiments) => {
-    return { subredditName, ...experiments };
+  (state) => { return features.withContext({ state }); },
```
schwers: :nit: It would be better to return the `campaign` string you're computing below. `feature.withContext` will always return a new object, which won't `===` compare to your last return value. strings will pass `===` if they're the same . You could still compute requireLogin` with the `campaign` string:

```js
const requireLogin = (loggedOut && (
  campaign === 'xpromo_fp_login_required' || 
  campaign === 'xpromo_subreddit_login_required'
));
```

PR #966, opened 2017-01-25
Fix modTool apiResponse status bug

scarow:
For users who are moderators on >100
subreddits, the fetch pagination returns
two apiResponses, which was causing a bug
when attempting to get apiRespose.status.
Instead, we will now only capture errors,
as we don't really need the apiResponse code

### DIFF
--- a/src/app/actions/modTools.js
+++ b/src/app/actions/modTools.js
@@ -107,13 +107,12 @@ export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
 
   const subredditsAlreadyFetched = (
     state.moderatingSubreddits.loading === true ||
-    state.moderatingSubreddits.responseCode === 200);
+    !!state.moderatingSubreddits.names);
 
   // don't make API call again
   if (subredditsAlreadyFetched) { return; }
 
   dispatch(fetchingSubs());
-
   try {
     const moderatingSubs = await ModeratingSubreddits.fetch(apiOptions);
     dispatch(receivedSubs(moderatingSubs.apiResponse));
--- a/src/app/reducers/moderatingSubreddits.js
+++ b/src/app/reducers/moderatingSubreddits.js
@@ -4,8 +4,8 @@ import * as loginActions from 'app/actions/login';
 
 const DEFAULT = {
   loading: false,
-  responseCode: null,
-  names: [],
+  error: null,
+  names: null,
 };
 
 export default function(state=DEFAULT, action={}) {
@@ -13,8 +13,8 @@ export default function(state=DEFAULT, action={}) {
     case modToolActions.FETCHING_MODERATING_SUBREDDITS: {
       const moderatingSubreddits = {
         loading: true,
-        responseCode: null,
-        names: [],
+        names: null,
+        error: null,
       };
       return merge(state, moderatingSubreddits);
     }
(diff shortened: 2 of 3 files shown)


scarow:
👓 @schwers @birakattack 

@powerlanguage for visibility

src/app/actions/modTools.js
```diff
@@ -107,13 +107,11 @@ export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
 
   const subredditsAlreadyFetched = (
     state.moderatingSubreddits.loading === true ||
-    state.moderatingSubreddits.responseCode === 200);
+    state.moderatingSubreddits.names);
 
   // don't make API call again
   if (subredditsAlreadyFetched) { return; }
-
```
schwers: 🔕 Style-wise, I think its a little easier to read if we keep this whitespace

src/app/actions/modTools.js
```diff
@@ -107,13 +107,11 @@ export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
 
   const subredditsAlreadyFetched = (
     state.moderatingSubreddits.loading === true ||
-    state.moderatingSubreddits.responseCode === 200);
+    state.moderatingSubreddits.names);
```
schwers: :nit:  '!!' in front of this please (i.e. '!!state.moderatingSubreddits.names'). It will be equivalent in terms of 'truthiness' but it will ensure  'subredditsAlreadyFetched' is always a boolean.

src/app/reducers/moderatingSubreddits.js
```diff
@@ -24,16 +24,16 @@ export default function(state=DEFAULT, action={}) {
...
       };
       return merge(state, moderatingSubreddits);
     }
 
     case modToolActions.FETCH_FAILED_MODERATING_SUBREDDITS: {
       const moderatingSubreddits = {
         loading: false,
-        responseCode: action.error.status,
+        error: action.error,
         names: [],
```
schwers: I think this should be 'null' now that it's the default value for 'names'.  'subredditsAlreadyFetched' will be true in the fetch call, so it won't re-request if the last api call failed. In some ways that's good as it could stop us from re-requesting against that endpoint when it's in a broken state. However, I don't think that behavior is immediately obvious reading 'subredditsAlreadyFetched'. If we want it to still do that (not fetch if the last call was an error), it would be better if we explicitly checked (we could add '|| !!state.moderatingSubreddits.errors')

scarow: Hmm, sorry -- I'm not sure I understand.

``const subredditsAlreadyFetched = (
    state.moderatingSubreddits.loading === true ||
    !!state.moderatingSubreddits.names);``

`subredditsAlreadyFetched` will return `false` in error case (if we change `names: null`) because `loading == false` and `!!states.moderatingSubreddits.names == false`. Right?

scarow:
:haircut_man:

scarow:
:haircut_man:

scarow:
:haircut_man:

scarow:
:haircut_man:

src/app/reducers/moderatingSubreddits.js
schwers: Right! I think its fine either way. My old comment, re-worded is basically: when I read the `subredditsAlreadyFetched` check, I thought it would be true if we were currently making a request or if if we had received results of some kind. I thought if there was an error, that `subredditsAlreadyFetched` would be false just because there's no mention of errors in the conditional. We have tests documenting it so its pretty clear, 🔕

PR #967, opened 2017-01-25
Add `userId` data layer variable

dwick:
👓 @schwers 

related ticket: https://reddit.atlassian.net/browse/ADS-1245

### DIFF
--- a/src/lib/eventUtils.js
+++ b/src/lib/eventUtils.js
@@ -32,14 +32,20 @@ export function convertId(id) {
   return parseInt(removePrefix(id), 36);
 }
 
-
-export function buildSubredditData(state) {
+function getSubredditFromState(state) {
   const { subredditName } = state.platform.currentPage.urlParams;
 
-  if (!subredditName || isFakeSubreddit(subredditName)) { return {}; }
+  if (subredditName && !isFakeSubreddit(subredditName)) {
+    return state.subreddits[subredditName.toLowerCase()];
+  }
+}
 
-  const subreddit = state.subreddits[subredditName.toLowerCase()];
-  if (!subreddit) { return {}; }
+export function buildSubredditData(state) {
+  const subreddit = getSubredditFromState(state);
+
+  if (!subreddit) {
+    return {};
+  }
 
   return {
     sr_id: convertId(subreddit.name),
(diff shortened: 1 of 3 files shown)


src/lib/eventUtils.js
```diff
@@ -172,14 +172,15 @@ export function trackPreferenceEvent(state, additionalEventData={}) {
 
 const gtmPageView = state => {
   const { platform: { currentPage }} = state;
+  const userInfo = getUserInfoOrLoid(state);
 
   gtm.trigger('pageview', {
     subreddit: currentPage.urlParams.subredditName || '',
     pathname: currentPage.url || '/',
+    userId: userInfo.user_id || '',
```
dwick: added to `pageview` because a user may come in logged out and then log in leaving `userId` unset from `initGoogleTagManager`

src/lib/gtm.js
```diff
@@ -22,7 +22,7 @@ export const initGoogleTagManager = initialSubredditName => {
   gtmJail.style.display = 'none';
   gtmJail.id = GTM_JAIL_ID;
   gtmJail.name = JSON.stringify({
-    subreddit: initialSubredditName || '',
```
dwick: i checked and all of the tags which utilize this check for `undefined` so no need to default to `''`

src/lib/eventUtils.js
schwers: This is a tiny nit, but if we're defaulting to empty-string here, should we be doing that in the call to initGoogleTagManager?

dwick:
shouldn't need the defaults, so i removed them. i'll do some more testing before merging this too.

dwick:
💇

dwick:
💇 added `advertiserCategory` and removed custom variables from `initGoogleTagManager` since all tags are fired when `pageview` is triggered.

src/lib/eventUtils.js
```diff
@@ -171,15 +171,18 @@ export function trackPreferenceEvent(state, additionalEventData={}) {
 }
 
 const gtmPageView = state => {
-  const { platform: { currentPage }} = state;
+  const { platform: { currentPage: { url, urlParams: { subredditName } } } } = state;
+  const subreddit = state.subreddits[subredditName];
```
schwers: This should use `subredditName.toLowerCase()`

src/lib/eventUtils.js
```diff
@@ -171,15 +171,18 @@ export function trackPreferenceEvent(state, additionalEventData={}) {
...
+  const userInfo = getUserInfoOrLoid(state);
 
   gtm.trigger('pageview', {
-    subreddit: currentPage.urlParams.subredditName || '',
-    pathname: currentPage.url || '/',
+    userId: userInfo.user_id,
+    subreddit: subredditName,
```
schwers: 🔕  This might need to be `toLowerCase()`'d as well

dwick:
💇

PR #968, opened 2017-01-27
Add distinguish action to posts

scarow:
comments:
https://reddit.atlassian.net/browse/CE-384
https://reddit.atlassian.net/browse/CE-374

posts:
https://reddit.atlassian.net/browse/CE-368
https://reddit.atlassian.net/browse/CE-379


scarow:
👓 @birakattack @madbook 

Notes: 
1. The `distinguish icon` is borked --  ignore for now. This should be an easy fix, just want to get the logic review out of the way.
2. You must pull & link to `node-api-client` https://github.com/reddit/node-api-client/pull/188

scarow:
toatssitst_-_toasties_test

src/app/actions/modTools.js
```diff
@@ -101,6 +123,27 @@ export const approve = id => async (dispatch, getState) => {
...
+  const apiOptions = apiOptionsFromState(state);
+  const model = modelFromThingId(id, state);
+
+  dispatch(distinguishPending());
+
+  try {
+    await Modtools.distinguish(apiOptions, id, how);
+    // DISTINGUISH_TYPE[how] will return either 'moderator' or '', which
+    // is added to model.distinguished state
+    dispatch(distinguishSuccess(model, DISTINGUISH_TYPE[how]));
```
madbook: I think this is a little confusing – `distinguish` here has an arg `how` that is expected to be the string `yes` or `no`, while the action creator it calls (`distinguishSuccess`) also has an argument called `how` that expects to either be the string `moderator` or an empty string.  I think we could drop a little complexity by either:

1. replacing the yes/no strings with booleans, changing the argument to this method to something like `isDistinguished` and choosing which string to pass along to `distinguishSuccess` locally, or
2. dropping the yes/no strings and just passing either `moderator` or an empty string directly into this method.

src/app/components/Comment/CommentDropdown/index.jsx
```diff
@@ -93,6 +97,8 @@ CommentDropdown.propTypes = {
   removedBy: T.string,
   showModModal: T.bool,
   modModalId: T.string,
+  isDistinguished: T.string,
```
madbook: related to my above comment, it's a bit unexpected for a property name starting with `is` to be something other than a bool.  Maybe something like `distinguishType` for a string? (or keep it as is and change the type to bool)

src/app/components/ModeratorModal/index.jsx
```diff
@@ -9,8 +10,20 @@ import { getStatusBy, getApprovalStatus } from 'lib/modToolHelpers.js';
 import * as modActions from 'app/actions/modTools';
 
 const T = React.PropTypes;
+const DISTINGUISH_FALSEY_VALS = ['', 'no', undefined, null];
 
 export class ModeratorModal extends React.Component {
+
+  onDistinguish(isDistinguished) {
+    const how = includes(DISTINGUISH_FALSEY_VALS, isDistinguished) ? 'yes' : 'no';
```
madbook: I think some of this can also go away RE above comments.

scarow:
With separation between toggle/non-toggle items:

screen shot 2017-01-27 at 11 15 56 am

scarow:
:haircut_man: see new node-api-client changes as well

src/app/components/ModeratorModal/index.jsx
```diff
@@ -31,24 +49,35 @@ export class ModeratorModal extends React.Component {
...
-                onClick={ this.props.onSpam }
-                isSelected={ this.props.isSpam }
-              />
-              <DropdownRow
-                icon='check-circled'
-                text='Approve'
-                onClick={ this.props.onApprove }
-                isSelected={ this.props.isApproved }
-              />
+              { this.props.isMine
```
birakattack: Why is this needed if it is only shown in the `ModeratorModal`?

src/app/actions/modTools.js
```diff
@@ -47,6 +50,20 @@ export const approvalSuccess = (thing, username) => ({
   username,
 });
 
+export const distinguishPending = () => ({
+  type: MODTOOLS_DISTINGUISH_PENDING,
+});
+
+export const distinguishError = () => ({
```
birakattack: The action payload for this should be 
```
{
    type: ...,
    error,
    message: 'Some message goes here'
}
```

We can then listen for this action in `app/reducer/toaster.js` to emit a toast that contains the message passed in the above payload.

src/app/actions/modTools.js
```diff
@@ -47,6 +50,20 @@ export const approvalSuccess = (thing, username) => ({
   username,
 });
 
+export const distinguishPending = () => ({
```
birakattack: We can probably just get rid of this, since this is not doing anything.

src/app/actions/modTools.js
```diff
@@ -12,6 +12,9 @@ export const MODTOOLS_REMOVAL_SUCCESS = 'MODTOOLS_REMOVAL_SUCCESS';
 export const MODTOOLS_APPROVAL_PENDING = 'MODTOOLS_APPROVAL_PENDING';
 export const MODTOOLS_APPROVAL_ERROR = 'MODTOOLS_APPROVAL_ERROR';
 export const MODTOOLS_APPROVAL_SUCCESS = 'MODTOOLS_APPROVAL_SUCCESS';
+export const MODTOOLS_DISTINGUISH_PENDING = 'MODTOOLS_DISTINGUISH_PENDING'
```
birakattack: We can remove this since it's not really being used.

src/app/actions/modTools.js
```diff
@@ -101,6 +118,25 @@ export const approve = id => async (dispatch, getState) => {
   }
 };
 
+export const distinguish = (id, distinguishType) => async (dispatch, getState) => {
+  const state = getState();
+  const apiOptions = apiOptionsFromState(state);
+  const model = modelFromThingId(id, state);
+
+  dispatch(distinguishPending());
```
birakattack: We can remove this as well.

src/app/components/ModeratorModal/index.jsx
```diff
@@ -69,12 +98,15 @@ ModeratorModal.propTypes = {
   isSpam: T.bool.isRequired,
   removedBy: T.string,
   approvedBy: T.string,
+  isDistinguished: T.string,
+  isMine: T.bool,
```
birakattack: We can remove this.

src/app/components/Post/PostDropdown/index.jsx
```diff
@@ -35,6 +35,8 @@ export default function PostDropdown(props) {
     removedBy,
     showModModal,
     modModalId,
+    isDistinguished,
+    isMine,
```
birakattack: Just adding a note to remove.

src/app/components/Post/PostDropdown/index.jsx
```diff
@@ -60,6 +62,8 @@ export default function PostDropdown(props) {
         isSpam={ isSpam }
         approvedBy={ approvedBy }
         removedBy={ removedBy }
+        isMine={ isMine }
```
birakattack: Remove.

src/app/components/Post/PostFooter/index.jsx
```diff
@@ -112,6 +112,8 @@ export default class PostFooter extends React.Component {
         removedBy={ post.bannedBy }
         showModModal={ showModModal }
         modModalId={ modModalId }
+        isDistinguished={ post.distinguished }
+        isMine={ user && user.name === post.author }
```
birakattack: Remove.

src/app/components/ModeratorModal/index.jsx
```diff
@@ -10,14 +10,19 @@ import { getStatusBy, getApprovalStatus } from 'lib/modToolHelpers.js';
 import * as modActions from 'app/actions/modTools';
 
 const T = React.PropTypes;
-const DISTINGUISH_FALSEY_VALS = ['', 'no', undefined, null];
+const DISTINGUISH_FALSEY_VALS = ['', undefined, null];
```
madbook: You shouldn't need to do this anymore – since all of these values are already falsey, you should be able to just do

```javascript
const distinguishType = !isDistinguished ? DISTINGUISH_TYPES.MODERATOR : DISTINGUISH_TYPES.NONE;
```

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -215,10 +215,10 @@ function renderApprovalStatusFlair(post) {
 
   return (
     <span className='PostHeader__approval-status-flair'>
+      { distinguished === 'moderator' ? MOD_FLAIR : null }
```
madbook: can this use `DISTINGUISH_TYPES.MODERATOR` instead of the hard-coded string?

src/app/actions/modTools.js
```diff
@@ -63,10 +58,10 @@ export const distinguishError = () => ({
   type: MODTOOLS_DISTINGUISH_ERROR,
 });
 
-export const distinguishSuccess = (thing, how) => ({
+export const distinguishSuccess = (thing, distinguishType) => ({
```
madbook: 👍 👏

src/app/components/ModeratorModal/index.jsx
scarow: `isMine` refers to a post or comment that is yours (not a subreddit). As in, I can only distinguish a post or comment that i made.

src/app/components/ModeratorModal/index.jsx
scarow: Why?

src/app/components/Post/PostFooter/index.jsx
scarow: Mentioned above -- we need `isMine`

scarow:
:haircut_man: thanks guys!

src/app/components/ModeratorModal/index.jsx
```diff
@@ -9,8 +10,25 @@ import { getStatusBy, getApprovalStatus } from 'lib/modToolHelpers.js';
 import * as modActions from 'app/actions/modTools';
 
 const T = React.PropTypes;
+const DISTINGUISH_FALSEY_VALS = ['', undefined, null];
```
madbook: Is there a reason to leave this in?

madbook:
whoops, 💅 for harold update and slack ping

src/app/components/ModeratorModal/index.jsx
scarow: ehhh I suppose not. I guess I could just check which value it currently is, then toggle it

scarow:
:haircut_man:

src/app/components/ModeratorModal/index.jsx
birakattack: Ahh got it sorry for the noise on that

birakattack: Would it be better to move this into a function like `renderDistinguishButton(props)` and pass in the props to render the button. That way in the main render method we would just call `{ renderDistinguishButton(this.props) }` instead of the largish ternary operator?

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -202,16 +201,24 @@ function renderPostFlair(post, single) {
   );
 }
 
-function renderApprovalStatusFlair(isApproved, isRemoved, isSpam) {
-  if (!(isApproved || isRemoved || isSpam)) {
+function renderApprovalStatusFlair(post) {
```
birakattack: This should be renamed if it is being overloaded to items other than approve, remove and spam. Maybe something like `renderPostIcons`?

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -275,14 +279,13 @@ function renderPostDescriptor(
     authorOrNil,
     !hideSubredditLabel && flairOrNil,
   ]);
-  const approvalStatusFlair = renderApprovalStatusFlair(post.approved, post.removed, post.spam);
+  const approvalStatusFlair = renderApprovalStatusFlair(post);
```
birakattack: Should be renamed here as well.

src/app/components/Post/PostHeader/index.jsx
birakattack: Actually this will be fixed with @madbook's PR so we can just ignore this naming change for now as it will happen in subsequent PRs

src/app/actions/modTools.js
```diff
@@ -101,6 +115,27 @@ export const approve = id => async (dispatch, getState) => {
...
 };
 
+export const distinguish = (id, distinguishType) => async (dispatch, getState) => {
+  const state = getState();
+  const apiOptions = apiOptionsFromState(state);
+  const model = modelFromThingId(id, state);
+  const distinguishTypeMap = {
+    'moderator': 'yes',
+    '': 'no',
+  };
```
madbook: 💅 I think that I was confused as to which `distinguish` method I was looking at before – this mapping of 'moderator' to 'yes', etc really should be in the `distinguish` endpoint function in node-api-client.  If another app were to use this endpoint for some reason, they would also need to know to do this mapping before calling `Modtools.distinguish`.  Better to pass in the unmapped value and have Modtools.distinguish handle the mapping internally

scarow:
:haircut_man:

madbook:
🐟

PR #969, opened 2017-01-27
Add announcement post modtool

madbook:
1f09562 addresses [CE-367](https://reddit.atlassian.net/browse/CE-367)
6d94690 addresses [CE-378](https://reddit.atlassian.net/browse/CE-378)

requires reddit/node-api-client#190

👓 @scarow @birakattack

### DIFF
--- a/src/app/actions/modTools.js
+++ b/src/app/actions/modTools.js
@@ -147,7 +147,7 @@ export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
 };
 
 /**
- * Sticky or unsticky a post (AKA an "Annoucement Post").
+ * Sticky or unsticky a post (AKA an "Announcement Post").
  * @function
  * @param {string} id The fullname of the post to sticky / unsticky
  * @param {boolean} isSettingSticky true if setting sticky, false if unsetting
--- a/src/app/actions/modTools.js
+++ b/src/app/actions/modTools.js
@@ -5,6 +5,7 @@ const { ResponseError } = errors;
 
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import modelFromThingId from 'app/reducers/helpers/modelFromThingId';
+import stickyPostsFromState from 'app/reducers/helpers/stickyPostsFromState';
 
 export const MODTOOLS_REMOVAL_PENDING = 'MODTOOLS_REMOVAL_PENDING';
 export const MODTOOLS_REMOVAL_ERROR = 'MODTOOLS_REMOVAL_ERROR';
@@ -15,6 +16,9 @@ export const MODTOOLS_APPROVAL_SUCCESS = 'MODTOOLS_APPROVAL_SUCCESS';
 export const FETCHING_MODERATING_SUBREDDITS = 'FETCHING_MODERATING_SUBREDDITS';
 export const RECEIVED_MODERATING_SUBREDDITS = 'RECEIVED_MODERATING_SUBREDDITS';
 export const FETCH_FAILED_MODERATING_SUBREDDITS = 'FETCH_FAILED_MODERATING_SUBREDDITS';
+export const MODTOOLS_SET_STICKY_POST_ERROR = 'MODTOOLS_SET_STICKY_POST_ERROR';
+export const MODTOOLS_SET_STICKY_POST_PENDING = 'MODTOOLS_SET_STICKY_POST_PENDING';
+export const MODTOOLS_SET_STICKY_POST_SUCCESS = 'MODTOOLS_SET_STICKY_POST_SUCCESS';
 
 export const removalPending = spam => ({
   type: MODTOOLS_REMOVAL_PENDING,
@@ -59,6 +63,20 @@ export const receivedSubs = apiResponse => ({
 export const fetchSubsFailed = error => ({
   type: FETCH_FAILED_MODERATING_SUBREDDITS,
   error,
+});
+
+const setStickyPostPending = () => ({
+  type: MODTOOLS_SET_STICKY_POST_PENDING,
+});
+
+const setStickyPostError = () => ({
+  type: MODTOOLS_SET_STICKY_POST_ERROR,
+});
+
+const setStickyPostSuccess = (thing, isStickied) => ({
+  type: MODTOOLS_SET_STICKY_POST_SUCCESS,
+  thing,
+  isStickied,
 });
 
 export const remove = (id, spam) => async (dispatch, getState) => {
@@ -120,6 +138,38 @@ export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
   } catch (e) {
     if (e instanceof ResponseError) {
       dispatch(fetchSubsFailed(e));
+    } else {
+      throw e;
+    }
+  }
+};
+
+/**
+ * Sticky or unsticky a post (AKA an "Annoucement Post").
+ * @function
+ * @param {string} id The fullname of the post to sticky / unsticky
+ * @param {boolean} isSettingSticky true if setting sticky, false if unsetting
+ */
+export const setStickyPost = (id, isSettingSticky) => async (dispatch, getState) => {
+  const state = getState();
+  const apiOptions = apiOptionsFromState(state);
+  const model = modelFromThingId(id, state);
+
+  dispatch(setStickyPostPending());
+
+  try {
+    await Modtools.setSubredditSticky(apiOptions, id, isSettingSticky);
+    dispatch(setStickyPostSuccess(model, isSettingSticky));
+
+    if (isSettingSticky) {
+      const modelToUnsticky = stickyPostsFromState(state)[1];
+      if (modelToUnsticky) {
+        dispatch(setStickyPostSuccess(modelToUnsticky, false));
+      }
+    }
+  } catch (e) {
+    if (e instanceof ResponseError) {
+      dispatch(setStickyPostError(e));
     } else {
       throw e;
     }
--- a/src/app/actions/modTools.js
+++ b/src/app/actions/modTools.js
@@ -17,7 +17,6 @@ export const FETCHING_MODERATING_SUBREDDITS = 'FETCHING_MODERATING_SUBREDDITS';
 export const RECEIVED_MODERATING_SUBREDDITS = 'RECEIVED_MODERATING_SUBREDDITS';
 export const FETCH_FAILED_MODERATING_SUBREDDITS = 'FETCH_FAILED_MODERATING_SUBREDDITS';
 export const MODTOOLS_SET_STICKY_POST_ERROR = 'MODTOOLS_SET_STICKY_POST_ERROR';
-export const MODTOOLS_SET_STICKY_POST_PENDING = 'MODTOOLS_SET_STICKY_POST_PENDING';
 export const MODTOOLS_SET_STICKY_POST_SUCCESS = 'MODTOOLS_SET_STICKY_POST_SUCCESS';
 
 export const removalPending = spam => ({
@@ -63,10 +62,6 @@ export const receivedSubs = apiResponse => ({
 export const fetchSubsFailed = error => ({
   type: FETCH_FAILED_MODERATING_SUBREDDITS,
   error,
-});
-
-const setStickyPostPending = () => ({
-  type: MODTOOLS_SET_STICKY_POST_PENDING,
 });
 
 const setStickyPostError = (error, isSettingSticky) => ({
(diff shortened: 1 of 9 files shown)


madbook:
<details>
  <summary>Screenshots</summary>
  screen shot 2017-01-27 at 12 51 21 pm
  screen shot 2017-01-27 at 12 51 30 pm
  screen shot 2017-01-27 at 12 51 51 pm
</details>

src/app/components/ModeratorModal/index.jsx
```diff
@@ -32,6 +32,11 @@ export class ModeratorModal extends React.Component {
           <div onClick={ this.props.onClick }>
             <div className='ModeratorModalRowWrapper'>
               <DropdownRow
+                icon='sticky'
+                text={ this.props.isSticky ? 'Unpin as announcement' : 'Pin as annoucement' }
+                onClick={ this.props.onToggleSticky }
+              />
+              <DropdownRow
```
scarow: Will this show for comments moderatorModals too? I think it should be hidden until comments can be stickied

src/app/actions/modTools.js
```diff
@@ -125,3 +143,35 @@ export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
...
+  dispatch(setStickyPostPending());
+
+  try {
+    await Modtools.setSubredditSticky(apiOptions, id, isSettingSticky);
+    dispatch(setStickyPostSuccess(model, isSettingSticky));
+
+    if (isSettingSticky) {
+      const modelToUnsticky = stickyPostsFromState(state)[1];
+      if (modelToUnsticky) {
+        dispatch(setStickyPostSuccess(modelToUnsticky, false));
```
scarow: I dont understand this part -- the double `dispatch(setStickyPostSuccess...` feels weird. Maybe I'm just not understanding what's happening

src/app/components/ModeratorModal/index.jsx
```diff
@@ -32,6 +32,11 @@ export class ModeratorModal extends React.Component {
           <div onClick={ this.props.onClick }>
             <div className='ModeratorModalRowWrapper'>
               <DropdownRow
+                icon='sticky'
+                text={ this.props.isSticky ? 'Unpin as announcement' : 'Pin as annoucement' }
+                onClick={ this.props.onToggleSticky }
```
scarow: Does this need an `isSelected` prop? I think we are planning to toggle the colors of the dropdown icon still.

madbook: ooh, right – i think i got confused by the discussion yesterday, but I do believe that's where we landed.

src/app/components/ModeratorModal/index.jsx
madbook: oh, yikes – i think i kind of didn't realize this was shared 😱

src/app/actions/modTools.js
madbook: There can only be two sticky posts in a subreddit at a time. If you sticky a post while there are 2 other stickied posts, the 2nd stickied post is replaced (unstickied).  This checks the post listing to see if there are 2 stickied posts, and if so, updates the model of the second one to remove the sticky icon.

scarow: Maybe leave a comment in the code here?

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -202,16 +200,24 @@ function renderPostFlair(post, single) {
...
 
   return (
     <span className='PostHeader__approval-status-flair'>
-      { isApproved ? APPROVED_FLAIR : null }
-      { isRemoved ? REMOVED_FLAIR : null }
-      { isSpam ? SPAM_FLAIR : null }
+      { approved ? APPROVED_FLAIR : null }
+      { removed ? REMOVED_FLAIR : null }
+      { spam ? SPAM_FLAIR : null }
+      { stickied ? STICKY_FLAIR : null }
```
scarow: Can you move this before the approved, removed, spam icons? just had IRL discussion that the approval status should always appear to the far-right of any other icons

src/app/reducers/helpers/stickyPostsFromState.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Return the list of sticky posts from the active PostList based on state
+ * @function
+ * @param {Object} state
+ * @returns {Object[]}
+ */
+export default function(state) {
+  const { currentPage } = state.platform;
+  const postsListParams = PostsFromSubredditHandler.pageParamsToSubredditPostsParams(currentPage);
+  const postsListId = paramsToPostsListsId(postsListParams);
```
birakattack: :+1:

madbook:
💅

madbook:
aaaah, i'll never learn

:haircut_man:

src/app/actions/modTools.js
```diff
@@ -61,6 +65,20 @@ export const fetchSubsFailed = error => ({
   error,
 });
 
+const setStickyPostPending = () => ({
+  type: MODTOOLS_SET_STICKY_POST_PENDING,
+});
+
+const setStickyPostError = () => ({
```
birakattack: We can pass the exception into this action and then construct the payload to look like:
```
{
  type: MODTOOLS_SET_STICKY_POST_SUCCESS,
  error,
  message: 'Message to be displayed in toast error',
}
```

You can then listen for this action in `app/reducers/toaster.js` with all the other failure method in order to emit a toaster with the passed message in the action.

src/app/reducers/posts.js
```diff
@@ -109,6 +109,21 @@ export default function(state=DEFAULT, action={}) {
...
+    case modToolActions.MODTOOLS_SET_STICKY_POST_SUCCESS: {
+      const { thing, isStickied } = action;
+
+      if (thing.type !== POST) { return state; }
+      
+      return mergeUpdatedModel(
+        state,
+        {
+          model: thing.set({
+            stickied: isStickied,
```
birakattack: Instead of passing `isStickied` you could also just do `!thing.stickied` which will achieve the same toggle functionality I believe you're after.

src/app/components/ModeratorModal/index.jsx
```diff
@@ -31,6 +33,15 @@ export class ModeratorModal extends React.Component {
           />
           <div onClick={ this.props.onClick }>
             <div className='ModeratorModalRowWrapper'>
+              { this.props.targetType === ModelTypes.POST
```
birakattack: Instead of adding all of this logic here would it be better to make another method called `renderStickyButton` that contains the logic to display the dropdown only for posts and then within the main render call just use `renderStickButton(this.props)` to prevent this area from getting really cluttered once all the post actions have been integrated?

src/app/reducers/posts.js
madbook: This is true, but then the action ends up looking really weird to me.  `modTools.setStickyPost` still requires the `isSettingSticky` bool argument, but it then effectively is only used to determine if we should try to _unsticky_ a post that's being replaced.  If it's called incorrectly for some reason, i think it'll lead to weird behavior:

case | target thing | `isSettingSticky` | ends up stickied? | 2nd existing post is unstickied?
---|---|---|---|---
A | false | false | **true** | false
B | false | true | true | true
C | true | false | false | true
D | true | true | **false** | **true**

I've emboldened the weird values above – in case A, if we called 

```javascript
modTools.setSubredditSticky(notStickyPostId, false);
```

we'd end up with a sticky post.  More weirdly, if we called

```javascript
modTools.setSubredditSticky(alreadyStickyPostId, true);
```

we'd end up unstickying that post _and_ unstickying the current \#2 sticky post in the subreddit (at least visually).

These scenarios aren't likely to happen if we are doing things correctly, but I prefer knowing that if I call that action creator with a specific value for `isSettingSticky`, that's the state it will end up in.

The other option would be to remove that argument altogether, pull the current value of `thing.stickied` from within the action creator, and fork the logic on that.  Is that generally preferred vs passing in explicit arguments? I'm more in favor of being explicit, but equally in favor of consistency :D

src/app/components/ModeratorModal/index.jsx
madbook: Personally, I am not a fan of breaking things out that way.  My experience has been that inlining is often better for readability than breaking small bits out into top level functions that live elsewhere.  I'd be more inclined to tidy it up by 

1. storing the result of the condition check in a const and using that in the template
2. using `a && b` instead of `a ? b : null`

something like

```jsx
const isTargetTypePost = this.props.targetType === ModelTypes.POST;
```

```jsx
<div className='ModeratorModalRowWrapper'>
  { isTargetTypePost && (
    <DropdownRow
      icon='sticky'
      text={ this.props.isSticky ? 'Unpin as announcement' : 'Pin as annoucement' }
      onClick={ this.props.onToggleSticky }
      isSelected={ this.props.isSticky }
    />
  ) }
  ...
</div>
```

madbook:
fixed a thing and responded to some things :haircut_man:

src/app/actions/modTools.js
```diff
@@ -125,3 +145,38 @@ export const fetchModeratingSubreddits = () => async (dispatch, getState) => {
...
+ * @function
+ * @param {string} id The fullname of the post to sticky / unsticky
+ * @param {boolean} isSettingSticky true if setting sticky, false if unsetting
+ */
+export const setStickyPost = (id, isSettingSticky) => async (dispatch, getState) => {
+  const state = getState();
+  const apiOptions = apiOptionsFromState(state);
+  const model = modelFromThingId(id, state);
+
+  dispatch(setStickyPostPending());
```
scarow: Are we using sticky pending state anywhere? Jesjit mentioned I should remove my `pending` actions if we weren't using them. Just thought I'd mention that here as well.

src/app/actions/modTools.js
```diff
@@ -61,6 +65,22 @@ export const fetchSubsFailed = error => ({
...
 });
 
+const setStickyPostPending = () => ({
+  type: MODTOOLS_SET_STICKY_POST_PENDING,
+});
+
+const setStickyPostError = (error, isSettingSticky) => ({
+  type: MODTOOLS_SET_STICKY_POST_ERROR,
+  error,
+  message: `Failed to ${isSettingSticky ? 'pin' : 'unpin'} announcement`,
```
birakattack: Well aren't we fancy :1st_place_medal:

src/app/actions/modTools.js
birakattack: I would second the notion that we can remove this as it does nothing at the moment. We can add this later if need be but for now I don't think this is necessary.

src/app/components/ModeratorModal/index.jsx
birakattack: That works for me 👍

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -275,7 +281,7 @@ function renderPostDescriptor(
     authorOrNil,
     !hideSubredditLabel && flairOrNil,
   ]);
-  const approvalStatusFlair = renderApprovalStatusFlair(post.approved, post.removed, post.spam);
+  const approvalStatusFlair = renderModStatusFlair(post);
```
birakattack: We should probably change the name of the variable here to `modStatusFlair` as well.

src/app/reducers/posts.js
birakattack: Ahh thanks for all the detail, I misunderstood how you were using `isSettingSticky`. :+1:

src/app/components/Post/PostHeader/index.jsx
madbook: doh! good catch

madbook:
:haircut_man:

scarow:
lgtm!! 🐟

src/app/reducers/posts.js
```diff
@@ -109,6 +109,21 @@ export default function(state=DEFAULT, action={}) {
...
+          }),
+        },
+      );
+    }
+
```
scarow: Should add a test for this

madbook: 👓  @scarow I've already rebased and resolved conflicts for this, so rather than force pushing up and losing history here, [I threw the diff up here](https://gist.github.com/madbook/fd0a8762c6a0370d880f4c018529c1f8)

madbook:
9c73506d5da644d7f33706e1666bc9910be8702d...b406ab189ddea033cd74ba621a300cc7c3697e38

PR #972, opened 2017-01-30
Comment flatlist WIP

curioussavage:
## ticket
https://reddit.atlassian.net/browse/CONT-35

## dependencies
 https://github.com/reddit/node-api-client/pull/197

## staging
https://comment-flatlist.mobile.staging.snooguts.net/

PR #975, opened 2017-01-31
Add sticky comment mod tools

madbook:
This PR is against a branch containing https://github.com/reddit/reddit-mobile/pull/969, which is not yet deployed.

for [CE-375](https://reddit.atlassian.net/browse/CE-375)

The primary action creator here, `actions/modTools#setStickyComment`, works similarly to the one for sticky posts (`actions/modTools#setStickyPost`), in that it additionally _unstickies_ any existing sticky comments in the thread.  A new helper is added to pull the existing sticky comment for the active thread out of state, `stickyCommentsFromState`, which works very similarly to the one added for sticky posts, `stickyPostsFromState`.

EDIT: requires reddit/node-api-client#194 to work

👓 @birakattack @scarow

### DIFF
--- a/src/app/actions/modTools.js
+++ b/src/app/actions/modTools.js
@@ -6,6 +6,7 @@ const { ResponseError } = errors;
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import modelFromThingId from 'app/reducers/helpers/modelFromThingId';
 import stickyPostsFromState from 'app/reducers/helpers/stickyPostsFromState';
+import stickyCommentsFromState from 'app/reducers/helpers/stickyCommentsFromState';
 
 export const MODTOOLS_REMOVAL_PENDING = 'MODTOOLS_REMOVAL_PENDING';
 export const MODTOOLS_REMOVAL_ERROR = 'MODTOOLS_REMOVAL_ERROR';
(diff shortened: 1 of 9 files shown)


src/app/reducers/comments.js
```diff
@@ -132,6 +132,21 @@ export default function(state=DEFAULT, action={}) {
...
+      return mergeUpdatedModel(
+        state,
+        {
+          model: thing.set({
+            stickied: isStickied,
+          }),
+        },
+      );
+    }
+
```
scarow: SHould add a test for this

madbook:
:haircut_man: for added test

src/app/reducers/comments.test.js
```diff
@@ -510,5 +510,61 @@ createTest({ reducers: { comments } }, ({ getStore, expect }) => {
...
+        });
+
+        store.dispatch(modToolActions.setStickyCommentSuccess(COMMENT_STICKIED, false));
+
+        const { comments } = store.getState();
+        expect(comments).to.eql({
+          [COMMENT_STICKIED.uuid]: COMMENT_UNSTICKIED,
+        });
+      });
+    });
```
scarow: Would it be possible to add a test here about stickied posts and comments bumping other stickies? i.e a test for when you already have 2 posts stickied, and you sticky another post, the stickied posts should include the most recent?

if that is complicated, we can skip

scarow:
💅  just wondering about one/two more tests. otherwise looks great

src/app/reducers/comments.test.js
madbook: I think it's a little complicated – that's handled inside the `setStickyComment` action creator, which is where the API call gets made.  I'm not exactly sure how to go about mocking that out.

src/app/actions/modTools.js
```diff
@@ -204,3 +219,35 @@ export const setStickyPost = (id, isSettingSticky) => async (dispatch, getState)
...
+    dispatch(setStickyCommentSuccess(model, isSettingSticky));
+
+    // If setting a sticky comment in a comments thread, the new sticky will
+    // replace an existing one. 
+    if (isSettingSticky) {
+      const modelToUnsticky = stickyCommentsFromState(state)[0];
+      if (modelToUnsticky) {
+        dispatch(setStickyCommentSuccess(modelToUnsticky, false));
+      }
+    }
```
madbook: the unsticky is a purely client-side action (i.e. there's no separate API call out to do the unsticky, since its an existing side-effect of the initial API call to sticky), so this _shouldn't_ be cause for concern.

madbook:
:haircut_man:

curioussavage:
fyi I will need to refactor some of this (the reducer/action) when merging https://github.com/reddit/reddit-mobile/pull/972

madbook:
b3ebe084d18000a027c0c11e2898cb162836b9f5

Thanks for the heads up @curioussavage

PR #978, opened 2017-01-31
Clear 'token' cookie on logout

scarow:
We are having a cookie persistence error on mweb. Essentially, when a user logs out, then navigates to a subreddit, the `token` cookie becomes populated with that user's old cookie, which logs them back in.

This is especially problematic when a user (`user1`) logs out, then logs in on an alt (`user2`), then navigates to a subreddit and is unknowingly logged in as `user1` again.

### DIFF
--- a/src/server/session/logoutproxy.js
+++ b/src/server/session/logoutproxy.js
@@ -10,6 +10,7 @@ export default (router) => {
     // (and amp, modmail, etc) is now available on the root reddit domain, so
     // we can unify these sessions and only have one cookie to clear.
     ctx.cookies.set('reddit_session', undefined, { domain: config.rootReddit });
+    ctx.cookies.set('token', undefined, { domain: config.rootReddit });
     ctx.cookies.set('reddit_session');
 
     clearSessionCookies(ctx);


scarow:
👓 @schwers @birakattack @prashtx @wting 

This has not been tested locally, as this problem does not exist in a local env. After speaking with @wting we believe it is safe to deploy.

src/server/session/logoutproxy.js
```diff
@@ -10,6 +10,7 @@ export default (router) => {
     // (and amp, modmail, etc) is now available on the root reddit domain, so
     // we can unify these sessions and only have one cookie to clear.
     ctx.cookies.set('reddit_session', undefined, { domain: config.rootReddit });
+    ctx.cookies.set('token', undefined, { domain: config.rootReddit });
     ctx.cookies.set('reddit_session');
```
prashtx: Do we need to also duplicate this empty `set` line for the `token` cookie?

prashtx: Spoke IRL. We should be good and will find out soon enough if we need the other line.

prashtx:
:fish:

src/server/session/logoutproxy.js
scarow: Took convo offline but consensus is that `undefined` clears the cookie in a more defined way (as opposed to `ctx.cookies.set('token')`), so we will try this version for now

wting:
:fish:

PR #980, opened 2017-02-01
Chan 304 improve xpromo events

telaviv:
👓: @nramadas @uzi @prashtx @arjunb023 

This change is split into two commits

1) The first is an implementation [xpromo utm app store tracking spec](https://reddit.atlassian.net/wiki/display/DA/XPromo+app+store+tracking+spec.)
 
For the spec I needed to think of ways to get information about the current relevant experiment and variant. For instance. We have several related experiments related to `require_login`. Some having to do with android/ios or front page/listing. For a particular experiment I needed to identify the relevant one and see if a user is part of that experiment. So if a user is viewing the "require login" interstital on a listing page on android I had to just make sure we were considering the `mweb_xpromo_require_login_listing_android` experiment. This also applied to users who were bucketed into a control of this experiment.

To accomplish all this, I blew up the number of feature flags related the the experiments. The alternative would have been copy and pasting the same checks that I do with feature flags like `isAndroid` or `isFrontPage`. This seemed less bad. 

2) The second commit is an implementation of the new [xpromo event](https://reddit.atlassian.net/wiki/display/DA/XPromo+Events)

You'll note I do something kind of unusual by creating a new type of middleware. The reason is because I needed to fire these events from pretty random parts of the codebase. Components, actions, and even a reducer. In order to fire the event for realsy I always needed to have full access to the state, which was either difficult, or in one case ( the reducer) impossible. The cleaner solution seemed to be just dispatch an action, and let the middleware take care of side effects that have nothing to do with the state of the application.


src/app/actions/xpromo.js
```diff
@@ -38,8 +56,19 @@ export const close = () => async (dispatch, getState) => {
 
 };
 
-export const checkAndSet = () => async (dispatch) => {
-  if (shouldShowBanner()) {
+export const checkAndSet = (getState) => async (dispatch) => {
```
telaviv: should be `async(dispatch, getState)`

src/app/actions/xpromo.js
```diff
@@ -38,8 +56,19 @@ export const close = () => async (dispatch, getState) => {
...
+  } else {
+    dispatch(trackXPromoEvent(XPROMO_VIEW, { interstitial_type: interstitialType(state) }));
     dispatch(show());
   }
 };
+
+export const navigateToAppStore = (url, visitType) => async (dispatch) => {
+  dispatch(trackXPromoEvent(XPROMO_DISMISS, { dismiss_type: 'app_store_visit' }));
+  dispatch(trackXPromoEvent(XPROMO_APP_STORE_VISIT, { visit_trigger: visitType }));
+  dispatch(redirect(url));
```
telaviv: flush events before redirect

prashtx:
From IRL discussion:

General nit: we try to use `state => state.foo` rather than `(state) => { return state.foo; }`, etc.

We want to add experiment name + variant to the xpromo events.

src/app/components/DualPartInterstitial/Footer/index.jsx
```diff
@@ -115,42 +112,37 @@ class DualPartInterstitialFooter extends React.Component {
...
+
+  return getBranchLink(state, payload);
 }
 
 const selector = createSelector(
   getSubreddit,
+  requireXPromoLogin,
   (state) => { return features.withContext({ state }); },
   (state) => { return state.user.loggedOut; },
   (state) => {
```
uzi: Yeah, these should all be updated to stuff like `state => state.user.loggedOut,`

src/app/featureFlags.js
```diff
@@ -176,46 +183,83 @@ const config = {
...
+        { variant: 'mweb_xpromo_require_login_listing_ios:login_required' },
+      ] },
+    ],
+  },
+  [VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_IOS_CONTROL]: {
+    and: [
+      { allowedPages: ['listing'] },
+      { allowNSFW: false },
+      { allowedDevices: [IPHONE] },
+    ],
```
uzi: Don't need to specify the variant on this one?

src/app/featureFlags.js
```diff
@@ -321,8 +364,8 @@ flags.addRule('variant', function (name) {
         experiment_id,
         experiment_name,
         variant,
-        user_id: !this.state.user.loggedOut ? parseInt(user.id, 36) : null,
-        user_name: !this.state.user.loggedOut ? user.name : null,
+        user_id: !this.state.user.loggedOut ? parseInt(this.state.user.id, 36) : null,
+        user_name: !this.state.user.loggedOut ? this.state.user.name : null,
```
uzi: Might be nice to toss a `const { user } = this.state;` above and use that instead of the whole "this.user.state" thing a bunch.

src/lib/experiments.js
```diff
@@ -0,0 +1,14 @@
+function extractUser(state) {
```
uzi: This is also defined in xpromo.js.  Consolidate these files or remove the other one?

src/lib/eventUtils.js
```diff
@@ -50,7 +56,7 @@ export function getUserInfoOrLoid(state) {
     };
   }
 
-  const loid = state.loid; 
+  const loid = state.loid;
```
uzi: Could also do `const { loid } = state;`

uzi:
💅 Pretty reasonable overall -- just needs some cleanup love.

src/app/featureFlags.js
telaviv: whoa, good catch

telaviv:
💇 ok i did find a bug here. It's complicated so let's talk in person.

src/app/components/DualPartInterstitial/Footer/index.jsx
```diff
@@ -115,42 +114,37 @@ class DualPartInterstitialFooter extends React.Component {
...
-  (state) => { return features.withContext({ state }); },
-  (state) => { return state.user.loggedOut; },
-  (state) => {
-    return (campaign, medium) => {
-      return createNativeAppLink(state, campaign, medium);
+  requireXPromoLogin,
+  state => features.withContext({ state }),
+  state => state.user.loggedOut,
+  state => {
+    return (linkType) => {
```
uzi: Remove the parenthesis around `linkType`... also, can this be

``` state => linkType => createNativeAppLink(state, linkType);```

?

File: src/app/selectors/xpromo.js [hunk 9]
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export function loginRequiredEnabled(state) {
+  const featureContext = features.withContext({ state });
+  return some([
+    VARIANT_XPROMO_LOGIN_REQUIRED_FP_IOS,
+    VARIANT_XPROMO_LOGIN_REQUIRED_FP_ANDROID,
+    VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_IOS,
+    VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_ANDROID,
+  ], (feature) => { return featureContext.enabled(feature); });
```
uzi: I think this last part can be

```], feature => featureContext.enabled);```

File: src/app/selectors/xpromo.js [hunk 10]
```diff
@@ -0,0 +1,68 @@
...
+  const featureFlag = find([
+    VARIANT_XPROMO_LOGIN_REQUIRED_FP_IOS,
+    VARIANT_XPROMO_LOGIN_REQUIRED_FP_ANDROID,
+    VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_IOS,
+    VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_ANDROID,
+    VARIANT_XPROMO_LOGIN_REQUIRED_FP_IOS_CONTROL,
+    VARIANT_XPROMO_LOGIN_REQUIRED_FP_ANDROID_CONTROL,
+    VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_IOS_CONTROL,
+    VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_ANDROID_CONTROL,
+  ], (feature) => { return featureContext.enabled(feature); });
```
uzi: Same here

uzi:
💅  Few more nits, but looking better.

telaviv:
💇

src/app/actions/xpromo.js
```diff
@@ -56,13 +53,8 @@ export const close = () => async (dispatch, getState) => {
...
 };
 
-export const checkAndSet = () => async (dispatch, getState) => {
-  const state = getState();
-  const ineligibilityReason = shouldNotShowBanner();
-  if (ineligibilityReason) {
-    dispatch(trackXPromoEvent(XPROMO_INELIGIBLE, { ineligibility_reason: ineligibilityReason }));
-  } else {
-    dispatch(trackXPromoEvent(XPROMO_VIEW, { interstitial_type: interstitialType(state) }));
+export const checkAndSet = () => async (dispatch) => {
```
uzi: Remove parenthesis around `dispatch`

src/app/selectors/xpromo.js
```diff
@@ -26,14 +28,21 @@ const EXPERIMENT_NAMES = {
...
+
 export function loginRequiredEnabled(state) {
   const featureContext = features.withContext({ state });
   return some([
     VARIANT_XPROMO_LOGIN_REQUIRED_FP_IOS,
     VARIANT_XPROMO_LOGIN_REQUIRED_FP_ANDROID,
     VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_IOS,
     VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_ANDROID,
-  ], (feature) => { return featureContext.enabled(feature); });
+  ], feature => featureContext.enabled(feature));
```
uzi: Try making this `], feature => featureContext.enabled);`

src/app/selectors/xpromo.js
```diff
@@ -47,7 +56,7 @@ function loginExperimentName(state) {
     VARIANT_XPROMO_LOGIN_REQUIRED_FP_ANDROID_CONTROL,
     VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_IOS_CONTROL,
     VARIANT_XPROMO_LOGIN_REQUIRED_SUBREDDIT_ANDROID_CONTROL,
-  ], (feature) => { return featureContext.enabled(feature); });
+  ], feature => featureContext.enabled(feature));
```
uzi: And this

uzi:
💅

telaviv:
💇 Fixed two issues:

1) The ineligibility event wasn't triggering as often as it should be. We should have either a view or ineligibility event trigger for every listing page, including the front page.

2) The events sent right before the redirect may not finish. I've added a quarter second delay for now. It's short enough to not feel noticeable. The cool thing is we should be able to cross reference this with branch and see how many people we are losing by not delaying for longer.

src/app/actions/xpromo.js
```diff
@@ -39,7 +54,16 @@ export const close = () => async (dispatch, getState) => {
...
-  if (shouldShowBanner()) {
+  if (!shouldNotShowBanner()) {
     dispatch(show());
   }
 };
+
+export const navigateToAppStore = (url, visitType) => async (dispatch) => {
+  dispatch(trackXPromoEvent(XPROMO_DISMISS, { dismiss_type: 'app_store_visit' }));
+  dispatch(trackXPromoEvent(XPROMO_APP_STORE_VISIT, { visit_trigger: visitType }));
+  // TODO (skrisman): There's no guarantee that these events will actually finish
```
uzi: And this is something we'll do as a follow-up patch?

telaviv: Yes! We will need to do a patch on `event-tracker` first and then use those changes here. So it's a bit more involved.

telaviv:
💇

uzi:
:fish:

PR #982, opened 2017-02-02
Modtool fixup 1

scarow:
👓 @schwers @birakattack 

You can test on https://modtool-fixup-1.mobile.staging.snooguts.net (note that production cache behavior is different on staging so you can't actually repro the problem -- I am basically making sure that the cache headers don't crash anything or log people out unexpectedly).

Note that on staging, when you navigate to `r/pics` (for example) via the url (not inside the app), you get logged out. This is old behavior, not introduced by this change.

I also added some checks to `fetchModeratingSubreddits` to avoid API calls.

### DIFF
--- a/src/app/actions/user.js
+++ b/src/app/actions/user.js
@@ -1,4 +1,5 @@
 import * as accountActions from 'app/actions/accounts';
+import * as modToolActions from 'app/actions/modTools';
 
 export const fetchMyUser = () => async (dispatch, getState) => {
   const state = getState();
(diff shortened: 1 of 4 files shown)


src/app/actions/user.js
```diff
@@ -7,4 +8,5 @@ export const fetchMyUser = () => async (dispatch, getState) => {
   // we prevent the promise that `fetchMyUser` returns from resolving until
   // the dispatch is complete.
   await dispatch(accountActions.fetch({ name: 'me', loggedOut: !state.session.accessToken }));
+  await dispatch(modToolActions.fetchModeratingSubreddits());
```
scarow: oh yeah. good catch heh!

schwers: ^ I'd say yes. The handlerCommon invocation probably won't do anything the first time its called, as the user won't be fetched yet. But it'd be a lot clearer if we removed it there

scarow:
:haircut_man:

PR #986, opened 2017-02-03
Pep8 Cleanup

birakattack:
Removed the errors I introduced from my mod tools patch. This removes all errors from those commits. 

:eyeglasses: @scarow


scarow:
🐟 lgtm once lint passes

PR #987, opened 2017-02-04
Force the captcha to reload on registration errors

nramadas:
When a user gets form errors when trying to register an account, the captcha that they created is no longer valid. However, we don't prompt them to re-do the captcha and get re-issued a valid `gRecaptchaResponse` to pass to the api.

This patch will cause the recaptcha to reset if the user had errors while trying to create an account.

NOTE: This isn't a _fantastic_ product solution since this potentially forces the user to have to do the captcha verification more than once. In an ideal world, we would validate the form first, and if everything looks ok, ask the user to verify through recaptcha. However, what's in this patch is a quick fix that will restore the ability for users to register.


telaviv:
fix the lint and 🐟  !

PR #988, opened 2017-02-06
Fix modal position

scarow:
👓 @madbook


src/app/components/Dropdown/styles.less
```diff
@@ -30,8 +30,9 @@
   max-width: 95%;
```
madbook: I think you'll also need to change this from `max-width` to `width`

scarow:
💅

scarow:
:haircut_man:

PR #984, opened 2017-02-02
Build subreddit rules page on mweb

madbook:
for [CE-462](https://reddit.atlassian.net/browse/CE-462)

requires reddit/node-api-client#195

This adds support for the about/rules page.  It looks like this:

screen shot 2017-02-02 at 6 33 41 pm

This is essentially a direct port of the read-only on desktop.  Creating/editing rules (for mods) is not yes supported.

👓 @birakattack @scarow

PR #993, opened 2017-02-08
Add link to rules from community details

madbook:
For [**CE-497**](https://reddit.atlassian.net/browse/CE-497)

5cbfd7f is purely a refactor, inlining the 3 local render functions into the main component body makes it easier to reason about how props affect the rendering, as well as reduce the size of the component overall (the increased line count comes from the addition of jsdoc comment, propTypes, defaultProps, etc).

9875acb actually adds in the link.  It's positioned for consistency with the "About this community" link that exists on the subreddit listing view

subreddit listing | about page
---|---
screen shot 2017-02-08 at 1 33 52 pm | screen shot 2017-02-08 at 1 34 01 pm

While this doesn't _strictly_ depend on #984, it should not go out before it (otherwise the new link would just go to a 404 page).

👓 @scarow @birakattack

### DIFF
--- a/src/app/components/SubredditAbout/index.js
+++ b/src/app/components/SubredditAbout/index.js
@@ -28,12 +28,19 @@ function SubredditAbout(props) {
   return (
     <div className='SubredditAbout'>
     { subreddit
-      ? <RedditLinkHijacker>
-          <div
-            className='SubredditAbout__description'
-            dangerouslySetInnerHTML={ { __html: subreddit.descriptionHTML } }
-          />
-        </RedditLinkHijacker>
+      ? <div className="SubredditAbout__content">
+          <div className="SubredditAbout__community-rules-link">
+            <Anchor href={ `/r/${subredditName}/about/rules` }>
+              View this community&rsquo;s rules
+            </Anchor>
+          </div>
+          <RedditLinkHijacker>
+            <div
+              className='SubredditAbout__description'
+              dangerouslySetInnerHTML={ { __html: subreddit.descriptionHTML } }
+            />
+          </RedditLinkHijacker>
+        </div>
       
     : subredditRequest && subredditRequest.failed 
       ? <div className='SubredditAbout__loading-error'>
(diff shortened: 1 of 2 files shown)


src/app/components/SubredditAbout/index.js
```diff
@@ -9,41 +9,75 @@ import { Anchor } from '@r/platform/components';
...
+        
+      : <Loading />
+    }
     </div>
   );
+}
+
+SubredditAbout.propTypes = {
+  subreddit: T.shape({
+    descriptionHTML: T.string.isRequired,
```
scarow: Oooo what is `descriptionHTML`?

src/app/components/SubredditAbout/index.js
```diff
@@ -9,41 +9,75 @@ import { Anchor } from '@r/platform/components';
...
+  }),
+  user: T.shape({
+    loggedOut: T.bool.isRequired,
+  }),
+};
+
+SubredditAbout.defaultProps = {
+  subreddit: null,
+  subredditRequest: null,
+  user: null,
```
scarow: Are these props required above? Not sure if the required items in `T.shape` force all props to be passed. If so, we don't need default props.

scarow:
💅 looks good just some questions

src/app/components/SubredditAbout/index.js
madbook: The `T.shape` validator should not auto-force props in the shape to be passed, unless you tack on `isRequired` at the end, so these are still nullable

src/app/components/SubredditAbout/index.js
madbook: it comes from the `@r/api-client` model [here](https://github.com/reddit/node-api-client/blob/3X/src/models2/Subreddit.es6.js#L59).  It's the subreddit's sidebar text rendered to HTML.

madbook:
:haircut_man: RE the questions

scarow:
🐟

madbook:
1f412035cdc728387afb97a074c70241b85cbb34...3943be214563e72e8a83d9fbe7575e0d5f43e681

PR #995, opened 2017-02-08
Begin payload reduction

schwers:
This is the first in a series of payload optimizations. 

For an overview (@madbook, this should help explain @curioussavage's patch) please see: https://github.com/schwers/reddit-mobile/pull/3 
* note: these numbers are somewhat outdated as there has since been additional commits to mweb, and we're now including raven. 

👓  @curioussavage (cc @madbook, @birakattack)

### Commits
The first commit in this patch sets us up to do analysis step by step, and have a visualization to guide our efforts.

The second commit is "the low hanging fruit" of potential optimizations we can make.

### Results
Before and After:
* 357 KB gzipped -> 264 KB gzipped
* 1.3 MB -> 1017 KB

This work, and the rest of the work in the bundle audit can be visualized with `webpack-bundle-analyzer`. With the first commit you can run `npm run analyze` to generate interactive visualizations locally. The sizing information it prints, is slightly different than the numbers you'll see in chrome. The number's I'm referencing in the these results are from gzipping `ProductionClient` and running `npm run start`, and using the stats from Chrome's `Network` tab.

Before this patch, you can see a large chunks of: `elliptic`, `bn`(big nums), `browserify-aes`, `asn1`, `parse-asn1`, `hash`, `sha`, `des`, `diffie-hellmen`, node's `Buffer`

image

After this patch's second commit, that bulk is removed:

image

### DIFF
--- a/src/server/meta/index.js
+++ b/src/server/meta/index.js
@@ -1,5 +1,5 @@
-import crypto from 'crypto';
-
+import { atob } from 'Base64';
+import HmacSha1 from 'crypto-js/hmac-sha1';
 import superagent from 'superagent';
 
 import { formatLogJSON } from 'lib/errorLog';
@@ -196,16 +196,8 @@ export default (router, apiOptions) => {
       return;
     }
 
-    const secret = (new Buffer(apiOptions.actionNameSecret, 'base64')).toString();
-    const algorithm = 'sha1';
-
-    const hmac = crypto.createHmac(algorithm, secret);
-    hmac.setEncoding('hex');
-    hmac.write(timings.actionName);
-    hmac.end();
-
-    const hash = hmac.read();
-
+    const secret = atob(apiOptions.actionNameSecret);
+    const hash = HmacSha1(timings.actionName, secret).toString();
     timings.verification = hash;
 
     superagent
--- a/blueprints.config.js
+++ b/blueprints.config.js
@@ -13,6 +13,36 @@ function generateReleaseVersion() {
   }
 }
 
+function makeBundleAnalyzer(name, port) {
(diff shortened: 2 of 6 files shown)


madbook:
🔕 This is cool stuff! Is there a ticket linked to this?

curioussavage:
Technically its lumped in with this ticket just like my redux archiver one. https://reddit.atlassian.net/browse/CONT-163

There will be at least one more pull request we extract from @schwers work. Just nicer to test/review it in parts.

src/server/meta/index.js
```diff
@@ -196,16 +196,8 @@ export default (router, apiOptions) => {
...
-
-    const hmac = crypto.createHmac(algorithm, secret);
-    hmac.setEncoding('hex');
-    hmac.write(timings.actionName);
-    hmac.end();
-
-    const hash = hmac.read();
-
+    const secret = atob(apiOptions.actionNameSecret);
+    const hash =  HmacSha1(timings.actionName, secret);
```
curioussavage: looks like the space after equals is the cause of the linter error

curioussavage:
Only concern is if the output is still the same for the modified functions.

I was able to login while testing this.
💅

curioussavage:
woops, didn't mean to close this.

schwers:
I made a branch you can use to verify these changes produce the same values as before. Testing relies on running that code locally, and verifying in the debug console. I added an explicit call of the amp link generation to test, as it seems `createAmpHtmlLinkFromState` has a minor bug (I'll make a separate ticket).

https://github.com/schwers/reddit-mobile/commit/b7dc4dbc7baec5ac5577e942e12f4ff8583b199e

:haircut_man:

curioussavage:
I just ran your test branch. everything looks good so far.

curioussavage:
🐟

PR #996, opened 2017-02-08
Stop bucketing users who aren't part of an experiment.

telaviv:
We were bucketing users who were not part of an experiment. Mostly due to a bad truthy check.

👓  @uzi @arjunb023 @prashtx


src/app/featureFlags.js
```diff
@@ -345,7 +345,7 @@ const firstBuckets = new Set();
 flags.addRule('variant', function (name) {
   const [experiment_name, checkedVariant] = name.split(':');
   const experimentData = getExperimentData(this.state, experiment_name);
-  if (experimentData) {
+  if (Object.keys(experimentData).length !== 0) { // can be an empty object.
```
prashtx: Is it better to return `null` from `getExperimentData`? It sounds like it's always an error to access properties of the return value of `getExperimentData` when the current user is not part of the experiment in question. That would be more straightforward than checking the number of keys and would alert future consumers via an error if they've used this function counter to your intention.

prashtx:
💅

telaviv:
Two more feature requests here.

1) We fixed the bucketing issues surrounding require_login. We should only fire bucketing events if the user is both logged out and hasn't dismissed the event. Note I couldn't get this to work for the transparent experiment because we need to check the transparent bucket to know if we should even fire an xpromo view/ineligiblity event. 

Since we are purposefully trying not check variants, we also will be missing experiment data for the xpromo ineligiblity events. There's a weird catch 22 here. We can't know what experiment we're in unless we check it. We can't check it without fire a bucketing event. The bucketing event, at the moment, means they saw it. So we're a bit trapped with the current feature flag system.

2) Changed the experiment names. 

💇

telaviv:
💇 removed our control experiment

src/app/components/DualPartInterstitial/Footer/index.jsx
```diff
@@ -135,10 +141,8 @@ const selector = createSelector(
...
   requireXPromoLogin,
   state => features.withContext({ state }),
-  state => state.user.loggedOut,
   state => linkType => createNativeAppLink(state, linkType),
-  (subredditName, requireXPromoLogin, featureContext, loggedOut, createLink) => {
-    const requireLogin = loggedOut && requireXPromoLogin;
+  (subredditName, requireLogin, featureContext, createLink) => {
     const nativeInterstitialLink = createLink('interstitial');
     const nativeLoginLink = createLink('login');
     return { subredditName, requireLogin, nativeInterstitialLink, nativeLoginLink };
```
uzi: Do we still intend on returning `requireLogin` which is not undefined?  Can we explicitly set it if so?

uzi: Oh wait... brain fart...

uzi: Change that... do we need featureContext here?  Looks unused.

telaviv:
💇

telaviv:
i missed the experiment name
:haircut_man:

uzi:
:fish:

telaviv:
:haircut_man:

uzi:
:fish: assuming tests pass

telaviv:
:haircut_man:

uzi:
:fish: more fishes for you

PR #997, opened 2017-02-09
Use eventUtils to generate user info on bucketing events

schwers:
In testing https://github.com/reddit/node-api-client/pull/187, I noticed what seems to be a bug in bucketing events. `app/featureFlags` was adding a `user_id` field on its events, but it's not using any of the common event payload code. As a sideffect, it's not generating the save `user_id` value as the other events are. 

This patch pulls in `getUserInfoOrLoid` to do that work. I'm not sure what fields are expected on the `bucketing` events, but it might be better to use `getBasePayload`

👓  @telaviv @nramadas @curioussavage

### DIFF
--- a/src/app/featureFlags.js
+++ b/src/app/featureFlags.js
@@ -2,14 +2,16 @@ import Flags from '@r/flags';
 import omitBy from 'lodash/omitBy';
 import isNull from 'lodash/isNull';
 import sha1 from 'sha1';
+import url from 'url';
+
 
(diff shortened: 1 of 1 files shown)


curioussavage:
🐟

telaviv:
Yeah I think basePayload makes more sense. I think when it comes to data, the more the better. 
💅

schwers:
💇  now using `getBasePayload` 

cc  @drunken-economist

telaviv:
🐟

PR #1001, opened 2017-02-14
Make mweb consistent in nsfw rendering

birakattack:
JIRA: https://reddit.atlassian.net/browse/CE-480

Currently mweb does a regex check to see if the post contains `NSFW` or `NSFL` (case-insensitive) and if it does it renders the post as `NSFW`. This is inconsistent with our other platforms so we have opted to remove it so that the rendering will respect what the API returns. 

:eyeglasses: @madbook


madbook:
🐟 !

PR #1004, opened 2017-02-14
Comment report actions

scarow:
https://reddit.atlassian.net/browse/CE-369
https://reddit.atlassian.net/browse/CE-382
https://reddit.atlassian.net/browse/CE-534 (add reports modal -- just combined the PRs)
https://reddit.atlassian.net/browse/CE-373
https://reddit.atlassian.net/browse/CE-383

https://comment-report-actions.mobile.staging.snooguts.net

Note that when a post/comment gets approved, the reports are cleared on refresh.
Also note that approved posts can not be reported (you can technically report them from the interface, but a report flag will not show up).

### DIFF
--- a/src/app/components/Comment/CommentDropdown/index.jsx
+++ b/src/app/components/Comment/CommentDropdown/index.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { DropdownModal, DropdownRow, DropdownLinkRow } from 'app/components/Dropdown';
 import ModeratorModal from 'app/components/ModeratorModal';
+import { ReportsModal } from 'app/components/ReportsModal';
 
 import { COMMENT } from 'apiClient/models/thingTypes';
 
@@ -30,6 +31,8 @@ export default function CommentDropdown(props) {
     modModalId,
     isMine,
     distinguishType,
+    reportModalId,
+    reports,
   } = props;
 
   const userIsAuthor = commentAuthor === username;
@@ -57,21 +60,34 @@ export default function CommentDropdown(props) {
 
   if (showModModal && isSubredditModerator) {
     modal = (
-      <ModeratorModal
-        id={ id }
-        modModalId={ modModalId }
-        onClick={ onToggleModal }
-        isRemoved={ isRemoved }
-        isApproved={ isApproved }
-        isSpam={ isSpam }
-        approvedBy={ approvedBy }
-        removedBy={ removedBy }
-        distinguishType={ distinguishType }
-        isMine={ isMine }
-        isSticky={ isSticky }
-        targetType={ COMMENT }
-      >
-      </ModeratorModal>
+      <div>
+        <ModeratorModal
+          id={ id }
+          modModalId={ modModalId }
+          onClick={ onToggleModal }
+          isRemoved={ isRemoved }
+          isApproved={ isApproved }
+          isSpam={ isSpam }
+          approvedBy={ approvedBy }
+          removedBy={ removedBy }
+          distinguishType={ distinguishType }
+          isMine={ isMine }
+          isSticky={ isSticky }
+          targetType={ COMMENT }
+          reports={ reports }
+          reportModalId={ reportModalId }
+        />
+        <ReportsModal
+          reportModalId={ reportModalId }
+          isRemoved={ isRemoved }
+          isApproved={ isApproved }
+          isSpam={ isSpam }
+          approvedBy={ approvedBy }
+          removedBy={ removedBy }
+          reports={ reports }
+          onClick={ onToggleModal }
+        />
+      </div>
     );
   } else {
     modal = (
@@ -104,6 +120,8 @@ CommentDropdown.propTypes = {
   modModalId: T.string,
   distinguishType: T.string,
   isMine: T.bool,
+  reportModalId: T.string,
+  reports: T.object,
 };
 
 CommentDropdown.defaultProps = {
@@ -119,7 +137,9 @@ CommentDropdown.defaultProps = {
   approvedBy: null,
   removedBy: null,
   showModModal: false,
-  modModalId: null,
+  modModalId: '',
   isDistinguished: '',
   isMine: false,
+  reports: {},
+  reportModalId: '',
 };
--- a/src/app/components/Comment/CommentHeader/index.jsx
+++ b/src/app/components/Comment/CommentHeader/index.jsx
@@ -4,20 +4,21 @@ import React from 'react';
 import fill from 'lodash/fill';
 
 import { short } from 'lib/formatDifference';
+import { sumReportsCount } from 'lib/modToolHelpers.js';
 
 const T = React.PropTypes;
 const separator = <div className='CommentHeader__separator'> • </div>;
 
 const APPROVED_FLAIR = (
-  <td className='CommentHeader__approvalStatus icon icon-check-circled green' />
+  <td className='CommentHeader__modStatus icon icon-check-circled green' />
 );
 
 const REMOVED_FLAIR = (
-  <td className='CommentHeader__approvalStatus icon icon-delete_remove ban-red' />
+  <td className='CommentHeader__modStatus icon icon-delete_remove ban-red' />
 );
 
 const SPAM_FLAIR = (
-  <td className='CommentHeader__approvalStatus icon icon-spam nsfw-salmon' />
+  <td className='CommentHeader__modStatus icon icon-spam nsfw-salmon' />
 );
 
 function getAuthorIcon(authorType) {
@@ -126,7 +127,15 @@ export default function CommentHeader(props) {
     isApproved,
     isRemoved,
     isSpam,
+    reports,
   } = props;
+
+  const reportCounts = sumReportsCount(reports);
+
+  const REPORT_FLAIR = ([
+    <td className={ `CommentHeader__modStatus icon icon-flag ${ isApproved ? '' : 'warning-yellow' }` }/>,
+    <td className={ `CommentHeader__modStatus m-report-count ${ isApproved ? '' : 'warning-yellow' }` }>{ reportCounts }</td>,
+  ]);
 
   return (
     <div className='CommentHeader' onClick={ props.onToggleCollapse }>
@@ -139,6 +148,7 @@ export default function CommentHeader(props) {
             { isApproved ? APPROVED_FLAIR : null }
             { isRemoved ? REMOVED_FLAIR : null }
             { isSpam ? SPAM_FLAIR : null }
+            { reportCounts > 0 ? REPORT_FLAIR : null }
           </tr>
         </tbody>
       </table>
@@ -160,6 +170,7 @@ CommentHeader.propTypes = {
   isApproved: T.bool,
   isRemoved: T.bool,
   isSpam: T.bool,
+  reports: T.object,
 };
 
 CommentHeader.defaultProps = {
@@ -172,4 +183,5 @@ CommentHeader.defaultProps = {
   isApproved: false,
   isRemoved: false,
   isSpam: false,
+  reports: null,
 };
--- a/src/app/components/ModeratorModal/index.jsx
+++ b/src/app/components/ModeratorModal/index.jsx
@@ -5,19 +5,19 @@ import { createSelector } from 'reselect';
 
 import { POST, COMMENT } from 'apiClient/models/thingTypes';
 import Modtools from 'apiClient/apis/ModTools';
-
-import { Modal } from '@r/widgets/modal';
+import { Modal, ModalTarget } from '@r/widgets/modal';
 import { ModalBanner } from 'app/components/ModalBanner';
-import { DropdownRow } from 'app/components/Dropdown';
+import { DropdownRow, DropdownClose } from 'app/components/Dropdown';
 import modelFromThingId from 'app/reducers/helpers/modelFromThingId';
 import { getStatusBy, getApprovalStatus } from 'lib/modToolHelpers.js';
 import * as modActions from 'app/actions/modTools';
+import { USER_REPORT_KEY, MODERATOR_REPORT_KEY } from 'app/reducers/reports';
+
 
 const DISTINGUISH_TYPES = Modtools.DISTINGUISH_TYPES;
 const T = React.PropTypes;
 
 export class ModeratorModal extends React.Component {
-
   onDistinguish(distinguishType) {
     const type = (distinguishType === DISTINGUISH_TYPES.NONE
                           ? DISTINGUISH_TYPES.MODERATOR
@@ -27,6 +27,22 @@ export class ModeratorModal extends React.Component {
 
   showDistinguish(distinguishType) {
     return !(distinguishType === DISTINGUISH_TYPES.NONE);
+  }
+
+  preventModalClose(e) {
+    e.stopPropagation();
+  }
+
+  hasReports(reports) {
+    if (!reports) { return false; }
+    if (reports[USER_REPORT_KEY] 
+        && Object.keys(reports[USER_REPORT_KEY]).length > 0) {
+      return true;
+    }
+    if (reports[MODERATOR_REPORT_KEY]
+        && Object.keys(reports[MODERATOR_REPORT_KEY]).length > 0) {
+      return true;
+    }
   }
 
   render() {
@@ -44,12 +60,7 @@ export class ModeratorModal extends React.Component {
           id={ this.props.modModalId }
           className='DropdownModal ModeratorModal'
         >
-          <div className='DropdownClose'>
-            <DropdownRow
-              icon='x'
-              onClick={ this.props.onClick }
-            />
-          </div>
+          <DropdownClose onClick={ this.props.onClick }/>
           <div onClick={ this.props.onClick }>
             <ModalBanner
               status={ getApprovalStatus(this.props.isApproved,
@@ -85,6 +96,19 @@ export class ModeratorModal extends React.Component {
                   />,
                 ]
                 : null
+              }
+              { this.hasReports(this.props.reports) &&
+                <div onClick={ (e) => this.preventModalClose(e) }>
+                  <ModalTarget
+                    id={ this.props.reportModalId }
+                  >
+                    <DropdownRow
+                      icon='flag'
+                      text='Reports'
+                      isSelected={ true }
+                    />
+                  </ModalTarget>
+                </div>
               }
               { this.props.isMine
                 ? <DropdownRow
@@ -150,12 +174,14 @@ ModeratorModal.propTypes = {
   approvedBy: T.string,
   distinguishType: T.string,
   isMine: T.bool,
+  reports: T.object,
   target: T.object,
   targetType: T.oneOf([COMMENT, POST]).isRequired,
 };
 
 ModeratorModal.defaultProps = {
   target: null,
+  reports: null,
 };
 
 const selector = createSelector(
--- a/src/app/components/Post/PostDropdown/index.jsx
+++ b/src/app/components/Post/PostDropdown/index.jsx
@@ -9,6 +9,8 @@ import {
   DropdownLinkRow,
 } from 'app/components/Dropdown';
 
+import { ReportsModal } from 'app/components/ReportsModal';
+
 const T = React.PropTypes;
 
 export default function PostDropdown(props) {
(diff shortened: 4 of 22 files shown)


scarow:
👓 @madbook @birakattack @powerlanguage

src/app/components/ReportsModal/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+    isRemoved,
+    isSpam,
+    removedBy,
+    approvedBy,
+    reportModalId,
+    onClick,
+  } = props;
+
+  return (
+    <div className='ModeratorModalWrapper' onClick={ onClick }>
```
madbook: Why this className instead of `ReportsModal`?

src/app/components/ReportsModal/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+      <Modal
+        id={ reportModalId }
+        className='DropdownModal ModeratorModal'
+      >
+        <div className='DropdownClose'>
```
madbook: Similarly, this feels like it's breaking with our class naming conventions.  Something like `ReportsModal__dropdown-close` seems better.

src/app/components/ReportsModal/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+          <DropdownRow
+            icon='x'
+            onClick={ onClick }
+          />
+        </div>
+        <div className='ModeratorModalRowWrapper' onClick={ onClick }>
```
madbook: This also, maybe `ReportsModal__row-wrapper`

src/app/components/ReportsModal/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+        </div>
+      </Modal>
+    </div>
+  );
+}
+
+function showReports(reports, reportType) {
+  if (reports.length > 0) {
+    return (
+      <div className='Reports'>
```
madbook: This className should also be under the `ReportsModal` namespace, or broken out into a proper standalone component

src/lib/modToolHelpers.js
```diff
@@ -11,3 +11,27 @@ export function getApprovalStatus(approved, removed, spam) {
...
+  modReports.forEach(function(report) {
+    if (report[1] === username) {
+      reportReplaced = true;
+      report[0] = reason;
+    }
+  });
+
+  // if mod hasn't reported post yet, add report
+  if (!reportReplaced) {
+    modReports.push([reason, username]);
```
madbook: since this array comes directly out of state and this push modifies the array, i think this violates the functional purity of the reducers that use this helper.

src/lib/modToolHelpers.js
```diff
@@ -11,3 +11,27 @@ export function getApprovalStatus(approved, removed, spam) {
   return null;
 }
 
+export function sumReportsCount(userReports, modReports) {
+  let reportCounts = userReports.map(r => r[1]);
+  reportCounts.push(modReports.length);
+  return reportCounts.reduce((a, b) => a + b, 0);
+}
+
+export function getNewModReports(username, modReports, reason) {
```
madbook: super nitpicky, but the order of the arguments here seems a bit odd to me because of the type.  `getNewModReports(modReports, reason, username)` seems to make more sense to me (old state in first, order of other arguments matches the order they appear in the modReports array)

src/lib/modToolHelpers.js
```diff
@@ -11,3 +11,27 @@ export function getApprovalStatus(approved, removed, spam) {
...
+  let reportReplaced;
+
+  // if mod has already reported post, update reason
+  modReports.forEach(function(report) {
+    if (report[1] === username) {
+      reportReplaced = true;
+      report[0] = reason;
+    }
+  });
```
madbook: ❓ Do the order of the mod reports matter? If not, this could all be a lot easier if `modReports` was just a dictionary mapping username -> report reason.

src/app/reducers/posts.js
```diff
@@ -67,6 +70,24 @@ export default function(state=DEFAULT, action={}) {
...
+          {
+            model: model.set({
+              modReports: getNewModReports(username, modReports, reason),
+            }),
+          }
+        );
+      }
+
+      return state;
+    }
```
madbook: ❓ The duplication here makes me wonder if the reports state ought to be kept completely separately (i.e. in its own reducer)

src/app/components/Comment/CommentDropdown/index.jsx
```diff
@@ -104,6 +123,9 @@ CommentDropdown.propTypes = {
   modModalId: T.string,
   distinguishType: T.string,
   isMine: T.bool,
+  reportModalId: T.string,
+  userReports: T.arrayOf(T.arrayOf(T.oneOf(T.string, T.number))),
+  modReports: T.arrayOf(T.arrayOf(T.string, T.number)),
```
madbook: Why the different types between the two?

src/app/components/Comment/CommentHeader/index.jsx
```diff
@@ -172,4 +185,6 @@ CommentHeader.defaultProps = {
   isApproved: false,
   isRemoved: false,
   isSpam: false,
+  userReports: [],
+  modReports: [],
```
madbook: 🔕 we should think about how to clean this up so we don't need to define this stuff in so many places

src/app/components/ModeratorModal/index.jsx
```diff
@@ -149,13 +165,17 @@ ModeratorModal.propTypes = {
   removedBy: T.string,
   approvedBy: T.string,
   distinguishType: T.string,
+  modReports: T.arrayOf(T.arrayOf(T.string)),
   isMine: T.bool,
   target: T.object,
   targetType: T.oneOf([COMMENT, POST]).isRequired,
+  userReports: T.arrayOf(T.arrayOf(T.string)),
```
madbook: Why are the proptypes different here?

src/app/components/Post/PostDropdown/index.jsx
```diff
@@ -142,6 +162,8 @@ PostDropdown.propTypes = {
   removedBy: T.string,
   modModalId: T.string,
   showModModal: T.bool,
+  userReports: T.arrayOf(T.arrayOf(T.oneOf(T.string, T.number))),
+  modReports: T.arrayOf(T.arrayOf(T.oneOf(T.string, T.number))),
```
madbook: More proptype shenanigans, here they match

src/app/components/Post/PostDropdown/index.jsx
```diff
@@ -150,10 +172,12 @@ PostDropdown.defaultProps = {
   isSaved: false,
   isLoggedIn: false,
   isHidden: false,
+  modModalId: null,
```
madbook: defaultProp for modModalId is `null`, but for reportModalId it's an empty string – may as well make them both one or the other

src/app/components/ReportsModal/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+        <div className='m-reports-title'>{ `${reportType} Reports:` }</div>
+        {
+          reports.map(function(report) {
+            return (
+              <div>{ `${report[1]}: ${report[0]}` }</div>
+            );
+          })
```
madbook: this can be simplified to something like

```javascript
reports.map(([reason, label]) => (
  <div>{ `${label}: ${reason}` }</div>
))
```

src/app/components/ReportsModal/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+ReportsModal.defaultProps = {
+  userReports: [],
+  modReports: [],
+  isApproved: true,
+  isRemoved: false,
+  isSpam: false,
+  removedBy: null,
+  approvedBy: null,
+  reportModalId: null,
```
madbook: should probably default this to an empty string (empty string is the preferred default type for string props, and `reportModalId` is doing that elsewhere)

src/lib/modToolHelpers.js
```diff
@@ -11,3 +11,27 @@ export function getApprovalStatus(approved, removed, spam) {
   return null;
 }
 
+export function sumReportsCount(userReports, modReports) {
+  let reportCounts = userReports.map(r => r[1]);
+  reportCounts.push(modReports.length);
```
madbook: you could also jus tadd `modReports.length` to the result of the reduce below (to avoid an array mutation)

src/app/components/ReportsModal/index.jsx
scarow: Mostly for css reasons -- I can change the name and copy the css though if that's preferable

src/lib/modToolHelpers.js
scarow: It would be so much easier that way -- but that's not the way the data is structured when its returned from r2. I could restructure the data to look more like:

```
modReports = { 'toasties': 'Spam' }
userReports = { 'Spam': 2, 'Illegal Content': 4 }
```

I could do this in either the reducer or in node.

What do you think?

src/app/components/Comment/CommentDropdown/index.jsx
scarow: modReports look like this: `[['Spam', 'toasties']]`
userReports look like this: `[['Spam', 1], ['Illegal Content', 4]]`

src/app/components/ModeratorModal/index.jsx
scarow: My bad -- error on my part

src/app/components/ReportsModal/index.jsx
scarow: DropdownClose is used across a few different modals -- maybe I should move this to its own component actually

src/app/reducers/posts.js
scarow: What do you mean/why?

scarow:
:haircut_man:

src/app/components/ReportsModal/index.jsx
madbook: That would be 💯

src/lib/modToolHelpers.js
madbook: doing it in the reducer seems fine to me – having the same data structure between them seems like a good idea.  It seems weird to have the report reason be the key in one dict and the value in the other though (thought it makes sense in a way too)

Thinking about it a bit, what do you think about combining them into a single dict, essentially putting all the user reports under a single key.  Something like

```js
const USER_REPORT_KEY = new Symbol('USER_REPORTS');
reports = {
  [USER_REPORT_KEY]: {
    'Spam' : 2,
    'Illegal Content': 4,
  },
  'toasties': {
    'Spam': 1,
  },
};
```

src/app/reducers/posts.js
madbook: the reducer code here & in comments is pretty much identical, and reports are sort of a separate thing.  instead of having...

```js
state = {
  posts: {
    [postFullname]: {
      ...postProps,
      reports: { ... },
    },
  },
  comment: {
    [commentFullname]: {
      ...commentProps
      reports: { ... },
    },
  },
};
```

we could do

```js
state = {
  reports: {
    [postFullname]: { ... },
    [commentFullname]: { ... },
  },
  posts: {
    [postFullname]: {
      ...postProps,
    },
  },
  comment: {
    [commentFullname]: {
      ...commentProps
    },
  },
};
```

Then we could only pull that state in when we need it, and since the data structure for all kinds of reports are the same, they can all live under the same reducer and the logic to update state only has to live in one place.  It's also easy to extend support for other reportable things (like PMs and subreddits (yes, you can report subreddits!));

src/app/components/Comment/CommentDropdown/index.jsx
madbook: Should modReports be `T.arrayOf(T.arrayOf(T.string))` then?

scarow:
:haircut_man:

src/app/reducers/reports.js
```diff
@@ -0,0 +1,86 @@
...
+// instead of arrays
+const formatReports = (modelList) => {
+  let reports = {};
+
+  for (var key in modelList) {
+    reports[key] = {
+      MODERATOR_REPORT_KEY: {},
+      USER_REPORT_KEY: {},
+    };
+
```
scarow: I'm not sure how I feel about this -- this creates an empty reports object for every listing and comment on the page, because otherwise, I need to check

`reports && reports[key] && reports[key].USER_REPORT_KEY` everywhere.

Thoughts? Should we just have reports ONLY contain keys where there are reports available? This seems like the optimal solution but forces complicated logic everywhere else.

src/app/components/ModeratorModal/index.jsx
```diff
@@ -31,6 +33,12 @@ export class ModeratorModal extends React.Component {
     e.stopPropagation();
   }
 
+  hasReports(reports) {
+    if (!reports) { return false; }
+    return !!(Object.keys(reports.USER_REPORT_KEY).length > 0
```
madbook: i think you'll want this to be `reports[USER_REPORT_KEY]` to use the action key string imported from `app/reports/reducer` (same for `MODERATOR_REPORT_KEY` below)

src/app/components/ReportsModal/index.jsx
```diff
@@ -19,8 +19,12 @@ export function ReportsModal(props) {
     onClick,
   } = props;
 
+  if (!reports) { return null; }
+
   const status = getApprovalStatus(isApproved, isRemoved, isSpam);
   const statusBy = getStatusBy(isApproved, isRemoved, isSpam, removedBy, approvedBy);
+  const modReports = reports.MODERATOR_REPORT_KEY;
+  const userReports = reports.USER_REPORT_KEY;
```
madbook: same re `reports[USER_REPORT_KEY]`

src/app/components/ReportsModal/index.jsx
```diff
@@ -37,22 +41,53 @@ export function ReportsModal(props) {
...
+function getUserReports(reports) {
+  const userReports = reports.USER_REPORT_KEY;
+
+  return Object.keys(userReports).map(([reason, count]) => (
+    <div>{ `${count}: ${reason}` }</div>
+  ))
+}
+
+function getModReports(reports) {
+  const modReports = reports.MODERATOR_REPORT_KEY;
```
madbook: same re `reports[MODERATOR_REPORT_KEY]` here and above for `USER_REPORT_KEY`

src/app/reducers/reports.js
```diff
@@ -0,0 +1,86 @@
...
+// This function arranges modReports and userReports into an object
+// instead of arrays
+const formatReports = (modelList) => {
+  let reports = {};
+
+  for (var key in modelList) {
+    reports[key] = {
+      MODERATOR_REPORT_KEY: {},
+      USER_REPORT_KEY: {},
```
madbook: same here, if you want to use the exported keys this should be
```
  [MODERATOR_REPORT_KEY]: {},
  [USER_REPORT_KEY]: {},
```

An alternative would be to drop the exporting and just change these to something like
```
  moderator: {},
  user: {},
```
which I think is a fine thing to just hardcode and not worry about passing around a const representing the key

src/app/reducers/reports.js
madbook: I think you are right in that creating a reports data structure for every `thing` is wasteful, especially since it is only relevant to mods.  My suggesting is to let things w/o report data stay undefined in the reports state, and to wrap up that complicated check into a helper function that you can call in multiple places.

src/app/reducers/reports.js
```diff
@@ -0,0 +1,86 @@
...
+      const { model, reason, username } = action;
+      let reports = state;
+
+      // in case there is no report yet, set to empty hash
+      if (!reports[model.name]) {
+        reports[model.name] = {};
+        reports[model.name].MODERATOR_REPORT_KEY = {};
+      }
+
+      reports[model.name].MODERATOR_REPORT_KEY[username] = reason;
```
madbook: same dealio with the keys here

src/app/reducers/reports.js
```diff
@@ -0,0 +1,86 @@
...
+export default (state=DEFAULT, action={}) => {
+  switch (action.type) {
+    case loginActions.LOGGED_IN:
+    case loginActions.LOGGED_OUT: {
+      return DEFAULT;
+    }
+
+    case reportingActions.SUCCESS: {
+      const { model, reason, username } = action;
+      let reports = state;
```
madbook: isn't `state` here just the part of the state tree that this reducer operates on (i.e. `state.reports`?)

src/app/reducers/reports.js
madbook: should we be checking here if the user reporting is a moderator or not, and if not updating the user reports instead?

src/app/reducers/reports.js
```diff
@@ -0,0 +1,86 @@
...
+
+      // in case there is no report yet, set to empty hash
+      if (!reports[model.name]) {
+        reports[model.name] = {};
```
madbook: This is mutating! I think you'll want to do something like

```javascript
return merge(state, {
  [model.name]: {
    [MODERATOR_REPORT_KEY]: {
      [username]: reason,
    },
  },
});
```

I _think_ `merge` handles merging nested objects correctly in that case (not 100% on that, def test)

src/lib/modToolHelpers.js
```diff
@@ -11,26 +13,16 @@ export function getApprovalStatus(approved, removed, spam) {
...
+  let sum = Object.keys(reports.MODERATOR_REPORT_KEY).length || 0;
+  const userReports = reports.USER_REPORT_KEY || {};
 
-  // if mod hasn't reported post yet, add report
-  if (!reportReplaced) {
-    modReports.push([reason, username]);
-  }
+  Object.values(userReports).forEach(function(count){
+    sum += count;
+  })
```
madbook: nit: missing `;`

madbook: Another way to do this would be to use `reduce` with the count of mod reports as the starting value

```javascript
const numModReports = Object.keys(reports[MODERATOR_REPORT_KEY]).length;
return Object.values(userReports).reduce((a, b) => a + b, numModReports);

```

src/lib/modToolHelpers.js
```diff
@@ -11,26 +13,16 @@ export function getApprovalStatus(approved, removed, spam) {
...
+  if (!reports) { return null };
 
-  // if mod has already reported post, update reason
-  modReports.forEach(function(report) {
-    if (report[1] === username) {
-      reportReplaced = true;
-      report[0] = reason;
-    }
-  });
+  let sum = Object.keys(reports.MODERATOR_REPORT_KEY).length || 0;
```
madbook: the `|| 0` shouldn't be necessary, since an object with no keys should produce an array with length of `0`.

src/lib/modToolHelpers.js
```diff
@@ -11,26 +13,16 @@ export function getApprovalStatus(approved, removed, spam) {
...
 
-export function sumReportsCount(userReports, modReports) {
-  let reportCounts = userReports.map(r => r[1]);
-  return reportCounts.reduce((a, b) => a + b, 0) + modReports.length;
-}
-
-export function getNewModReports(modReports, reason, username) {
-  let reportReplaced;
+export function sumReportsCount(reports) {
+  if (!reports) { return null };
```
madbook: Seems like this should return `0`

src/app/reducers/reports.js
scarow: Since users can't see the metadata (and this is purely used to update metadata without a refresh), we don't need to add anything to `USER_REPORT_KEY` but we should check to see that the person reporting is not a moderator, just to make the code more explicit.

src/app/reducers/reports.js
scarow: Ah -- good idea!

scarow:
:haircut_man:

src/lib/modToolHelpers.js
```diff
@@ -14,14 +14,18 @@ export function getApprovalStatus(approved, removed, spam) {
...
-  const userReports = reports.USER_REPORT_KEY || {};
+  let sum = 0;
+  const userReports = reports[USER_REPORT_KEY] || {};
 
-  Object.values(userReports).forEach(function(count){
+  if (reports[MODERATOR_REPORT_KEY]) {
+    sum += Object.keys(reports[MODERATOR_REPORT_KEY]).length;
+  }
+
+  Object.values(userReports).forEach(function(count) {
```
madbook: Should be using a lambda (arrow function) here, though i still think a `reduce` would make more sense (and let you avoid mutating `sum`

```js
return Object.values(userReports).reduce((sum, count) => sum + count, countOfModReports);
```

scarow:
:haircut_man: thanks @madbook !!

PR #1005, opened 2017-02-14
Convert extraneous lodash imports to lodash-es.

schwers:
This removes some unneeded lodash code in the payload.
Before and After (from Chrome Network tab):
* 985KB         -> 967KB
* 255KB gzipped -> 250KB gzipped

👓  @curioussavage


curioussavage:
🐟

PR #1011, opened 2017-02-17
PostContent: render selftext using the selfTextHTML property on Post

curioussavage:
This fixes an issue with a new ads feature that allows embeds and selftext in the same post.
The embed has priority so it is set as the expandedContent property on the model and we were
relying on the expandedContent property to contain the self text.

👓  @schwers  

cc @zeantsoi


curioussavage:
so this will render the embed with the selftext below it just like in r2

PR #1014, opened 2017-02-22
Fix horizontal overflow caused by modtools icons

madbook:
for [CE-550](https://reddit.atlassian.net/browse/CE-550)

Seems like the flex attributes just weren't _quite_ right here.  These seem to work in the iOS simulator, and they make a bit more sense.

1. The `post-descriptor-line` element _only_ sets the `flex-grow` property to `1`.
2. The `metadata` element _only_ sets the `flex-basis` property to `auto`.

With these settings, the `metadata` element should only ever take up as much space on the line as it needs to given its contents (no growing, no shrinking), and the `post-descriptor-line` element should grow to take up the rest of the available space.

👓 @scarow

### DIFF
--- a/src/app/components/Post/PostHeader/styles.less
+++ b/src/app/components/Post/PostHeader/styles.less
@@ -74,7 +74,7 @@
     overflow: hidden;
     position: relative;
     vertical-align: middle;
-    flex: 1 1 100%;
+    flex: 1 0;
 
     .themeify({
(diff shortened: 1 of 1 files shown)


scarow:
🐟 bless up 🙏

madbook:
e87306971e9bdd267aa90e096b856d8e596750a9

PR #1010, opened 2017-02-16
Popular is popular

keithblaha:
:eyeglasses: @prashtx @schwers @nramadas 

looking for a quick review, would like to deploy today if possible

see https://reddit.atlassian.net/browse/HOT-117 for more details

PR #1015, opened 2017-02-22
removes redundant Popular link from CommunityOverlayMenu for logged out users

keithblaha:
:eyeglasses: @uzi @schwers

I recently submitted https://github.com/reddit/reddit-mobile/pull/1010, which added r/Popular to the CommunityOverlayMenu. However, @spladug reported that for logged out users this doesn't make sense since Front Page is already r/Popular for them, making them redundant. Relevance in coordination with Design decided the right approach is to only show Popular on the CommunityOverlayMenu for logged out users.

It also looks like there was a non-working prior attempt to say My Front Page instead of Front Page for logged in users. This PR removes that since it didn't work and the concern of this PR isn't around that (we also think it makes sense since across other platforms this distinction isn't made).


spladug:
thanks!

keithblaha:
thank you for the quick review!

PR #1017, opened 2017-02-23
Use subreddit rules as report reasons

madbook:
for [CE-499](https://reddit.atlassian.net/browse/CE-499)

That ticket is for adding support for the new `violation_reason` field to the report modal here.  However, the report modal isn't even currently using subreddit rules (or even the most recent reddit rules!) so the scope of this got a little bigger.

Not ready for review _just_ yet (I want to do a QA pass first).


madbook:
👓 @birakattack @scarow

src/app/components/ReportingModal/index.jsx
```diff
@@ -74,40 +101,60 @@ class ReportingModal extends React.Component {
...
 
 const dispatcher = dispatch => ({
   onClickOutside: () => dispatch(modalActions.closeModal()),
-  onSubmit: (thingId, reason) => dispatch(reportingActions.submit(thingId, reason)),
-});
-
-const mergeProps = (stateProps, dispatchProps) => ({
-  ...dispatchProps,
-  onSubmit: reason => dispatchProps.onSubmit(stateProps.thingId, reason),
+  onSubmit: (report) => dispatch(reportingActions.submit(report)),
```
scarow: nit: i believe this does not need parenthesis for a single parameter

scarow:
Are the rules fetched when you click on `report`? Or do they load when we fetch the subreddit?

scarow:
:haircut_man: just a question -- looks really good!

scarow:
💅

madbook:
@scarow the rules are fetched when you click `report` (unless they are already in state)

madbook:
:haircut_man: for the fixup and the question

scarow:
🐟  lgtm

src/app/models/Report.js
```diff
@@ -0,0 +1,30 @@
...
+export default class Report {
+  /**
+   * @constructor
+   * @param {string} thingId The id of the thing this report applies to.
+   * @param {string} reason The name of the rule violated
+   * @param {string} displayText The report text displayed to the user.
+   */
+  constructor(thingId, rule, reason) {
+    this.thingId = thingId;
+    this.rule = rule;
```
birakattack: nit: The comment says the parameter is `reason` but the actual parameter is `rule`. Also would `ruleName` be more descriptive here?

madbook:
didn't realize this actually broke the `modal` reducer test because the `report` action now expects the model of the thing you are reporting on to exist in state – 044c4bc fixes up the test, and f451e3c fixes the nit pointed out by @birakattack

:haircut_man:

PR #1021, opened 2017-02-25
Add XPromo comments page interstitial.

schwers:
Other notable improvements:
* xpromo_ineligible event should be more accurate on listing
    and index pages now.
* xpromo_dimiss and app_store_visit events should now accurately
    contain experiment info and screen_view data

👓  @tomsoir @prashtx


prashtx:
:fish:

PR #1025, opened 2017-02-27
Make use Anchors for continueThread

curioussavage:
👓 @schwers 

Just making it so you can open continue thread in a new tab.

### DIFF
--- a/src/app/components/CommentTree/index.jsx
+++ b/src/app/components/CommentTree/index.jsx
@@ -5,7 +5,7 @@ import './index.less';
 import React from 'react';
 import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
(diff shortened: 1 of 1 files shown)


src/app/components/CommentTree/index.jsx
```diff
@@ -136,7 +137,14 @@ const renderNode = (props, depth, data, isHidden) => {
 
             { ...returnDispatchers(commentDispatchers, uuid) }
           />
-        : renderContinueThread(() => onContinueThread(data), data, depth === 0, dotsNum) }
+        : renderContinueThread(
+            (e) => onLoadMore(data, e),
```
schwers: :nitpick: no parens for single args

PR #1028, opened 2017-02-28
Add .envrc to gitignore so devs can use solutions like direnv

curioussavage:
👓  @schwers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ build/
 .tox/
 .DS_Store
 reddit-mobile.log
+.envrc
+.env


curioussavage:
I guess I should add `.env` too?

.gitignore
```diff
@@ -10,3 +10,4 @@ build/
 .tox/
 .DS_Store
 reddit-mobile.log
+.envrc
```
schwers: Could you add `.env` as well

curioussavage:
:haircut_man:

PR #1030, opened 2017-03-01
Fix a few comments bugs from state and seo.

schwers:
Noteable changes:
  * Updates access to `state.comments` and `state.commentsPages`
  * Fix api model toJSON serialization that was causing some
      Javascript errors on full (seo) renders. `uuid` and `type`
      weren't being serialized properly. This could affect
      SEO crawlers that run JavaScript.
  * Fix `undefined` in href's on continue thread / more comments
      links
   * Fixes tests

### DIFF
--- a/src/apiClient/apiBase/Model.js
+++ b/src/apiClient/apiBase/Model.js
@@ -211,8 +211,8 @@ export default class Model {
         obj[key] = this[key];
       }
     });
-
-    obj.__type = this.type;
+    obj.uuid = this.uuid;
+    obj.type = this.type;
     return obj;
   }
 }
--- a/src/app/actions/commentsPage.js
+++ b/src/app/actions/commentsPage.js
@@ -55,7 +55,7 @@ export const visitedCommentsPage = (postId) => ({
 export const fetchCommentsPage = commentsPageParams => async (dispatch, getState) => {
   const state = getState();
   const commentsPageId = paramsToCommentsPageId(commentsPageParams);
-  const commentsPage = state.commentsPages[commentsPageId];
+  const commentsPage = state.commentsPages.api[commentsPageId];
 
   if (commentsPage) { return; }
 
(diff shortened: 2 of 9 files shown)


schwers:
👓  @curioussavage @wting

src/app/actions/commentsPage.js
```diff
@@ -55,7 +55,7 @@ export const visitedCommentsPage = (postId) => ({
 export const fetchCommentsPage = commentsPageParams => async (dispatch, getState) => {
   const state = getState();
   const commentsPageId = paramsToCommentsPageId(commentsPageParams);
-  const commentsPage = state.commentsPages[commentsPageId];
+  const commentsPage = state.commentsPages.api[commentsPageId];
```
wting: :no_bell: Wouldn't something like this just be permanently broken all the time?

test/lib/getContentIdFromState.test.js
```diff
@@ -30,8 +30,8 @@ describe('lib: getContentIdFromState', () => {
     set(sampleState, 'platform.currentPage.urlParams.subredditName', 'pics');
     set(sampleState, 'platform.currentPage.urlParams.postId', 'foo');
     set(sampleState, 'subreddits.pics.name', 't5_2qh0u');
-    set(sampleState, 'commentsPages.current', 'deadbeef');
-    set(sampleState, 'commentsPages.deadbeef.postId', 't3_foo');
+    set(sampleState, 'commentsPages.data.current', 'deadbeef');
+    set(sampleState, 'commentsPages..data.deadbeef.postId', 't3_foo');
```
wting: Is this supposed to be `..`?

schwers: Nope, good catch!

src/app/actions/commentsPage.js
schwers: Yup! `commentsPages` is just an object, so looking it up does't break anything. But a side-effect of this is comments pages are currently refetched every time, instead of staying cached

schwers:
You can find the details of this ticket at: CHAN-447

:haircut_man:

src/app/actions/commentsPage.js
curioussavage: I'm glad you caught this 👍

src/apiClient/apiBase/Model.js
```diff
@@ -211,8 +211,8 @@ export default class Model {
         obj[key] = this[key];
       }
     });
-
-    obj.__type = this.type;
+    obj.uuid = this.uuid;
```
curioussavage: so this is necessary because uuid is not in the PROPERTIES? 🔕

schwers: yup!

PR #1034, opened 2017-03-01
Handle recaptcha component unmounting before a promise resolves

curioussavage:
👓  @jackniu1 @schwers 

https://sentry.reddit.ue1.snooguts.net/sentry/mobile-web/issues/4440/

Looks like people are leaving the page before the promise resolves. I followed the suggestion here https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html for the fix.

### DIFF
--- a/src/app/components/ReCaptcha/index.jsx
+++ b/src/app/components/ReCaptcha/index.jsx
@@ -2,6 +2,8 @@ import React from 'react';
 import './styles.less';
 
 const T = React.PropTypes;
+const gRecaptchaUrl =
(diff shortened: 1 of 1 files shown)


jackniu1:
Could an extra `onloadCallback` resolve this? Following `https://github.com/appleboy/react-recaptcha/issues/181` (sorry for the bug introduced at first place!).

curioussavage:
Doesn't look like it. They are doing a different but kind similar thing. their component works differently but they still have a `componentWillUnmount` to destroy their timer that checks if its loaded every second.

curioussavage:
:haircut_man:

jackniu1:
🐟

PR #1037, opened 2017-03-03
Record more status codes

schwers:
We've seen an increase in 404's, and bug reports of 'not found' pages. 404's are now down in grafana after https://github.com/reddit/reddit-mobile/pull/1030 was merged. But there are continued reports of bad pages. 

Noteable fixes:
* listing and comments pages should now always dispatch `success` or `failed`, this should ensure we always update the status code properly.
* errorLogger middleware is updated to ensure we properly log comments page failures
* ads requests  aren't made on the server, and as failsafe the adblock check shortcircuits on the server. (this came up in testing SEO renders)
* client's sentryOnError is defaulted to a no-op for dev mode

👓 @curioussavage @prashtx @dwick

### DIFF
--- a/src/app/actions/commentsPage.js
+++ b/src/app/actions/commentsPage.js
@@ -69,11 +69,7 @@ export const fetchCommentsPage = commentsPageParams => async (dispatch, getState
     const apiResponse = await CommentsEndpoint.get(apiOptions, commentsPageParams);
     dispatch(received(commentsPageId, apiResponse));
   } catch (e) {
-    if (e instanceof ResponseError) {
-      dispatch(failed(commentsPageId, e));
-    } else {
-      throw e;
-    }
+    dispatch(failed(commentsPageId, e));
   }
 };
 
--- a/src/app/reduxMiddleware/errorLogger.js
+++ b/src/app/reduxMiddleware/errorLogger.js
@@ -118,6 +118,10 @@ const checkForSpecificErrors = action => {
       if (error) {
         return error;
       }
+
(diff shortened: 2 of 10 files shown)


src/app/actions/commentsPage.js
```diff
@@ -69,11 +69,7 @@ export const fetchCommentsPage = commentsPageParams => async (dispatch, getState
     const apiResponse = await CommentsEndpoint.get(apiOptions, commentsPageParams);
     dispatch(received(commentsPageId, apiResponse));
   } catch (e) {
-    if (e instanceof ResponseError) {
-      dispatch(failed(commentsPageId, e));
-    } else {
-      throw e;
-    }
+    dispatch(failed(commentsPageId, e));
```
curioussavage: You decided there was no need to throw again?

src/app/reduxMiddleware/errorLogger.js
```diff
@@ -118,6 +118,10 @@ const checkForSpecificErrors = action => {
       if (error) {
         return error;
       }
+
+      if (action.payload && action.payload.error) {
```
curioussavage: I see. This is why you don't need to throw. Nice.

schwers: Yup, exactly :) only dispatching simplifies things and ensures our status codes are proper

curioussavage:
🐟  can it go out today?

prashtx:
:fish:

PR #1038, opened 2017-03-06
Gracefully handle failed accounts and invalid loidCreated.

schwers:
There's a fairly frequent 'invalid time' exception being thrown in
`dispatchInitialUser`, that has the effect of returning no content
and issuing 404s. This seems to be often be associated with a
`FAILED_ACCOUNT` action on the redux stack for Googlebot.
So it's possible they're coming in with an invalid loidcreated cookie
that's causing `Date` to throw.

This patch:
* tracks account request failure status, and doesn't set loid
cookies when there was a failure.
* doesn't set loid cookies if the `me` account request failed
* adds try/catch w/logging when doing loidcreated conversion
* fixes a bug in reducers/loid that was using 'loid' and 'loidcreated'
from any recevied account (it should only be checking accounts.me)

👓  @prashtx @curioussavage

### DIFF
--- a/src/app/actions/accounts.js
+++ b/src/app/actions/accounts.js
@@ -1,6 +1,5 @@
 import { apiOptionsFromState } from 'lib/apiOptionsFromState';
 import AccountsEndpoint from 'apiClient/apis/AccountsEndpoint';
-import ResponseError from 'apiClient/errors/ResponseError';
 
 export const FETCHING_ACCOUNT = 'FETCHING_ACCOUNT';
 // options is for the accounts endpoint, its properties are
@@ -38,10 +37,6 @@ export const fetch = options => async (dispatch, getState) => {
     const apiResponse = await AccountsEndpoint.get(apiOptionsFromState(state), query);
     dispatch(received(options, apiResponse));
   } catch (e) {
-    if (e instanceof ResponseError) {
-      dispatch(failed(options, e));
-    } else {
-      throw e;
-    }
+    dispatch(failed(options, e));
   }
 };
--- a/src/app/reducers/accountRequests.test.js
+++ b/src/app/reducers/accountRequests.test.js
@@ -3,7 +3,7 @@ import accountRequests from './accountRequests';
 import * as accountActions from 'app/actions/accounts';
 import * as loginActions from 'app/actions/login';
 
-const REQUIRED_KEYS = ['id', 'loading' ];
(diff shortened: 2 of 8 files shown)


schwers:
:haircut_man:

prashtx:
💅

src/app/reducers/loid.js
```diff
@@ -1,22 +1,30 @@
-import pick from 'lodash/pick';
-
 import * as accountActions from 'app/actions/accounts';
 import * as loidActions from 'app/actions/loid';
 
-export const DEFAULT = { loid: '', loidCreated: '' };
+export const DEFAULT = { loid: '', loidCreated: ' ' };
```
schwers: good catch, I was trying to force a truty value for testing, I'll revert

schwers:
:haircut_man:

src/app/actions/accounts.js
```diff
@@ -38,10 +37,6 @@ export const fetch = options => async (dispatch, getState) => {
     const apiResponse = await AccountsEndpoint.get(apiOptionsFromState(state), query);
     dispatch(received(options, apiResponse));
   } catch (e) {
-    if (e instanceof ResponseError) {
-      dispatch(failed(options, e));
-    } else {
-      throw e;
-    }
+    dispatch(failed(options, e));
```
curioussavage: 🔕  Out of scope but are there more api requests where we need to change this behavior?

src/app/reducers/accountRequests.test.js
```diff
@@ -34,7 +34,12 @@ createTest({ reducers: { accountRequests } }, ({ getStore, expect }) => {
 
     describe('FETCHING_ACCOUNT', () => {
       it('should add an account optimistically', () => {
-        const ACCOUNT = { id: 'foobar', loading: true };
+        const ACCOUNT = {
```
curioussavage: 🔕  might be more concise to export `newAccountRequest` and use it here.

src/app/actions/accounts.js
schwers: Yes, I think it would be worthwhile to go through them and do this, potentially re-thinking our error logging strategy

curioussavage:
lgtm. So I know you said this looks like its happening to the Google bot, but If a human encounters the same issue - it will just log them out? 

oh, also 🐟

PR #1041, opened 2017-03-08
Rename confusing Modal component

madbook:
There are two upcoming mweb experiments that use a modal, and while starting to work on those I came across this handy comment: 

```javascript
// TODO: this was incorrectly named, and should be renamed to reflect the fact		
// that is is a full page and has a url associated with it. Modals are		
// components designed not to interrupt the main workflow, but this IS the		
// main workflow so it's not a modal.
```

Since there's another component called `Modal` floating around (from widgets), this is extra confusing, so I figured I'd take a crack at renaming it since it appears to only be used in two places.

I couldn't think of a great name TBH, but picked `DismissiblePage` because it's at least more accurate than `Modal`.  If either of you can think of something better I'd be happy to change it.

There should be no functional changes here, just renaming stuff.

👓 @phil303 @scarow


madbook:
:haircut_man: bump :P

prashtx:
Drive-by naming suggestion: `Interstitial` or `DismissableInterstitial`?

phil303:
DismissablePage feels slightly more appropriate. Interstitials for me have the connotation of being non-interactable and having an ad type component in them.  Don't feel strongly here obviously.

The code looks good though and any name is going to be better than Modal :fish:

scarow:
🐟 lgtm

PR #1042, opened 2017-03-08
Convert epoch millisecond loidCreated

schwers:
👓  @prashtx I realized that some users will have these set from fastly, this is an intermediate measure to convert to ISO


prashtx:
the `let` can be a `const`. with that, :fish:

PR #1043, opened 2017-03-09
Chan 338 prevent xpromo

tomsoir:
💇 @schwers


src/app/actions/xpromoClientOnly.js
```diff
@@ -8,9 +8,10 @@ import { show } from './xpromo';
 // need access to the smartBanner actions. So we separate the checkAndSet bound
 // action creator (which has a branch dependency) from the other smartBanner
 // action code, so we can confine branch-sdk to the client.
-export const checkAndSet = () => async (dispatch) => {
+export const checkAndSet = (client) => async (dispatch) => {
```
schwers: Redux function actions are passed `dispatch`, `getState`, and the `utils` object ([docs](https://github.com/reddit/node-middleware#thunker)). You can take advantage of that and define `checkAndSet` to be 

```js 
export const checkAndSet = () => async (dispatch, getState) => { ... }
```

PR #1044, opened 2017-03-09
Chan 477 frequency interstitial

tomsoir:
Task: https://reddit.atlassian.net/browse/CHAN-477


src/app/featureFlags.js
```diff
@@ -189,6 +193,50 @@ const config = {
       ] },
     ],
   },
+  [VARIANT_XPROMO_INTERSTITIAL_FREQUENCY_IOS]: {
+    and: [
+      { allowedDevices: [IPHONE] },
+      { allowedPages: ['comments'] },
```
schwers: Should this be `['index', 'listing', 'comments']`?

I think it might be a good idea to remove the `allowedPages` rule altogether. In theory, you could start on a page that isn't a listing or comments page, and _then_ navigate to one of those pages. In that case we would have wanted to put you in the frequency experiment, but wouldn't.

tomsoir: Yes, you are right this experiment should goes for the all pages (asked Jeremy).

tomsoir:
:haircut: @schwers

src/lib/smartBannerState.js
```diff
@@ -129,6 +129,21 @@ export function getXPromoLink(state, path, linkType, additionalData={}) {
   });
 }
 
+function getClosingTimeRange(state) {
+  const defaultRange = frequency[EVERY_TWO_WEEKS];
+  const experimentData = currentExperimentData(state);
```
schwers: Should this be explicitly checking for the frequency experiments? I think the login required experiment is taking priority as is.

tomsoir:
💇 @schwers

src/lib/smartBannerState.js
tomsoir: Fixed. Thank you.

src/lib/smartBannerState.js
```diff
@@ -207,19 +217,7 @@ export function shouldNotListingClick(state) {
...
-
-  // Check if there's been a listing click in the last two weeks
-  const lastClickedStr = localStorage.getItem(XPROMO_LAST_LISTING_CLICK_DATE);
-  const lastClicked = lastClickedStr ? new Date(lastClickedStr).getTime() : 0;
-  if (lastClicked + TWO_WEEKS > Date.now()) {
-    return 'dismissed_previously';
-  }
-
-  return false;
+  return shouldNotShowBanner(state);
```
schwers: The listing click experiment uses a different local storage key, and has its own frequency experiment. Don't change anything in `shouldNotListingClick` for now.

tomsoir: Left this function. Let's fix it up after

tomsoir: Double checked the Listing and Frequency experiments.  Works fine.

tomsoir:
💇 @schwers

PR #1047, opened 2017-03-10
Support comments page paths for profile posts

prashtx:
We need to support comments pages that start with `/user/:username/`. We
also need to support subreddits that may have a different visible naming
convention than what appears in the URL for traditional `/r/:subreddit/`
paths.


prashtx:
👓 @uzi

uzi:
:fish:

prashtx:
Merged 3d8d7b9fee42279aeb214f1524e15edb0395db2d

PR #1048, opened 2017-03-13
Use the prefixed subreddit display name

prashtx:
Display u/username for profile posts and r/subredditname for subreddit
posts.

Let the sort selector use the correct base URL for a user's posts
listing.

### DIFF
--- a/src/app/components/Post/PostHeader/index.jsx
+++ b/src/app/components/Post/PostHeader/index.jsx
@@ -110,16 +110,14 @@ function subredditLabelIfNeeded(sr_detail, subreddit, hideSubredditLabel,
     }
   }
 
-  const rSubreddit = `r/${subreddit}`;
-
   return (
     <Anchor
       className='PostHeader__subreddit-link'
-      href={ `/${rSubreddit}` }
+      href={ sr_detail.url }
       style={ keyColorStyle }
       onClick={ e => interceptListingClick(e, LISTING_CLICK_TYPES.SUBREDDIT) }
     >
-      { rSubreddit }
+      { sr_detail.display_name_prefixed }
     </Anchor>
   );
 }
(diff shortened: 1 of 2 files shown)


prashtx:
👓 @uzi @schwers

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -281,12 +279,12 @@ function renderPostDescriptor(
...
 ) {
   const {
-    sr_detail,
+    subredditDetail,
     subreddit,
   } = post;
 
   const postFlairOrNil = renderPostFlair(post, single);
-  const subredditLabelOrNil = subredditLabelIfNeeded(sr_detail, subreddit,
+  const subredditLabelOrNil = subredditLabelIfNeeded(subredditDetail, subreddit,
```
uzi: 🔕  Seems like these changes aren't really needed, but I'm not gonna fight over it.

prashtx: The actual field in the Post model is called `subredditDetail`, so before this change, we were just passing along `undefined`. Seemed more straightforward than using something like `const { subredditDetail: sr_detail } = post;`, just to pass it along as an unnamed parameter.

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -110,16 +110,14 @@ function subredditLabelIfNeeded(sr_detail, subreddit, hideSubredditLabel,
...
   return (
     <Anchor
       className='PostHeader__subreddit-link'
-      href={ `/${rSubreddit}` }
+      href={ sr_detail.url }
       style={ keyColorStyle }
       onClick={ e => interceptListingClick(e, LISTING_CLICK_TYPES.SUBREDDIT) }
     >
-      { rSubreddit }
+      { sr_detail.display_name_prefixed }
```
schwers: 🔕 We might want to start using `post.subreddit_name_prefixed` at some point. It would be nice if we didn't have to always ask for the subreddit details, there's a lot of extra data there.

PR #1049, opened 2017-03-14
Use subreddit details only when available

prashtx:
We recently started using the post model's subredditDetail object to
generate subreddit/profile links and labels for PostHeader. If the post
doesn't have a subredditDetail object, we fall back to the prior method
of using the subreddit name that gets passed in.

Include the `sr_detail=true` parametery on a couple of more API
requests, so that we have access to the subreddit details when possible.


prashtx:
👓  @uzi @schwers

uzi:
:fish:

PR #1050, opened 2017-03-14
Support new loid cookie format.

schwers:
👓  @pacejackson @phil303 @prashtx

### DIFF
--- a/src/server/initialState/dispatchInitialUser.js
+++ b/src/server/initialState/dispatchInitialUser.js
@@ -1,17 +1,32 @@
+import setCookieParser from 'set-cookie-parser';
+
 import * as userActions from 'app/actions/user';
 import * as loidActions from 'app/actions/loid';
 import { permanentRootCookieOptions } from './permanentRootCookieOptions';
-import { logServerError } from 'lib/errorLog';
 
 export const dispatchInitialUser = async (ctx, dispatch, getState) => {
   // the lack of camel casing on the 'loidcreated' cookie name is intentional.
-  const oldLoid = ctx.cookies.get('loid');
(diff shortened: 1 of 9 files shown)


src/server/initialState/dispatchInitialUser.js
```diff
@@ -1,17 +1,34 @@
+import setCookieParser from 'set-cookie-parser';
+
 import * as userActions from 'app/actions/user';
 import * as loidActions from 'app/actions/loid';
 import { permanentRootCookieOptions } from './permanentRootCookieOptions';
-import { logServerError } from 'lib/errorLog';
+
+const LOID_COOKIE_REGEXP = /^[a-zA-Z0-9]+\.\d+\.\d+\.[a-zA-Z0-9+=/]+$/
```
phil303: That last bit for the base64 encoding regex makes me a bit nervous. From a quick google, it _seems_ like it will cover all characters. @pacejackson opinions? 

It seems like we could just do what @nramadas did in his patch [here|https://github.com/reddit/reddit-desktop/pull/123/files#diff-ed6360bfced43935f7c3f1629c240c15R13] but maybe that's _too_ lose

pacejackson: That should be fine, it's the same regex that we have in the r2 change.

phil303:
:fish:

pacejackson:
Hmm, I think it might be better to just do what @nramadas did... I noticed that _something_ is url-encoding these cookies and I'm not sure what so there is the chance that the cookie won't match that regex.

schwers:
💇

pacejackson:
🐟

PR #1056, opened 2017-03-16
Add the follow/unfollow functionality for contributors

uzi:
I added a bit of extra data that will be useful to the
Account and Subreddit models, changed the loading behavior
so the top doesn't jump, and styled things as they
currently are.  Cosmetic changes will come in subsequent
work, but I at least wanted to get the functionality in.

👓 @schwers @nramadas @prashtx


prashtx:
talked about a couple of small changes IRL and missing semicolons

💅

uzi:
💇  Fixed!

prashtx:
:fish:

PR #1057, opened 2017-03-16
Add a verified badge to contributor usernames

prashtx:
Add a verified badge next to contributor usernames on the profile views and next to the profile link on profile posts. For profile posts, don't show the light-gray username (since we already indicate the u/username in the "subreddit" link).

We had some inconsistent usage of lowercased username, true-case username, and whatever casing was used in the URL. This caused problems when trying to find data in the state for usernames that are not lowercase or for situations where the casing in the URL doesn't match the actual username spelling. This change standardizes on using a lowercase account `uuid` field. Then we can always convert the username from the URL to lowercase, and we use the `uuid` to connect among the current user, the fetched accounts, and the profile subreddit.

There may be a bit more we can do to standardize the `Account.uuid` usage, but for now we seem to be accessing things correctly.


MelissaCole:
I don't think verified should be tied to contributor status. I'm passing in 'verified' in the account object (which currently is true if it's a contributor but that will change when we open it up to everyone) - would that work? I could also see a case for being verified and not having a user sr.

prashtx:
Yes, that's a good catch. Late-night coding has its drawbacks.

prashtx:
👓 @uzi @schwers @MelissaCole 

I think Uzi's patch will require me to resolve some merge conflicts, but I don't think it will impact the functional changes.

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -109,6 +109,7 @@ function subredditLabelIfNeeded(sr_detail, subreddit, hideSubredditLabel, interc
     }
   }
 
+  const profilePost = sr_detail && (sr_detail.subreddit_type === 'user');
```
schwers: 🔕  Would there be any downside to using the `subreddit_type` and `subreddit_name_prefixed` attributes on the post itself? It would be nice if we could remove our dependancy on requesting the sr_detail. An unscientific look at `/hot.json` shows adding `?sr_details=1` increases the over-the-wire size by ~3X

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -120,6 +121,7 @@ function subredditLabelIfNeeded(sr_detail, subreddit, hideSubredditLabel, interc
       onClick={ e => interceptListingClick(e, LISTING_CLICK_TYPES.SUBREDDIT) }
     >
       { displayLabel }
+      { profilePost && <span className="icon icon-verified lime"></span> }
```
schwers: :nitpick: This should be self-closing to match the style guide

src/app/components/Post/PostHeader/index.jsx
```diff
@@ -156,7 +161,7 @@ function renderAuthorAndTimeStamp(post, interceptListingClick, single, hideWhen,
     return authorLink;
   }
 
-  if (single) {
+  if (single && !profilePost) {
```
schwers: If the `profilePost` case was handled first, this could be simplified to 

```js
if (profilePost) {
`...`
} else if (single) {
`...`
}

`...`
```

src/app/components/UserProfileHeader/index.jsx
```diff
@@ -20,13 +20,17 @@ UserProfileHeader.propTypes = {
...
 const UserProfileBanner = props => {
-  const { userName } = props;
+  const { userName, isVerified } = props;
   return (
     <div className='UserProfileHeader__banner'>
-      <h3 className='UserProfileHeader__banner-user-name'>{ userName }</h3>
+      <h3 className='UserProfileHeader__banner-user-name'>
+        { userName }
+      </h3>
+      { isVerified && <div className='UserProfileHeader__verified icon icon-verified lime'></div> }
```
schwers: :nitpick: This should be self-closing to match the style guide

src/app/pages/UserActivity/index.jsx
```diff
@@ -15,22 +15,25 @@ import { paramsToActiviesRequestId } from 'app/models/ActivitiesRequest';
 const mapStateToProps = createSelector(
   userAccountSelector,
   state => state.activitiesRequests,
+  (state, props) => state.accounts[props.urlParams.userName.toLowerCase()],
```
schwers: In general we should aim for selectors that return what's ultimately being sent to the component. Partially to avoid re-renders, but more so for clarity. Could you move the `isVerified` logic here?

src/app/pages/UserProfile/index.jsx
```diff
@@ -14,21 +14,33 @@ const GILD_URL_RE = /u\/.*\/gild$/;
 
 const mapStateToProps = createSelector(
   userAccountSelector,
-  (state, props) => state.accounts[props.urlParams.userName],
+  (state, props) => state.accounts[props.urlParams.userName.toLowerCase()],
   (state, props) => state.accountRequests[props.urlParams.userName],
```
schwers: Should this use the lowercase name?

prashtx: The account request is issued based on the casing in the URL. Once it completes, I think we're good, since we then merge in the model based on its uuid, independent of the name assigned to the request that fetched that model. So rather than start tearing up the accountRequests references, I opted to leave that functionality as is.

src/app/pages/UserActivity/index.jsx
prashtx: Yes, that makes sense after removing the reliance on the subreddits. But I'm rebasing onto a recent patch that does start to make use of the `queriedUser` and `queriedUserRequest` objects, so I think it's moot now.

src/app/components/Post/PostHeader/index.jsx
prashtx: :+1:

src/app/components/UserProfileHeader/index.jsx
prashtx: :+1:

src/app/components/Post/PostHeader/index.jsx
prashtx: :+1:

prashtx:
:haircut_man:

uzi:
:fish:

prashtx:
Merged 0de49a1e73a70e332402b4943de019a1c92017f9

PR #1059, opened 2017-03-16
Fix content-clipping bug in firefox

schwers:
👓  @curioussavage @tomsoir 

This fixes a bug in firefox, where the dual part interstitials duplicate listing is not being clipped. This results in a second listing being rendered without a background, offset from the pages primary listing. 

e.g.
image

[source](https://www.reddit.com/r/softwaregore/comments/5zinoy/great_new_mobile_reddit_feature_get_twice_the/)

### DIFF
--- a/src/app/components/DualPartInterstitial/styles.less
+++ b/src/app/components/DualPartInterstitial/styles.less
@@ -28,6 +28,7 @@
   &__content {
     width: 100%;
     height: 100vh;
+    overflow: hidden;
   }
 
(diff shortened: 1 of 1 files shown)


tomsoir:
@schwers  :fish: Confirmed this bug in FF (UserAgent: Mozilla/5.0 (Android 4.4; Mobile; rv:41.0) Gecko/41.0 Firefox/41.0). Thank you.

PR #1062, opened 2017-03-17
Pass the edgebucket cookie along with API requests.

prashtx:
We need the API to have any edgebucket cookie we may have received
upstream, which it will use along with the loid cookie, if present.

### DIFF
--- a/src/app/reducers/loid.js
+++ b/src/app/reducers/loid.js
@@ -8,6 +8,7 @@ export const DEFAULT = {
   loidCreated: '',
   loidCookie: '',
   loidCreatedCookie: '',
+  edgeBucket: '',
 };
 
 export default (state=DEFAULT, action={}) => {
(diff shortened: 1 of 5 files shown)


src/app/reducers/loid.js
```diff
@@ -50,6 +51,13 @@ export default (state=DEFAULT, action={}) => {
       return state;
     }
 
+    case loidActions.SET_EDGE_BUCKET: {
+      const { edgeBucket } = action;
+
+      if (!edgeBucket) { return DEFAULT; }
```
schwers: Should this return `state`?

src/app/reducers/loid.js
```diff
@@ -50,6 +51,13 @@ export default (state=DEFAULT, action={}) => {
       return state;
     }
 
+    case loidActions.SET_EDGE_BUCKET: {
```
schwers: We should add a test for this case

prashtx: :+1:

prashtx:
:haircut_man:

prashtx:
:haircut_man: bump

prashtx:
👓 @uzi

uzi:
:fish:

PR #1063, opened 2017-03-17
Use author.toLowerCase when comparing against logged in username.

schwers:
A recent patch (c249ea6fe21efc5630b4ae8f01e5000cb73d083b),
changed account uuids to be lowercased, to fix some
issues when comparing the url name to the received account name.
As a side-effect this disabled editing and deleting of comments
for some accounts, and a few other edge cases.
This patch simply calls toLowerCase for those comparisons.
Adding to lowercase in the models would break username display.

👓  @madbook @prashtx

### DIFF
--- a/src/app/components/Comment/CommentDropdown/index.jsx
+++ b/src/app/components/Comment/CommentDropdown/index.jsx
@@ -35,7 +35,7 @@ export default function CommentDropdown(props) {
     reports,
   } = props;
 
-  const userIsAuthor = commentAuthor === username;
+  const userIsAuthor = commentAuthor.toLowerCase() === username;
 
   const modalContent = [
     userIsAuthor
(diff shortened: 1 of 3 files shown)


schwers:
Agreed, I was thinking of adding an `isSameUserName` function that compared the lowercase values of two input users names. But from a naming perspective is not obvious why you can't compare them directly yourself

prashtx:
Can we enrich the post/comment models to have an `authorUuid` field, which would then match the Account model's `uuid` field?

prashtx:
💅

schwers:
@prashtx I like that idea, I think that's the correct longer term solution. I'm happy to make a followup patch with that

prashtx:
:fish:

PR #1064, opened 2017-03-18
Various style fixes for profile updates

uzi:
* Restyle follow/subscribe buttons
* Display more things by their displayNamePrefixed format
* Display names with canonical casing
* Other little cleanups

👓  @prashtx @nramadas @schwers

### DIFF
--- a/src/app/components/CommunityHeader/index.jsx
+++ b/src/app/components/CommunityHeader/index.jsx
@@ -67,20 +67,17 @@ const renderBannerRow = (subreddit, theme) => {
   );
 };
 
-const followIconClass = subscriber => {
-  return subscriber ? 'icon-check-circled lime' : 'icon-follow blue';
+const renderSubscribeButton = subscriber => {
+  const subscribeClass = subscriber ? 'subscriber' : 'subscribe';
+  return (
+    <button
+      type='submit'
+      className={ `CommunityHeader-${subscribeClass}-button` }
+    >
+      { ` ${subscriber ? 'Subscribed' : 'Subscribe'} ` }
+    </button>
+  );
 };
-
-const renderSubscribeButton = subscriber => (
-  <button type='submit' className='CommunityHeader-text-row-blue CommunityHeader-no-outline'>
-    { ` ${subscriber ? 'Subscribed' : 'Subscribe'} ` }
-    <span className='CommunityHeader-subscribe-button' >
-      <span
-        className={ `CommunityHeader-subscribe-icon icon ${followIconClass(subscriber)}` }
-      />
-    </span>
-  </button>
-);
 
 const CommunityHeader = props => {
   const {
@@ -111,13 +108,12 @@ const CommunityHeader = props => {
       { banner }
       <div className='CommunityHeader-text-row'>
         <h4 className='CommunityHeader-community-title'>
-          { subreddit.displayName }
+          { subreddit.displayNamePrefixed }
         </h4>
       </div>
       <div className='CommunityHeader-text-row'>
         <span>{ `${formatNumber(subreddit.subscribers)} subscribers` }</span>
-      { onlineCount }
-        { ` ${UTF8Circle}` }
+        { onlineCount }
         <SubredditSubscribeForm
           subredditName={ subreddit.uuid }
           className='CommunityHeader-subscribe-form CommunityHeader-no-outline'
--- a/src/app/components/CommunityOverlayMenu/index.jsx
+++ b/src/app/components/CommunityOverlayMenu/index.jsx
@@ -31,15 +31,17 @@ const renderSubscriptions = (subscriptions, loading, theme) => (
     text='Subscribed'
     subtext={ numCommunitiesText(subscriptions, loading) }
   >
-    { map(subscriptions, subreddit => (
+    { map(subscriptions, subreddit => {
+      const [ subType, subName ] = subreddit.displayNamePrefixed.split('/');
+      return (
         <LinkRow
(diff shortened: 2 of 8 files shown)


src/app/components/CommunityHeader/index.jsx
```diff
@@ -67,18 +67,12 @@ const renderBannerRow = (subreddit, theme) => {
...
 
-const followIconClass = subscriber => {
-  return subscriber ? 'icon-check-circled lime' : 'icon-follow blue';
-};
-
 const renderSubscribeButton = subscriber => (
-  <button type='submit' className='CommunityHeader-text-row-blue CommunityHeader-no-outline'>
+  <button
+    type='submit'
+    className={ `CommunityHeader-subscribe${ subscriber ? 'r' : '' }-button` }
```
prashtx: nit: I think it's a little clearer to have a condition that selects between `CommunityHeader-subscribe-button` and `CommunityHeader-subscriber-button`, rather than selecting between `'r'` and `''`. It also avoid generating some garbage, since we'll just have two string literals.

uzi: fffffffffffine...

src/app/components/CommunityOverlayMenu/index.jsx
```diff
@@ -38,8 +38,13 @@ const renderSubscriptions = (subscriptions, loading, theme) => (
...
           text={ (
             <span>
-              <span className='CommunityRow__rSlash'>r/</span>
-              { subreddit.displayName }
+              <span className='CommunityRow__rSlash'>
+                { subreddit.subredditType === 'user' ? 'u/' : 'r/' }
+              </span>
+              { subreddit.subredditType === 'user'
+                ? subreddit.displayName.substring(2)
+                : subreddit.displayName
```
prashtx: I think it's clearer here to use `subreddit.displayNamePrefixed` (which clearly indicates that there's some form of prefix in the string) and then pulling out the first two characters (`str.substring(0,2)`) and the remainder (`str.substring(2)`). Then we don't even need to condition on the subreddit type. Thoughts?

src/app/components/CommunityRow/index.js
```diff
@@ -30,13 +30,22 @@ const renderIcon = (iconUrl, url, color, theme) => {
...
 
   return (
     <Anchor className='CommunityRow__details' href={ url }>
       <div className='CommunityRow__name'>
-        <span className='CommunityRow__rSlash'>r/</span>
-        { displayName }
+        <span className='CommunityRow__rSlash'>
+          { subredditType === 'user' ? 'u/' : 'r/' }
+        </span>
+        { subredditType === 'user' ? displayName.substring(2) : displayName }
```
prashtx: Same thought as in CommunityOverlayMenu.

src/app/components/UserProfileHeader/index.jsx
```diff
@@ -11,18 +11,9 @@ import SubredditSubscribeForm from 'app/components/SubredditSubscribeForm';
 
 const T = React.PropTypes;
 
-const followIconClass = follower => {
-  return follower ? 'icon-check-circled lime' : 'icon-follow blue';
-};
-
 const renderFollowButton = follower => (
-  <button type='submit' className='UserProfileHeader__text-row-blue UserProfileHeader__no-outline'>
+  <button type='submit' className={ `UserProfileHeader__follow${follower ? 'er' : '' }` }>
```
prashtx: Same nit as in CommunityHeader above.

src/app/pages/UserActivity/index.jsx
```diff
@@ -14,7 +14,7 @@ import { paramsToActiviesRequestId } from 'app/models/ActivitiesRequest';
 
 const mapStateToProps = createSelector(
   userAccountSelector,
-  (state, props) => state.accounts[props.urlParams.userName.toLowerCase()],
+  (state, props) => state.accounts[props.urlParams.userName.toLowerCase()] || {},
```
prashtx: Do we use reference equality to determine if these selectors have changed value? If so, I think the `{}` will yield a different object after unrelated state changes. I think we can avoid that by returning `myUser || {}` in the combiner function, though it feels like passing an empty object will mask potential bugs where we should know that we don't yet have user data.

src/app/pages/UserProfile/index.jsx
```diff
@@ -14,7 +14,7 @@ const GILD_URL_RE = /u\/.*\/gild$/;
 
 const mapStateToProps = createSelector(
   userAccountSelector,
-  (state, props) => state.accounts[props.urlParams.userName.toLowerCase()],
+  (state, props) => state.accounts[props.urlParams.userName.toLowerCase()] || {},
```
prashtx: Same thought as in UserActivity.

prashtx:
A couple of small thoughts and nits. Also, the build is failing on account of unused variables, I believe.

💅

src/app/pages/UserActivity/index.jsx
uzi: Fair -- fixed

src/app/pages/UserProfile/index.jsx
uzi: Fixing here, too

src/app/components/CommunityOverlayMenu/index.jsx
uzi: Agreed -- except I think I'll split on `/`

uzi:
💇 See if this is more to your liking

prashtx:
:fish:

PR #1069, opened 2017-03-22
Fix linter errors in CommunityOverlayMenu

madbook:
Looks like 0242da273fa20fe6a2f491590a308a0b3e581e7a introduced some linter errors that are causing new builds to fail linting (e.g. https://travis-ci.org/reddit/reddit-mobile/jobs/213859601)

👓 @luisatlive


luisatlive:
:fish:

My fault, thought we had clean builds but I must've missed that on the second iteration to fix the launch day issue. Will double check offline to make sure that's the case.

PR #1070, opened 2017-03-22
Fixing of "current" key-param in a Comments page

tomsoir:
The problem appears after navigation from the comments page (to index page,
for example). In this case, when you request a subreddit name, it will be
returned from the last visited comment page.

### DIFF
--- a/src/app/pages/UserActivity/index.jsx
+++ b/src/app/pages/UserActivity/index.jsx
@@ -65,9 +65,7 @@ export const UserActivityPage = connect(mapStateToProps)(props => {
       <PostAndCommentList
         requestLocation='activitiesRequests'
         requestId={ activitiesId }
-        thingProps={ {
-          userActivityPage: true,
-        } }
+        thingProps={ {userActivityPage: true} }
       />
     </div>
   );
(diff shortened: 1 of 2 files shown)


src/app/reducers/commentsPages/data/index.js
```diff
@@ -21,6 +23,9 @@ const DEFAULT = {};
  **/
 export default (state=DEFAULT, action={}) => {
   switch (action.type) {
+    case platformActions.SET_PAGE: {
```
schwers: After thinking about this a bit more, I think this isn't a complete fix. `state.commentsPage.data.current` will be `null` if you visit a comments page, go to any other page on mweb, then hit 'back'.  So we could try one of two things:

1. Instead of clearing `current` on page changes, `getSubredditFromState` could check if there's a `postId` in the url params. This would be similar to `getContentIdFromState` and is less code changes overall.
2. On every `NAVIGATE_TO_URL`. `SET_PAGE` and `GOTO_PAGE_INDEX` action, recompute the current comments page id. If we're now on a comments page, it should be that pages id, otherwise it should be null. Then, the `current` setting could be removed from the `RECEIVED_COMMENTS_PAGE` and `MORE_COMMENTS_SUCCESS` cases.

(1) is less code changes and easier to test, but (2) makes the concept behind `current` more obvious and understandable imo.

PR #1073, opened 2017-03-23
Use yarn for managing dependencies 

schwers:


schwers:
👓  @prashtx @griffinmichl 

cc: @ckwang8128

griffinmichl:
🐟 LGTM. We should probably put a docker image with yarn pre-installed on quay at some point. It'll make our CI builds faster to not have to install yarn every time.

PR #1074, opened 2017-03-25
Search sort

mikeparas:
Currently, when viewing a search page, if you select a sort option, you get redirected back to the main subreddit view. This change will look for "/search" in the URL, and, if found, the sort query will be added allowing for a sortable search.

(This is my first time working with Redux, so any feedback is welcome. I was initially trying to pass in an onSortChange function to the SortAndTimeSelector in the search page component. I wasn't quite sure how to set everything up, so I ended up just putting this block here.)


mikeparas:
Didn't realize there was a conflict.

mikeparas:
...Any feedback?

PR #1075, opened 2017-03-27
Fix NSFW page check in app/featureFlags

schwers:
👓  @tomsoir @prashtx 

I think this would be good to add tests for, but I wanted to get the patch in review for now.

### DIFF
--- a/src/app/featureFlags.js
+++ b/src/app/featureFlags.js
@@ -572,7 +572,7 @@ export const isNSFWPage = state => {
   }
 
   if (subredditName) {
-    const subreddit = state.subreddits[subredditName];
+    const subreddit = state.subreddits[subredditName.toLowerCase()];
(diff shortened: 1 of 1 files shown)


prashtx:
Good catch.

:fish:

src/app/featureFlags.js
```diff
@@ -572,7 +572,7 @@ export const isNSFWPage = state => {
   }
 
   if (subredditName) {
-    const subreddit = state.subreddits[subredditName];
+    const subreddit = state.subreddits[subredditName.toLowerCase()];
```
schwers: Agreed, this might be better served by adding a `subredditModelFromState` function. (I think in a few of those cases those lookups use lowercased names that are passed to the function. But definetly not all, and very easy to break)

PR #1076, opened 2017-03-27
Xpromo refactors

schwers:
This is a relatively large refactor of how we handle some xpromo stuff.

Each commit's descriptions will have a more thorough explanation of these proposed changes. At a highlevel this patch:
* Changes how we navigate to branch links. This should give us better deeplinking for devices with the Reddit app installed. It should also prevent those rare intent urls we've seen on Android.
* Renames `smartBanner` to `xpromo` in a few places; also removes some unused state keys.

👓  @tomsoir @prashtx


schwers:
bump :haircut_man:

prashtx:
:fish:

PR #1077, opened 2017-03-30
Rearrange User Pages and routes with Posts being the default

uzi:
This rotates the About tab to the end of the list and
makes the Posts the default for viewing user profiles,
more like the native mobile apps.

Other changes/cleanups:
* Removed the anonying `activity` queryParam field
* Displayed karma to match native mobile, fixed models
* Removed the tab icons to look more like native mobile

👓  @schwers @nramadas @prashtx @sffriend


uzi:
For context, this is what it now looks like:

screen shot 2017-03-29 at 6 36 57 pm

src/app/router/handlers/UserActivity.js
```diff
@@ -41,8 +41,8 @@ export default class UserActivityHandler extends BaseHandler {
 
     const { platform: { currentPage }} = getState();
     const { urlParams } = currentPage;
-    this.queryParams.activity = urlParams.commentsOrSubmitted;
-    const activitiesParams = UserActivityHandler.pageParamsToActivitiesParams(this);
+    const activity = urlParams.commentsOrSubmitted || POSTS_ACTIVITY;
```
prashtx: Should we factor this into a helper, so we don't have to repeat this logic in pages/UserActivity? I think it's not obvious why we need this param + OR to determine the current activity without understanding the routing, so it might help to consolidate that into one spot.

prashtx:
1 small question, looks good otherwise

💅

src/app/router/handlers/UserActivity.js
uzi: Yeah, I think it makes sense and lets me clean up things a little bit.  Done!

uzi:
💇

src/apiClient/models/Account.js
```diff
@@ -61,6 +62,9 @@ export default class Account extends RedditModel {
   }
 
   static DERIVED_PROPERTIES = {
+    karma(data) {
+      return data.link_karma + data.comment_karma;
```
schwers: This should check for `data.karma`, that way way if there's a call to `.set()` this will work without the existence of `data.link_karma` and `data.comment_karma`

src/app/components/UserProfileHeader/styles.less
```diff
@@ -5,7 +5,7 @@
 
 .UserProfileHeader {
   &__banner {
-    padding: 2 * @grid-size @grid-size;
+    padding: @grid-size @grid-size;
```
schwers: :nitpick: this can be `padding: @grid-size;`

src/app/router/index.js
```diff
@@ -45,10 +45,11 @@ export default [
   ['/comments/:postId/:postTitle?', CommentsPageHandler, { name: 'comments' }],
   ['/comments', CommentsPageHandler],
   ['/user/:userName/activity', UserActivityRerouteHandler],
+  ['/user/:userName/about', UserProfilerHandler, { name: 'user' }],
   ['/user/:userName/gild', UserProfilerHandler],
   ['/user/:userName/:savedOrHidden(saved|hidden)', SavedAndHiddenHandler],
   ['/user/:userName/:commentsOrSubmitted(comments|submitted)', UserActivityHandler],
-  ['/user/:userName', UserProfilerHandler, { name: 'user' }],
+  ['/user/:userName/', UserActivityHandler],
```
schwers: Do we need the trailing slash here? I believe path-to-regexp adds `/?` by default

src/apiClient/models/Account.js
uzi: Cool, adding

src/app/components/UserProfileHeader/styles.less
uzi: Cool, fixing

src/app/router/index.js
uzi: Nope -- gone

uzi:
💇

schwers:
:fish: 🐠

PR #1088, opened 2017-04-14
Add tutorial to home feed for users with 0 subscriptions

keithblaha:
:eyeglasses: @schwers @brainix 

Adds a tutorial page for new users, for more info see https://reddit.atlassian.net/browse/HOT-140


src/app/components/Tutorial/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { Anchor } from 'platform/components';
+
+class Tutorial extends React.Component {
+  render() {
+    return (
+      <div id='Tutorial'>
+        <div className='TutorialRow'>
+          <img src='/img/tutorial-post.png' />
+          <span>Welcome! Reddit has subreddit communities for every topic imaginable</span>
```
brainix: Not a big deal, and potentially nothing actionable here...  But I prefer using semantic markup as far as possible, even within client-side rendered apps.  If it's not a CSS nightmare, I would've chosen a `<p>` tag here instead of a `<span>`.

src/app/components/Tutorial/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+        </div>
+        <div className='TutorialRow'>
+          <img src='/img/tutorial-vote.png' />
+          <span>Vote on posts and help communities lift the best content to the top</span>
+        </div>
+        <div className='TutorialRow'>
+          <img src='/img/tutorial-discuss.png' />
+          <span>Subscribe to communities to fill up this empty screen with fresh posts</span>
+        </div>
+        <Anchor className='TutorialButton' href='/r/popular'>Explore popular communities</Anchor>
```
brainix: The feedback from yesterday's demo to /u/spez probably shouldn't hold up this PR...  But we may have to change this call to action.  Just leaving a note here for posterity.

brainix:
🐟 from me!  I left a couple of comments, but nothing actionable.

src/app/components/Tutorial/styles.less
```diff
@@ -0,0 +1,42 @@
+@import (reference) '~app/less/variables';
+@import (reference) '~app/less/themes/themeify';
+
+#Tutorial {
```
schwers: The structure of the selectors in this file need a few changes to match our guidelines. Specifically we should:
* not use an id for styling (or for the element, unless we need to support a #/fragment url)
* selectors should be BEM like ([general overview of BEM](https://css-tricks.com/bem-101/))
* *  (i.e. no nesting, `&__` to generate the selector for subcomponents)
* avoid tag names for styling; this makes the styles' intent easier to read and is more flexible (you can generally swap tag types with less changes)

With that in mind I would expect the outline of this file to look something like:

```less
.Tutorial {
  // some styles

  &__row {
    // This will compile to the selector `.Tutorial__row`. 
    // This ensures there is no cascading, but visually its clear that this selector 
    // is for an element inside the `Tutorial` component.
  }

  &__rowImage {
  }

  &__rowText {
  }

  &__button {
  }
}
```

src/app/pages/PostsFromSubreddit.jsx
```diff
@@ -21,10 +22,14 @@ const mapStateToProps = createSelector(
   state => state.subreddits,
   state => state.preferences,
   state => state.modal.id,
-  (pageProps, postsLists, subreddits, preferences, modalId) => {
+  state => state.user,
```
schwers: In general, selectors should be doing as much of the computation as possible (as opposed to computing logged out and has subs in the combine function). It ends up being much more readable, and has subtle effects on how `reselect` memoizes the rendering of the underlying component. 

I'd propose having one selector function that computes `showingTutorial`, and put all of the tutorial logic there.
It could be something like 

```js
(state, pageProps) => {
  const postsListParams = PostsFromSubredditHandler.pageParamsToSubredditPostsParams(pageProps);
  const isFrontPage = !postsListParams.subredditName;
  const { loggedOut } = state.user;
  const { subreddits: subscriptions } = state.subscribedSubreddits;
  const hasSubs = subscriptions && subscriptions.length;

  return !isFrontPage && !loggedOut && hasSubs;
}
```

keithblaha:
:haircut_woman: thank you for the feedback, and for the advice on how to incorporate it! much cleaner this way

src/app/components/Tutorial/index.jsx
keithblaha: i agree with you in general, but it seems throughout this codebase that `<span>` is more common than `<p>` so im going with local convention over global correctness

src/app/components/Tutorial/index.jsx
keithblaha: good to keep in mind, i think on mobile app the CTA is much more clear though and mweb is a bit lower priority so i think itll be ok

keithblaha:
@schwers sorry to poke you about this but it's a bit time sensitive (we've already deployed the r2 portion and this should go out ASAP)- will you be able to review today?

schwers:
@keithblaha sure thing! I'll take a look now

keithblaha:
@schwers ty!

PR #1095, opened 2017-04-20
Chan 515 xpromo exclude subreddits

tomsoir:
https://reddit.atlassian.net/browse/CHAN-515


src/app/featureFlags.js
```diff
@@ -303,7 +298,6 @@ const config = {
   [VARIANT_XPROMO_INTERSTITIAL_COMMENTS_ANDROID]: {
     and: [
       { allowedDevices: [ANDROID] },
-      { allowNSFW: false },
```
schwers: I think in general we still want to have these NSFW checks

src/app/featureFlags.js
```diff
@@ -277,8 +275,6 @@ const config = {
   [VARIANT_XPROMO_PERSISTENT_ANDROID]: {
     and: [
       { allowedDevices: [ANDROID] },
-      { allowedPages: ['index', 'listing', 'comments'] },
```
schwers: It looks like the `allowedPages` should stay here

src/app/selectors/xpromo.js
```diff
@@ -354,8 +355,7 @@ export function isXPromoPersistent(state) {
 
 export function shouldShowXPromo(state) {
   return state.xpromo.interstitials.showBanner &&
-    xpromoIsEnabledOnPage(state) &&
-    xpromoIsEnabledOnDevice(state);
+    anyFlagEnabled(state, [XPROMOBANNER]);
```
schwers: This slightly changes what devices can see xpromo's, specifically it would enable `IPAD`, is that intentional?

src/app/featureFlags.js
tomsoir: We use these two flags (grouped in VARIANT_XPROMO_INTERSTITIAL_COMMENTS_ANDROID) only once, here:
--
export function commentsInterstitialEnabled(state) {
  if (!shouldShowXPromo(state)) { return false };
  return anyFlagEnabled(state, COMMENTS_PAGE_BANNER_FLAGS);
}
--
shouldShowXPromo(state) - now checks allowNSFW() from the 'anyFlagEnabled(state, [XPROMOBANNER])'. So i can leave it still, but this check will be duplicated.

tomsoir:
@schwers 💇‍♂️

src/app/subredditRulesList.js
```diff
@@ -0,0 +1,187 @@
+export const modalExperimentList = [
```
schwers: :nitpick: This should be all upper case with underscores to match the style-guide

src/app/subredditRulesList.js
```diff
@@ -0,0 +1,187 @@
...
+  'csgo', 
+  'lounge', 
+  'prettygirls', 
+  'Overwatch', 
+  'AngelDensetsu', 
+  'food', 
+  'military', 
+  'darknet'];
+
+export const xpromoList = [
```
schwers: :nitpick: This should be upper case as well

tomsoir:
@schwers 💇‍♂️

tomsoir:
@schwers 💇‍♂️

tomsoir:
squash

PR #1099, opened 2017-05-02
chan-525 fix prefixed subreddit names at the XPromo banner

tomsoir:
https://reddit.atlassian.net/browse/CHAN-525

### DIFF
--- a/src/lib/getSubredditFromState.js
+++ b/src/lib/getSubredditFromState.js
@@ -1,5 +1,6 @@
-export default function getSubreddit(state) {
+import get from 'lodash/get';
 
+export default function getSubreddit(state) {
   if (state.platform.currentPage.urlParams.subredditName) {
     return state.platform.currentPage.urlParams.subredditName;
   }
@@ -24,4 +25,15 @@ export default function getSubreddit(state) {
   }
 
   return comment.subreddit;
+}
+
+export function getSubredditNamePrefixed(state) {
+  const subredditName = getSubreddit(state);
+  if (!subredditName) {
+    return;
+  }
+  return get(state,
+    `subreddits.${subredditName.toLowerCase()}.displayNamePrefixed`,
+    `r/${subredditName}`,
+  );
 }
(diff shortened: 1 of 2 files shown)


tomsoir:
💇‍♂️

src/lib/getSubredditFromState.js
```diff
@@ -25,3 +25,21 @@ export default function getSubreddit(state) {
...
+export function getSubredditNamePrefixed(state) {
+  const subredditName = getSubreddit(state);
+
+  if (!subredditName) { return; }
+  const { subreddits } = state;
+
+  if (!subreddits) { return; }
+  const subreddit = subreddits[subredditName];
+
+  if (subreddit) {
```
prashtx: We should be able to use `subreddit.displayNamePrefixed` directly, which already will use `r/fooBar` for true subreddits and `u/someUser` for profile subreddits.

src/lib/getSubredditFromState.js
```diff
@@ -25,3 +25,21 @@ export default function getSubreddit(state) {
...
 }
+
+export function getSubredditNamePrefixed(state) {
+  const subredditName = getSubreddit(state);
+
+  if (!subredditName) { return; }
+  const { subreddits } = state;
+
+  if (!subreddits) { return; }
+  const subreddit = subreddits[subredditName];
```
prashtx: 🔕  Maybe cleaner to use lodash for `const subreddit = get(state, ['subreddits', subredditName])`

tomsoir:
💇‍♂️  @prashtx @schwers

src/lib/getSubredditFromState.js
```diff
@@ -25,3 +26,10 @@ export default function getSubreddit(state) {
 
   return comment.subreddit;
 }
+
+export function getSubredditNamePrefixed(state) {
+  const subredditName = getSubreddit(state);
+  const seachPropPath = ['subreddits', subredditName, 'displayNamePrefixed'];
```
schwers: You'll need to lower case `subredditName` here, and handle the case where the subreddit doesn't exist. I think it would be cleaner to move the subreddit check up front, then you could something like

```js
const subredditName = getSubreddit(state);
if (!subredditName) {
  return;
}

return get(state, 
  `subreddits.${subredditName.toLowerCase()}.displayNamePrefixed`
  `r/${subredditName}`,
);
```

tomsoir:
@schwers  💇‍♂️

PR #1103, opened 2017-05-11
Use an interstitial for over18 profiles

prashtx:
For profiles that have been marked as NSFW, we should use the same NSFW interstitial flow that we do for subreddits.

### DIFF
--- a/src/app/pages/UserActivity/index.jsx
+++ b/src/app/pages/UserActivity/index.jsx
@@ -5,6 +5,8 @@ import { createSelector } from 'reselect';
 import { UserProfileHeader } from 'app/components/UserProfileHeader';
 import PostAndCommentList from 'app/components/PostAndCommentList';
 import SortAndTimeSelector from 'app/components/SortAndTimeSelector';
+import Loading from 'app/components/Loading';
+import NSFWInterstitial from 'app/components/NSFWInterstitial';
 
 import { Section } from '../UserProfile';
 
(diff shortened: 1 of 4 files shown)


prashtx:
👓  @uzi @schwers @sffriend

src/app/pages/UserActivity/index.jsx
```diff
@@ -42,11 +48,36 @@ export const UserActivityPage = connect(mapStateToProps)(props => {
...
+    }
+
+    if (userSubreddit.over18) {
+      return (
+        <div className='UserProfilePage'>
+          <NSFWInterstitial />
+        </div>
+      );
+    }
+  }
```
sffriend: Within the first if, could you do the following instead of having two ifs? It seems redundant to me and I think it would look cleaner.
`if (userSubreddit && userSubreddit.over18) { return NSFW...  } return ...`

src/app/pages/UserProfile/index.jsx
```diff
@@ -16,24 +17,59 @@ const mapStateToProps = createSelector(
...
-  const { myUser, queriedUser, queriedUserRequest, url, isVerified } = props;
+  const {
+    myUser,
+    queriedUser,
+    queriedUserRequest,
+    url,
+    isVerified,
+    queriedUserSubreddit: userSubreddit,
+    preferences,
+  } = props;
```
sffriend: Not sure if this is true for mobile but @fruchtose tells me everyday to make things alphabetical

schwers: Agreed, alphabetization is now preferred, though we haven't been doing it much in mweb.

schwers:
^ Comment updated, but github doesn't send emails for updates :(

uzi:
:fish:

src/app/pages/UserActivity/index.jsx
prashtx: if `userSubreddit` is truthy and `userSubreddit.over18` is false, then we don't want to return the loading component but rather want to fall through to the actual content. I agree that it feels a bit verbose. I think schwers's suggestion captures what you're pointing out, so hopefully I can implement that in a later iteration.

prashtx:
Merged 4c1a44b4162d2de74410c1ebf42514a6465b1ac7

PR #1108, opened 2017-05-18
Serve amphtml link data for all self posts

prashtx:
Render the amphtml link tag for all self post comments pages rather than
just 2%.  By doing the same on desktop, we will let crawlers find and
index the AMP version of all of our self post comments pages.


prashtx:
👓  @tomsoir @uzi

uzi:
:fish:

tomsoir:
🐟

PR #1113, opened 2017-05-25
Wait on user subreddit to determine NSFW status

prashtx:
If a user is a contributor, we need to ensure that their user
subreddit data is available before we can determine if the profile is
NSFW. We also need to look for the user subreddit data in state using
the correct casing, since we key off of lowercased subreddit names.


uzi:
:fish:

PR #1115, opened 2017-05-31
Removing experiment check for onboarding tutorial

curioussavage:


curioussavage:
also 👓  @schwers or @nramadas  we need to get rid of this check since this should be the default experience

keithblaha:
awesome, ty!

PR #1116, opened 2017-05-31
Fix linting errors in PostsFromSubreddit.jsx

curioussavage:
Fixes linting errors as well as one warning caused by my previous commit


keithblaha:
:fish: thanks for cleaning it up!

PR #1118, opened 2017-06-01
Add menu OptOut XPromo link

tomsoir:
https://reddit.atlassian.net/browse/CHAN-673
There are two optOut links now (regular and for the menu):
http://localhost:4444/?no_xpromo_interstitial=true
http://localhost:4444/?no_xpromo_interstitial_menu=true

### DIFF
--- a/src/app/components/LoadingXpromo/OptOutWrapper/index.jsx
+++ b/src/app/components/LoadingXpromo/OptOutWrapper/index.jsx
@@ -1,14 +1,13 @@
 import './styles.less';
 import React from 'react';
-import { OPT_OUT_XPROMO_INTERSTITIAL } from 'app/constants';
+import inlineScript from './script';
+import { OPT_OUT_FLAGS } from 'app/constants';
 
 export class OptOutWrapper extends React.Component {
   constructor(props) {
     super(props);
     this.scriptId = 'app_inline_loader';
-    this.urlFlag = OPT_OUT_XPROMO_INTERSTITIAL.URL_FLAG;
-    this.storageKey = OPT_OUT_XPROMO_INTERSTITIAL.STORE_KEY;
-    this.setExperimentEnabled();
+    this.isEnabled = this.shouldEnableExperiment();
   }
 
   // The first time (when rendering on the server side), we need to store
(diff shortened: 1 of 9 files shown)


src/app/components/LoadingXpromo/OptOutWrapper/index.jsx
```diff
@@ -17,20 +16,21 @@ export class OptOutWrapper extends React.Component {
   // next first rendering on the client side (it must be exactly the same
   // (HTML) as the server).
   setExperimentEnabled() {
```
prashtx: Since this no longers sets anything, should we rename to something like `shouldEnableExperiment`?

src/app/components/LoadingXpromo/OptOutWrapper/script.js
```diff
@@ -0,0 +1,59 @@
...
+          showXpromo = isSomeOptOutFlagOn;
+        }
+      }
+      // Check URL
+      // this check has priority N1. If url-flag is set on, we should apply it first 
+      // example: var url = 'http://localhost:4444/?no_xpromo_interstitial=false';
+      const url = (window.location.search || null);
+      if (url) {
+        const pairs = url.substring(url.indexOf('?')+1).split('&');
+        const isSomeOptOutFlagOn = (flag) => {
```
prashtx: It looks like this will be `true` if there's some arbitrary query parameter present. Should we use something like

    (pair.startsWith(`${flag.URL_FLAG}=`) || pair === flag.URL_FLAG) && pair !== `${flag.URL_FLAG}=false`

src/app/components/LoadingXpromo/OptOutWrapper/script.js
```diff
@@ -0,0 +1,59 @@
...
+          const isSomeOptOutFlagOn = !optOutFlags.reduce((prev, flag) => {
+            return (lsOptOutFlags[flag.STORE_KEY] || prev);
+          }, false);
+          showXpromo = isSomeOptOutFlagOn;
+        }
+      }
+      // Check URL
+      // this check has priority N1. If url-flag is set on, we should apply it first 
+      // example: var url = 'http://localhost:4444/?no_xpromo_interstitial=false';
+      const url = (window.location.search || null);
```
prashtx: Since this is just the search string, not the entire url, can we call it `search` or similar?

src/app/components/LoadingXpromo/OptOutWrapper/script.js
tomsoir: Actually, this should works properly just because it calls from 
optOutFlags.reduce(), and based on our constants (OPT_OUT_FLAGS) at first:
export const OPT_OUT_FLAGS = [
  OPT_OUT_XPROMO_INTERSTITIAL,
  OPT_OUT_XPROMO_INTERSTITIAL_MENU,
];
For example, this link will be ignored: http://localhost:4444/?some_flag=false

tomsoir:
@prashtx 💇‍♂️

src/app/components/LoadingXpromo/OptOutWrapper/script.js
```diff
@@ -0,0 +1,59 @@
...
+          showXpromo = !isSomeOptOutFlagOn;
+        }
+      }
+      // Check URL
+      // this check has priority N1. If url-flag is set on, we should apply it first 
+      // example: var url = 'http://localhost:4444/?no_xpromo_interstitial=false';
+      const search = (window.location.search || null);
+      if (search) {
+        const pairs = search.substring(search.indexOf('?')+1).split('&');
+        const isSomeOptOutFlagOn = (flag) => {
```
prashtx: Discussed IRL: let's rework this code a bit, so that it's more transparent. We can probably use `some` and `every` from lodash to make the code more clearly match intent.

prashtx:
💅

src/app/components/LoadingXpromo/OptOutWrapper/script.js
```diff
@@ -0,0 +1,68 @@
...
+      }, false)
+      showXpromo = !isSomeURLFlagOn; // (3)
+
+      // Case where
+      // URL: http://localhost:4444/?no_xpromo_interstitial=false
+      // LS: {"xpromoInterstitialMenu":true}
+      if(showXpromo && lsTrueFlags.length){ // (4)
+        var isTryToDismissDiffLsAndUrlFlags = lsTrueFlags.some((lsFlag) => {
+          var isEqual = validPairs.some((urlFlag) => ((lsFlag+'=false') === urlFlag));
+          return isEqual? false : true;
```
prashtx: Can this just be `return !isEqual;`?

prashtx:
Minor note, and the inline script needs some lint fixups, otherwise :fish:

PR #1127, opened 2017-06-21
Add 2fa login to mweb

iskandersamerkhanov:
👓 @pwildani @schwers @tomsoir 
Along with these changes are the changes in node-private repo https://github.com/reddit/node-private/pull/8

Added new input for the otp (app code or backup code).
The first login attempt checks whether 2F authentication is enabled. If it is, a field for entering the code is opened.

If you want to test locally and don't have 2fa, i think that @pwildani can add the ability to turn on 2fa login for your accounts.

JIRA Tickets:
https://reddit.atlassian.net/browse/AE-1021
https://reddit.atlassian.net/browse/AE-1024


iskandersamerkhanov:
screenshot from 2017-06-12 16-27-37
screenshot from 2017-06-21 13-05-04
screenshot from 2017-06-21 13-05-23
screenshot from 2017-06-21 13-05-50
screenshot from 2017-06-21 13-06-07
screenshot from 2017-06-21 13-06-37

pwildani:
In my opinion, The B_ prefix should be displayed in grey before and outside the entry box so the user understands which part of the code to type themselves, like how Google's backup codes and SMS codes work with the G- prefix. The native apps only ever ask the user to type the numeric portion.

(Also IMO, it should be "B-", but I was overruled)

iskandersamerkhanov:
👓 @schwers @tomsoir

src/app/components/Login/index.jsx
```diff
@@ -126,6 +231,13 @@ class Login extends React.Component {
         break;
       }
 
+      case loginErrors.WRONG_OTP: {
+        error.otp = activeForm === loginForms.APP_CODE 
+                      ? 'Sorry, that’s not the right app code'
```
pwildani: Do we have wording for these from the product anywhere?

pwildani:
More generally, What's mweb's story for translation and i18n?

The approach looks good from the pure functionality perspective, but I'm not the right person to review if the code fits mweb's style.

iskandersamerkhanov:
👓 @prashtx

uzi:
mweb doesn't really currently have a translation and i18n story ... the desktop 2x codebase does.

src/app/components/Login/index.jsx
iskandersamerkhanov: I replaced that string with string that we use on desktop

src/app/reducers/twoFactorAuthentication.js
```diff
@@ -0,0 +1,28 @@
...
+import * as tfaActions from 'app/actions/twoFactorAuthentication';
+import { loginForms } from 'app/constants';
+
+export const DEFAULT = {
+  activeForm: loginForms.AUTH,
+};
+
+export default function(state=DEFAULT, action={}) {
+  switch (action.type) {
+    case tfaActions.OPEN_APP_CODE_FORM: {
```
schwers: I think it would make sense for this reducer to revert to `DEFAULT` state when the url changes, and when the user logs in. Right now you can click ` browser back`, then `login` and the OTP form is still open instead. Is that intentional?

iskandersamerkhanov:
💇‍♂️ @schwers

PR #1130, opened 2017-06-23
Fix client<>session deeplink generation bug

tomsoir:
https://reddit.atlassian.net/browse/CHAN-728
Reproduced when the cookie 'session_tracker' does not exist.

### DIFF
--- a/src/lib/getSessionIdFromCookie.js
+++ b/src/lib/getSessionIdFromCookie.js
@@ -8,29 +8,30 @@
 
 import cookies from 'js-cookie';
 
-const SESSION_COOKIE_KEY = 'session_tracker';
+export const SESSION_COOKIE_KEY = 'session_tracker';
 
-const parseServerSideCookies = (ctxCookie) => {
-  if (ctxCookie && process.env.ENV === 'server') {
-    const delimiter = ';';
-    const cookieStr = [ctxCookie].join(delimiter);
-    const getSession = new RegExp(`${SESSION_COOKIE_KEY}=(.+?;).*`, 'g');
-    const sessionCookie = cookieStr.replace(getSession, '$1');
-    const isSessionCookieWasFound = new RegExp(delimiter).test(sessionCookie);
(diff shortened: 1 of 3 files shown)


tomsoir:
🕶

prashtx:
Could you add tests for `parseServerSideCookies`?

src/lib/getSessionIdFromCookie.js
```diff
@@ -12,14 +12,13 @@ const SESSION_COOKIE_KEY = 'session_tracker';
...
   if (ctxCookie && process.env.ENV === 'server') {
-    const delimiter = ';';
-    const cookieStr = [ctxCookie].join(delimiter);
-    const getSession = new RegExp(`${SESSION_COOKIE_KEY}=(.+?;).*`, 'g');
-    const sessionCookie = cookieStr.replace(getSession, '$1');
-    const isSessionCookieWasFound = new RegExp(delimiter).test(sessionCookie);
-    return (isSessionCookieWasFound ? sessionCookie : '');
+    const ctxCookieStr = ctxCookie.join();
+    const regexp = new RegExp(`.*${SESSION_COOKIE_KEY}=(.+?);.*`, 'g');
+    const sessionCookie = ctxCookieStr.replace(regexp,'$1');
```
prashtx: Shouldwe be using a `regexp.exec`, so we can avoid the initial `.*`, rather than a replace?

src/lib/getSessionIdFromCookie.js
```diff
@@ -12,14 +12,13 @@ const SESSION_COOKIE_KEY = 'session_tracker';
 
 const parseServerSideCookies = (ctxCookie) => {
```
prashtx: Could you comment on the expected format for ctxCookie here? It looks like we expect something very different between the old and new code.

tomsoir:
@prashtx 💇‍♂️

tomsoir:
@schwers 👓 💇‍♂️

PR #1132, opened 2017-06-27
Fixed offscreen videos buffering on initial load

birakattack:
This prevents all videos within a feed from all loading at the same time in card view. This was reviewed in reddit-mobile-private and tested there. We are just cherry picking the commit over to deploy.

:eyeglasses: @MoMitch


birakattack:
:fish:

MoMitch:
🐟

PR #1133, opened 2017-06-27
Prevent all gifs from loading on sub reddit open.

MoMitch:
Gifs should only load video when isPlaying is true, otherwise we load the preview image.


birakattack:
f7d9e613502dc0140481d3542b9be23a2d2722aa

PR #1134, opened 2017-06-27
*Implement events data for self hosted video content on mobile web

MoMitch:
This implements async event tracking for several consumption events of the new self-hosted video content.

### DIFF
--- a/src/app/components/HTML5StreamPlayer/index.jsx
+++ b/src/app/components/HTML5StreamPlayer/index.jsx
@@ -2,8 +2,12 @@ import React from 'react';
 import dashjs from 'dashjs';
 import { debounce } from 'lodash';
 import './styles.less';
-
 import PostModel from 'apiClient/models/PostModel';
+import { trackVideoEvent } from 'app/actions/posts';
+import { connect } from 'react-redux';
+import { VIDEO_EVENT } from 'app/constants';
+import { createSelector } from 'reselect';
+import { isCommentsPage } from 'platform/pageUtils';
 
 const T = React.PropTypes;
 
@@ -38,6 +42,8 @@ class HTML5StreamPlayer extends React.Component {
       mediaPlayer: null,
       videoLoaded: false,
       autoPlay: true,
+      lastUpdate: null,
+      totalServedTime: 0,
     };
   }
 
@@ -92,10 +98,12 @@ class HTML5StreamPlayer extends React.Component {
       if (videoIsInView) {
         if (video.paused && !window.fullScreen) {
           video.play();
+          this.sendTrackVideoEvent(VIDEO_EVENT.SCROLL_AUTOPLAY);
         }
       } else {
         if (!video.paused) {
           video.pause();
+          this.sendTrackVideoEvent(VIDEO_EVENT.SCROLL_PAUSE);
         }
       }
 
@@ -117,6 +125,12 @@ class HTML5StreamPlayer extends React.Component {
     if (this) {
       this.setState({videoLoaded: true});
       this.isScrolledIntoView();
+
+      const video = this.refs.HTML5StreamPlayerVideo;
+      if ((this.props.postData.videoPlaytime > 0) && (video.paused === false)) {
+        this.setState({totalServedTime: this.props.postData.videoPlaytime * 1000.0});
+        this.sendTrackVideoEvent(VIDEO_EVENT.CHANGED_PAGETYPE, this.getPercentServed());
+      }
     }
   }
 
@@ -156,6 +170,10 @@ class HTML5StreamPlayer extends React.Component {
   }
 
   componentWillUnmount() {
+    if (this.state.totalServedTime > 0) {
+      //Video has been watched and we are now removing it.
+      this.sendTrackVideoEvent(VIDEO_EVENT.SERVED_VIDEO, this.getPercentServed());
+    }
     const video = this.refs.HTML5StreamPlayerVideo;
     video.removeEventListener('canplay', this.videoDidLoad, false);
     window.removeEventListener('scroll', this.state.debounceFunc);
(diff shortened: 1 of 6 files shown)


MoMitch:
@birakattack

src/app/components/HTML5StreamPlayer/index.jsx
```diff
@@ -502,6 +550,201 @@ class HTML5StreamPlayer extends React.Component {
...
+    const payload = {
+      ...this.buildBaseEventData(),
+      source: 'videoplayer',
+      verb: 'scroll',
+      noun: 'autoplay',
+    }
+    this.sendTrackVideoEvent('videoplayer__scroll_autoplay',payload);
+  }
+
+  getPctServed() {
```
birakattack: nit: can we just spell out percent here, so `getPercentServed()`?

src/app/components/HTML5StreamPlayer/index.jsx
```diff
@@ -502,6 +550,201 @@ class HTML5StreamPlayer extends React.Component {
...
+
+  sendVideoServedEvent() {
+    const payload = {
+      ...this.buildBaseEventData(),
+      ...this.getPctServed(),
+      source: 'videoplayer',
+      verb: 'served',
+      noun: 'video',
+    }
+    this.sendTrackVideoEvent('videoplayer__served_video',payload);
```
birakattack: Can we just move the source, verb, noun stuff into the buildBaseEventData method and just pass the `event_type` and parse that to set those values?

birakattack: In fact can we just move all of this logic into `this.sendTrackVideoEvent` and allow an optional parameter that will be splatted into the base payload?

src/app/components/HTML5StreamPlayer/index.jsx
```diff
@@ -502,6 +551,117 @@ class HTML5StreamPlayer extends React.Component {
...
+      sr_name: postData.subreddit,
+      sr_fullname: postData.subredditId,
+      sr_id: parseInt(subredditShortID, 36),
+      page_type: pageType,
+    };
+
+    return payload;
+  }
+
+  sendTrackVideoEvent(eventType) {
```
birakattack: We should just add an optional parameter to this method that can take an object that should be splatted with the base event payload. 

```
sendTrackVideoEvent(eventType, optionalParams={}) {
    let payload = { 
        ...this.buildBaseEventData(),
        ...optionalParams,
    };
    this.props.dispatch(trackVideoEvent(eventType,payload));
}
```

src/app/components/HTML5StreamPlayer/index.jsx
```diff
@@ -502,6 +551,117 @@ class HTML5StreamPlayer extends React.Component {
...
+      max_timestamp_served: parseInt(this.state.totalServedTime),
+      percent_served: pctServed,
+    };
+
+    return payload;
+  }
 }
 
-export default HTML5StreamPlayer;
+export default connect(null)(HTML5StreamPlayer);
```
birakattack: Why the `connect(null)` here?

src/app/components/HTML5StreamPlayer/index.jsx
```diff
@@ -91,10 +95,12 @@ class HTML5StreamPlayer extends React.Component {
       const video = this.refs.HTML5StreamPlayerVideo;
       if (videoIsInView) {
         if (video.paused && !window.fullScreen) {
+          this.sendTrackVideoEvent(VIDEO_EVENT.SCROLL_AUTOPLAY);
```
birakattack: Should we move this call below the `video.play();` call?

src/app/components/HTML5StreamPlayer/index.jsx
```diff
@@ -91,10 +95,12 @@ class HTML5StreamPlayer extends React.Component {
...
           video.play();
         }
       } else {
         if (!video.paused) {
+          this.sendTrackVideoEvent(VIDEO_EVENT.SCROLL_PAUSE);
```
birakattack: Same as above should we move this to after `video.pause();` is called?

src/app/components/HTML5StreamPlayer/index.jsx
```diff
@@ -502,6 +551,117 @@ class HTML5StreamPlayer extends React.Component {
...
+      target_author_id: parseInt(postData.author, 36),
+      target_author_name: postData.author,
+      target_created_ts: msSinceCreation,
+      target_id: parseInt(postData.id, 36),
+      target_url: postData.cleanUrl,
+      target_url_domain: postData.domain,
+      sr_name: postData.subreddit,
+      sr_fullname: postData.subredditId,
+      sr_id: parseInt(subredditShortID, 36),
+      page_type: pageType,
```
birakattack: Can we add a `target_type` as well? It was recently added to the spec but essentially we want to know whether its a video or a gif.

src/app/components/HTML5StreamPlayer/index.jsx
MoMitch: As above.

src/app/components/HTML5StreamPlayer/index.jsx
MoMitch: Correct, moving now.

src/app/components/HTML5StreamPlayer/index.jsx
MoMitch: what should this return?

src/app/components/HTML5StreamPlayer/index.jsx
MoMitch: changed to connect() instead, but the connect in general is here to allow the getState for analytics.

MoMitch:
:eyeglasses: @birakattack

PR #1142, opened 2017-07-26
Swap out ad on view change

griffinmichl:
👓 @dwick 

When we change theme or compact mode, I destroy the slot and build a new one, so that we get a new ad with new kv (basically make view changes act like a page refresh for ads).

### DIFF
--- a/src/app/pages/Comments.jsx
+++ b/src/app/pages/Comments.jsx
@@ -58,11 +58,13 @@ function CommentsPage(props) {
         subredditShowSpoilers={ spoilersEnabled }
         key='post'
       />
-      <BannerAd
-        id='comments-banner'
-        shouldCollapse
-        listingName='comments'
-      />
+      { !post.promoted &&
+        <BannerAd
+          id='comments-banner'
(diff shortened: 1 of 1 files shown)


src/app/components/BannerAd/index.jsx
```diff
@@ -4,7 +4,12 @@ import { createStructuredSelector } from 'reselect';
 import classNames from 'lib/classNames';
 import getSubreddit from 'lib/getSubredditFromState';
 import createBannerProperties from 'lib/createBannerProperties';
-import { defineSlot, destroySlot, getSlotId } from 'lib/dfp';
+import {
+  defineSlot,
+  destroySlot,
+  getSlotId,
+  refreshSlot,
```
dwick: you aren't using this anywhere

griffinmichl: Ah good call. I originally tried to just refresh the slot, but refreshing doesn't update targeting

dwick:
🐟 CI is failing because of the unused method. LGTM otherwise.

PR #1148, opened 2017-08-08
Implement whitelist_status logic for in feed ads

dwick:
👓  @griffinmichl 

the logic for whitelist status is here: https://reddit.atlassian.net/browse/ADS-1850

need to get the api changes out before i can E2E test this.


src/apiClient/models/PostModel.js
```diff
@@ -50,6 +50,7 @@ export default class PostModel extends RedditModel {
     removed: T.bool,
     approved: T.bool,
     spam: T.bool,
+    whitelistStatus: T.nop,
```
dwick: `nop` because otherwise it coerces `null` into empty string

src/app/components/Post/index.jsx
```diff
@@ -148,7 +148,7 @@ export function Post(props, context) {
   const displayCompact = compact || mixedExpActive;
   if (post.isBlankAd) {
     // Return an empty div if it's a blank ad
-    return <div class='blankAd'/>;
+    return <div className='blankAd'/>;
```
dwick: the other option here would be to leave the class off completely to avoid adblock since this is just for counting. i ran into some confusion because `class` was defined and I couldn't find the div.

src/app/components/PostsList/index.jsx
```diff
@@ -59,38 +75,72 @@ const renderPostsList = props => {
...
-    return <Post { ...postProps } />;
+    return <Post { ...postProps } postId={ postId } key={ `post-id-${postId}` }/>;
   });
 
-  const dfpAdLocation = dfpAdLocationFromPosts(records.map(result => posts[result.uuid] || {}));
-
   // eslint-disable-next-line eqeqeq
-  if (shouldAdFallback && dfpAdLocation != null) {
-    injectDfp(postsList, dfpAdLocation);
+  if (adLocation != null && canRequestAds(adSlotWhitelistStatus)) {
```
dwick: I'm able to render both ads in the same location (blank & dfp) because the blank isn't prematurely inserted into the post list before calling `adLocationForSubreddit` and `adLocationForDefaultListing` like it used to be. this means we don't have to mess with the index like you were forced to before.

src/lib/ads.js
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export const adLocationForSubreddit = ({ posts=[], whitelistStatus }) => {
+  const canShowAdNextTo = (post, whitelistStatus) => {
+    // eslint-disable-next-line eqeqeq
+    if (post.whitelistStatus == null || post.whitelistStatus === 'no_ads') {
+      return false;
+    }
+
+    return post.whitelistStatus === whitelistStatus;
```
dwick: subreddit whitelists should always match. if a post has a status different than it's parent subreddit that means it was either flagged by smyte or is nsfw in a non-nsfw subreddit.

src/lib/ads.js
```diff
@@ -0,0 +1,106 @@
...
+  const postBefore = posts[index - 1];
+  // `posts` doesn't contain any ads, so the postAfter the ad
+  // has the same index the ad *will* have.
+  const postAfter = posts[index];
+  const whitelistStatus = getMostRestictiveWhitelistStatus(
+    postBefore.whitelistStatus,
+    postAfter.whitelistStatus,
+  );
+
+  if (DEFAULT_LISTING_DISALLOWED.has(whitelistStatus)) {
```
dwick: for default listings we just check if the posts around the ad are naughty since the listing itself doesn't have a whitelist status.

griffinmichl:
🐟

PR #1151, opened 2017-08-15
Add amazon a9 header bidding

dwick:
ticket: https://reddit.atlassian.net/browse/ADS-1909

Currently this doesn't have support for some params that r2 uses (`subreddit`, `SIS_category`, etc.), but the docs amazon passed me seem incomplete and I'm not sure how we're supposed to do that with a SPA. r2 passes them via `apstag.init`, which can only be called once. I'll reach out and fix that up if possible. This should still work however.

### DIFF
--- a/src/lib/dfp.js
+++ b/src/lib/dfp.js
@@ -3,6 +3,8 @@ import isFakeSubreddit from 'lib/isFakeSubreddit';
 
 const FRONTPAGE_NAME = 'frontpage';
 
+const adSlots = {};
+
 export const getSlotId = (listingName, subredditName) => {
   const parts = [config.dfpSlotPrefix];
 
@@ -23,14 +25,23 @@ export const defineSlot = (element, {
   properties,
   shouldCollapse,
   sizes,
+  a9 = false,
 }) => {
   const googletag = (window.googletag = window.googletag || {});
   googletag.cmd = googletag.cmd || [];
   return new Promise(resolve => {
     googletag.cmd.push(function() {
+      // make sure we don't try to create the same ad slot twice.
+      destroySlot(slot);
+
       const adSlot = googletag
         .defineSlot(slot, sizes, id)
         .addService(googletag.pubads());
+
+      // save a reference so we can look this up later.
+      // outside of this closure we should always be using this
+      // reference, not `adSlot`.
+      adSlots[slot] = adSlot;
 
       if (shouldCollapse) {
         adSlot.setCollapseEmptyDiv(true);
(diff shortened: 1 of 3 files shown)


src/lib/dfp.js
```diff
@@ -78,4 +93,18 @@ export const setupGoogleTag = () => {
...
+  // eslint-disable-next-line
+  !function(a9,a,p,s,t,A,g){if(a[a9])return;function q(c,r){a[a9]._Q.push([c,r])}a[a9]={init:function(){q('i',arguments)},fetchBids:function(){q('f',arguments)},_Q:[]};A=p.createElement(s);A.async=!0;A.src=t;g=p.getElementsByTagName(s)[0];g.parentNode.insertBefore(A,g)}('apstag',window,document,'script','//c.amazon-adsystem.com/aax2/apstag.js');
+
+  // this is defined in the above IIF
+  window.apstag.init({
+    pubID: '3379',
+    adServer: 'googletag',
+    bidTimeout: 2e3,
+    params: {
+      position: 'feed',
```
griffinmichl: What's this? Are we not setting up header bidding for banners as well? Banners are worth more

src/lib/dfp.js
```diff
@@ -40,10 +40,25 @@ export const defineSlot = (element, {
...
+        slots: [{
+          slotID: id,
+          sizes,
+        }],
+        timeout: 2e3,
+      }, (bids) => {
+        // we only request a single bid.
+        const bid = bids[0];
+
+        ['amznbid', 'amzniid'].forEach(function(key) {
```
griffinmichl: Has mike setup DFP for mweb ads to handle these keys?

src/lib/dfp.js
dwick: i forgot we had banners

src/lib/dfp.js
dwick: amazon does this i think

src/lib/dfp.js
griffinmichl: We'll probably want to get setup for those as well. They're at least 5x the value of in feed.

src/lib/dfp.js
griffinmichl: Are you sure? These are keys being passed to DFP. How would amazon configure them?

griffinmichl:
🐟 Once this is tested manually

dwick:
screen shot 2017-08-15 at 4 45 12 pm

💇

griffinmichl:
🐟

PR #1152, opened 2017-08-15
Add crosspost data to screenview events

uzi:
Fetch a bit of the crosspost data info into the
Post model. For now in the interest of time, we're
not bothering to process the data.

If said data exists, then we add it to the screenview
event when viewing a post comments page.

👓  @prashtx @schwers @nramadas

### DIFF
--- a/src/apiClient/models/PostModel.js
+++ b/src/apiClient/models/PostModel.js
@@ -21,6 +21,7 @@ export default class PostModel extends RedditModel {
     brandSafe: T.bool,
     cleanPermalink: T.link,
     cleanUrl: T.link,
+    crosspostParent: T.string,
     distinguished: T.string,
     domain: T.string,
     downs: T.number,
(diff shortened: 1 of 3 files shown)


uzi:
👓  @tomsoir Adding Artem because of the second patch ... maybe he has some thoughts as to why things were done the way they were.

src/apiClient/models/PostModel.js
```diff
@@ -201,6 +205,22 @@ export default class PostModel extends RedditModel {
     thumbnail(data) {
       return cleanThumbnail(data.thumbnail);
     },
+
+    crosspostParentObj(data) {
+      // Note: ultimately we'll want to store this as a top level post in
```
uzi: 👍

src/apiClient/models/PostModel.js
```diff
@@ -201,6 +205,22 @@ export default class PostModel extends RedditModel {
...
+      // the redux store and reference via crosspostIds.
+      if (data.crosspost_parent_list && data.crosspost_parent_list.length) {
+        return data.crosspost_parent_list[0];
+      }
+      return null;
+    },
+
+    crosspostIds(data) {
+      if (data.crosspost_parent_list && data.crosspost_parent_list.length) {
```
uzi: 👍

uzi:
💇  Right, I forgot about that... fixed, thanks

PR #1156, opened 2017-08-20
Add fallback value to PostsLists.props.posts.

schwers:
e4f62c71fa91c6a34b997abeccfa882bd0f1cf86 introduced in-feed dfp ads; computing this location depends on `state.posts`. A few places, notably the SearchPage, use the unconnected version
of PostsList, which caused search results to indefinetly load.


griffinmichl:
🐟

PR #1157, opened 2017-08-20
Remove DNT support

dwick:
👓 @griffinmichl 

ticket: https://reddit.atlassian.net/browse/ADS-1934


griffinmichl:
🐟 

Were we calculating DNT with GTM here?

PR #1158, opened 2017-08-20
a9: Ensure all targeting kvps are passed to gpt

dwick:
👓 @griffinmichl 

from mike:

> Since you are not using setdisplaybids and are manually managing the bids, you need to pass the amznsz value to us. The DFP LI is expecting amznsz=320x50 so right now we are not passing any bid values.
> Please make a small code change on your end to handle the new amznsz param (to handle the new orders). This is what you have currently:
> ["amznbid", "amzniid"].forEach(function(e) {
> Please change that to:
> apstag.targetingKeys().forEach(function(e) {

### DIFF
--- a/src/lib/dfp.js
+++ b/src/lib/dfp.js
@@ -80,7 +80,7 @@ export const defineSlot = (element, {
 
             // we only request a single bid.
             const bid = bids[0];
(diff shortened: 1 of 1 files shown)


griffinmichl:
🐟 Amazon seems legit to work with: "here's the code you need to change"

PR #1160, opened 2017-08-24
Prevent previously opened reddit videos from expanding in compact mode

MoMitch:
If no thumbnail is available then the reddit video should not expand due to user history in the feed, it can instead be viewed in the comments section.

https://reddit.atlassian.net/browse/MEDIA-410

### DIFF
--- a/src/app/components/Post/index.jsx
+++ b/src/app/components/Post/index.jsx
@@ -200,9 +200,12 @@ export function Post(props, context) {
     );
   }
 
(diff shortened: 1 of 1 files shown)


MoMitch:
👓 @birakattack

MoMitch:
https://reddit.atlassian.net/browse/MEDIA-410

MoMitch:
Implemented as part of: https://github.com/reddit/reddit-mobile/pull/1163

PR #1161, opened 2017-08-24
Refactor of reddit_video controls and updated UI

MoMitch:
This PR contains all UI and state updating improvements for the native reddit video player

PR #1166, opened 2017-09-26
Remove console logs

MoMitch:
Setting console logging to false for dashjs after creation.

Note that this should be landed after https://github.com/reddit/reddit-mobile/pull/1161 as the player creation's location has moved.

https://reddit.atlassian.net/browse/MEDIA-492


MoMitch:
👓  @birakattack

MoMitch:
This will be a part of the fix events patch

PR #1176, opened 2017-11-21
Fix page width issue with wide tables

curioussavage:
This sets tables to display in inline-block mode so we
can set a max width. Tables will scroll horizontally now


curioussavage:
let me double check. Can't remember if I looked at that.

curioussavage:
:haircut_man:  good call. I applied the style changes to the mixin for tables so it affects both. Works great now.

PR #1181, opened 2018-01-12
Replace Ads request with "withAds" query param

iandeboisblanc:
Addresses tickets ADS-2595 and ADS-2653


iandeboisblanc:
👓 @dwick 

(Hi Dave, I'm Ian. We work together now. I'd love if you'd take a look and drop some feedback when you get the chance)

package.json
```diff
@@ -27,6 +27,7 @@
     "@r/middleware": "~0.8.1",
     "@r/private": "1.1.0",
     "@r/widgets": "0.2.2",
+    "@researchgate/react-intersection-observer": "^0.5.0",
```
dwick: please peg this version to `"0.5.0"`

src/app/components/Ad/index.jsx
```diff
@@ -1,84 +1,93 @@
...
     // ownProps
-    adId: T.string.isRequired,
+    postId: T.string.isRequired,
     placementIndex: T.number.isRequired,
     postProps: T.object.isRequired,
 
     // props from connect
-    adTracked: T.bool.isRequired,
-    onAdMadeImpression: T.func.isRequired,
+    onMadeImpression: T.func.isRequired,
```
dwick: 🔕 i think `onImpression` would be a better name

src/app/components/Ad/index.jsx
```diff
@@ -1,84 +1,93 @@
...
+    onMadeViewableImpression: T.func.isRequired,
```
dwick: same as above

src/app/components/Ad/index.jsx
```diff
@@ -1,84 +1,93 @@
...
-      this.props.onAdMadeImpression();
+  handleViewabilityChange = e => {
+    // an element must be 50% viewable for at least 1 second
+    // to be consider viewable per the IAB standard.
+    // Regular impressions are being fired when 50% viewable ever.
+    if (e.isIntersecting && e.intersectionRatio >= VIEWABILITY_THRESHOLD) {
+      const { onMadeViewableImpression, onMadeImpression } = this.props;
+      if (!this.state.madeImpression) {
+        // we only need to track the impression once.
+        onMadeImpression();
```
dwick: reference error: i think you want `this.props. onMadeImpression`

src/app/components/Ad/index.jsx
```diff
@@ -1,84 +1,93 @@
...
+        this.setState({ madeImpression: true });
+      }
+      this.timer = window.setTimeout(() => {
+        onMadeViewableImpression();
```
dwick: same as above

src/app/components/Ad/index.jsx
```diff
@@ -1,84 +1,93 @@
...
-    const middleIsAboveBottom = middle < winHeight;
-    const middleIsBelowTop = bottom > TOP_NAV_HEIGHT + height / 2;
-
-    if (middleIsAboveBottom && middleIsBelowTop) {
-      window.removeEventListener('scroll', this.scrollListener);
-      this.props.onAdMadeImpression();
+  handleViewabilityChange = e => {
+    // an element must be 50% viewable for at least 1 second
+    // to be consider viewable per the IAB standard.
+    // Regular impressions are being fired when 50% viewable ever.
```
dwick: is 50% actually a requirement for an impression? i know you're just reimplementing what was already there, but it looks like the IAB standard for an impression is just a request/response from the ad server: https://www.iab.com/wp-content/uploads/2015/06/Ad-Impression-Measurment-Guideline-US.pdf

might be worth bringing this up with @griffinmichl

src/app/components/Ad/index.jsx
```diff
@@ -1,84 +1,93 @@
...
+    if (e.isIntersecting && e.intersectionRatio >= VIEWABILITY_THRESHOLD) {
+      const { onMadeViewableImpression, onMadeImpression } = this.props;
+      if (!this.state.madeImpression) {
```
dwick: 🔕 i'm a fan of wrapping prop callbacks in functions if there's any addition logic to firing the handler:

```
onImpression = () => {
  if (this.state.madeImpression) return;
  
  this.setState({ madeImpression: true }, this.props.onMadeImpression);
}
```

then you can just call `this.onImpression` here. same thing can be applied to the viewable impression

src/lib/ads.js
```diff
@@ -1,112 +1,29 @@
...
-const getMostRestictiveWhitelistStatus = (...statuses) => {
-  // eslint-disable-next-line eqeqeq
-  if (statuses.find(x => x == null)) {
-    return null;
-  }
-
-  const index = Math.min(...statuses.map(s => WHITELIST_STATUS.indexOf(s)));
-
-  return WHITELIST_STATUS[index];
+const FiredPixelSet = new Set();
```
dwick: Do you think an LRU cache would be a good idea here, especially since this is targeting phones?

dwick:
Hi Ian, nice to meet you! This looks really good, the only required change is the package version, everything else is just questions/suggestions.

Before merging this can you squash it into 2 commits (request ads server side and events overhaul) that can stand on their own. ie. the first one still works if you revert the 2nd one? If not, one commit is fine. As it stands now if I were to revert one of these commits it would break, which is not ideal.

💅

src/app/components/Ad/index.jsx
griffinmichl: We chatted about it. While firing the impression earlier is correct, firing is when the ad is in view would likely increase performance. I'll check with Jamie about his thoughts on when to fire impressions.

griffinmichl:
Note: Can't be deployed until ad server ships due to server side ad injection latency

src/app/components/Ad/index.jsx
griffinmichl: Jamie said impression pixels should fire when the component mounts.

src/lib/ads.js
iandeboisblanc: After speaking to Griffin originally we thought it might not be necessary; butttt then it seemed fun, so I implemented one based on the existing LRUCache in reddit-desktop.

iandeboisblanc:
I'm going to keep this PR open while I add support for other pixel trackers (upvote, comment, etc) per suggestion from @griffinmichl . However, I incorporated all of your feedback from the earlier review, @dwick . Thanks!

iandeboisblanc:
👓 @dwick 

This PR includes all the things you saw before (plus a few fixes), but it also now incorporates pixel tracking for impressions (when in viewport), viewable impressions (50% for 1 sec), viewing comments, submitting comment, votes, & comment votes. Each commit should be able to stand on its own now.

src/app/components/Ad/index.jsx
```diff
@@ -1,84 +1,110 @@
...
+  checkAdblock() {
+    if (isHidden(findDOMNode(this))) {
+      this.props.onAdblockDetected();
+    }
+  }
+  
+  handleImpressionChange = e => {
+    // an element must enter the viewport to register as impression
+    if (e.isIntersecting) {
+      this.onImpression();
```
dwick: I thought we were firing this on mount?

dwick:
💅 just the impression issue. lg!

src/app/components/Ad/index.jsx
griffinmichl: After talking to Jamie and Tom about it, we think scrolling into view is the way to go. While doing it on mount is technically a consistent definition, how it actually turns out on the client is much different between different clients (IOS and Desktop mounting is effectively scrolling into view, while on mweb there's a good chance it would mount but never be seen).

Sorry for the confusion.

dwick:
:fish:

PR #1182, opened 2018-01-15
Only show locked icon once per post

griffinmichl:
👓 @dwick 

Locked comments currently show the locked icon twice. Once on the left side of the header and once on the right. Confirmed with Jamie he wants to remove the one on the left side (to match the native apps).


dwick:
🐟

PR #1184, opened 2018-01-24
Add redirect to /register if user.loggedOut

aboekhoff:
Intercept the call to toggle and redirect to /register if user is logged out.
@arjunb023 
@sffriend


sffriend:
💅 I think you need to remove the package-lock.json file

arjunb023:
Yup once `package-lock.json` is removed, looks good to me. 💅

